PR #1, opened 2018-12-01
Fix .gitignore after building on mac OSX (using ./configure unix makefiles) still show…

fab672000:
…s not ignored binaries. Note that fltk blocks, checkers and sudoku games do have special content inside the .app folder, so ensure that these are not globally ignored while cleaning up the other test apps binary content that should be ignored.

Please team review the change.


fab672000:
Ok, added the .DS_Store files too as recommended, merged.

PR #3, opened 2018-12-08
Add COPIED_{IMAGE,DEIMAGE} widget user flags.

deech:


Albrecht-S:
Here's a quick analysis of **some parts** of the new code:

```
void Fl_Widget::image(Fl_Image* img) {
  if (label_.image == img) return;
  if (flags() & COPIED_IMAGE) {
    if (img) {
      label_.image = img->copy(img->w(),img->h()); // [1]
    }
    else {
      delete label_.image; // [2]
    }
  }
  else {
    label_.image = img;
  }
}

```
[1] you assign a new image but don't delete the old image
[2] you delete the old image but don't assign the new image (NULL)

The method Fl_Widget::image(Fl_Image* img) would have to /assign/ an image pointer (and not copy the image!) to be compatible with the previous version, even if the flag COPIED_IMAGE /was/ set previously. This method would always have to /clear/ the COPIED_IMAGE flag because it does not copy the image (your version copies the image).

We need another method to copy and assign the image.

I don't see any assignment of the COPIED_[DE]IMAGE flag in the proposed diff PR #3.

deech:
I made the code analogous to `label()` and `copy_label()`.

deech:
Happy New Year!

Could I get some feedback on this PR? If the core team agrees I'd love to have this functionality.

Albrecht-S:
Happy new year as well!

The PR as-is has some issues you can help to resolve, for instance that we decided to have the doxygen docs in the file where the function is **defined** (as opposed to declared). This means if the implementation is in the .cxx file, so should be the documentation.

I'm working on another issue: an `Fl_Image` can also be an `Fl_Shared_Image`, but simply calling operator `delete` on an Fl_Shared_Image is not allowed, so we must find a better solution. As said, I'm working on it, but I'm also working on other things, so please don't hold your breath.

I'll let you know when done.

MatthiasWM:
IMHO it would be enough to set a flag, telling the Widget to delete an image when the Widget itself is deleted or the image is replaced (transfer ownership). As opposed to copy_label() I would leave it to the user to provide a copy of an image if wanted.

If implemented, I will however sacrifice two more bits to manage the user_data field on exit as well.

MatthiasWM:
@Albrecht-S I would take this over from you if you like.

My implementation would be different though. All we would do is add methods to set and get flags that tell us if the user owns the image or if the widget owns it. If the widget owns it (the user transferred ownership to the widget), the widget will delete the image(s) when it is deleted itself.

So I would add
```
void bind_image(Fl_Image*); // moves ownership to the widget
void bind_deimage(Fl_Image*);
void bid_image(int doit); // doit=1: widget owns image, doit=0: user owns image
int image_bound();
void bind_deimage(int doit);
int deimage_bound();
```
Alternative names for `move_image()` could be `image_hold()`, `move_image()`, `retain_image()`

It's incredibly convenient to have a set-and-forget if you need more than one word of userdata, especially for dynamically allocated widgets. So I would also add:
```
class Fl_User_Data { virtual ~ Fl_User_Data() {} }; // to allow complex user data
bind_user_data(void *); // useful as a set-and-forget if more than one word of user data is needed
bind_user_data(Fl_User_Data *); 
callback_bind_userdata(Fl_Callback*, void*);
callback_bind_userdata(Fl_Callback*, Fl_User_Data*);
```

This requires no additional memory, six lines of new code beyond the setters and getters.

Albrecht-S:
@MatthiasWM Thanks, I would be glad if you took this PR. Assigned to you now.

- I'm also fine with your proposed changes using `bind_image()` etc. rather than `copy_image()`.
- The only important point is to use `image/deimage->release()` when the widget is deleted to be compatible with `Fl_Shared_Image`.
- I wouldn't like to mix the `user_data` stuff in here though. This should be discussed (in fltk.coredev?) and implemented separately.

I saw your review request but didn't do it yet. Please don't hold your breath...

Albrecht-S:
Note: most if not all of these "... mentioned ..." messages above are very likely fake messages because someone wrote '#' 3 (w/o the quotes) in an unrelated issue or PR.

Albrecht-S:
@MatthiasWM Regarding your review request: I'm confused because there's no code to review besides the OP's PR.

I believe the best you can do is to open a new PR and then (once confirmed/merged) we can close **this** PR #3 w/o further action.

Note: I removed the review request for now.

MatthiasWM:
I can't remember sending a review request, but further up it says "[Albrecht-S] self-requested a review"? Doesn't matter. I will do a PR for `bind_image`/`bind_deimage` first.

Albrecht-S:
> I can't remember sending a review request ...

Sorry, my fault. This PR is so old...

MatthiasWM:
I implemented this PR differently in #589 . Some initial testing worked well.

I had to make a few decisions for replacing images at runtime.

Calling `bind_image()` with a new image will delete the old image if it was bound, and then set the new image, and bind that. If old and new image are the same, nothing happens.

Calling `image()` with a new image will delete the old image if it was bound, and set the new image without binding it. If old and new are the same, the image will not be deleted, but it will be unbound.

Calling `image()` or `bind_image()` with NULL will delete the old image if it was bound and not set a new image.

Shared images are handled.

PR #8, opened 2018-12-26
Travis implementation supporting  build and doc generation of  four targets 

fab672000:
This implementation differs from existing fltk-test and also gitlab implementations :
 - minimum, fast light implementation: stripped all unused constructions compared to other implementations
 - supports successful builds of four targets (linux, osx and then gcc and clang for each of them)
- uses before_script instead of before_install
- uses after_success for building documentation, instead of building always if error
- only builds documentation on linux targets for now (can easily change in future)
As I don't have access to travis on this branch one can check the results on my other fork [here](https://travis-ci.org/fab672000/fltk)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,53 @@
+sudo: required
+dist: trusty
+cache: ccache
+
+language:
+  - c++
+# - objective-c
+
+os:
+  - linux
+  - osx
+  
+compiler:
+  - gcc
+  - clang
+
+branches:
+  only:
+  - master
+  - /^feature.*$/
+
+before_script:
+  - |
+    if [ $TRAVIS_OS_NAME == linux ]; then
+       sudo apt-get update -q
+       sudo apt-get install -y libxinerama-dev libxcursor-dev libasound2-dev
+       sudo apt-get install -y doxygen 
+    elif [ $TRAVIS_OS_NAME == osx ]; then
+      brew install ccache # need to install on OSX
+      export PATH="/usr/local/opt/ccache/libexec:$PATH"
+      brew install doxygen
+    fi
+
+script:
+  - |
+    autoconf
+    ./configure
+    make
+
+after_success:
+  - |
+    if [ $TRAVIS_OS_NAME == linux ]; then
+      doxygen --version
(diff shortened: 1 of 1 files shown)


MatthiasWM:
I don't know too much about Travis, but the file seems to be a valid starting point. It would be nice to get the CMake build tested as well, and also the MSWindows builds.

fab672000:
> I don't know too much about Travis, but the file seems to be a valid starting point. It would be nice to get the CMake build tested as well, and also the MSWindows builds.

Hi Matt I had the same thoughts, thanks for that!

And was also thinking of automatic unit testing for more TDD approach, 

Also thinking of adding a static analyzer that would reject if a new submission would have some potential problems,  

Finally thinking of adding a code beautifier that detects too badly out of formatting code styles (i.e. tabs are killing us softly, extra spaces in diffs are distracting, etc..)

Note that Albrecht did a great job for end of lines automatic transformations at the git level so we don't need to care about theses. 

What is nice with the Travis integration is that a PR can get automatically flagged as problematic and fixed by the author before code review starts, so we could reject more easily some pull requests that way.

Once you get used to it, it is actually great that assignees can get a chance to change the proposal by reviewers input before it gets to the master repository.

fab672000:
> I don't know too much about Travis, but the file seems to be a valid starting point. It would be nice to get the CMake build tested as well, and also the MSWindows builds.

@MatthiasWM I implemented cmake and also static analyzer! Look at my feature change  [here](https://github.com/fltk/fltk/pull/8/commits/8c44c9a6b15b077f418269465bf88420d2d2f043)

I also convenient env definitions based on matrix definitions to avoid too many demultiplied targets so that static analysing and doc only builds on linux target with clang as an example.

MatthiasWM:
Thanks for implementing this. I decided to merge the file in at this point. We can still extend it later.

MatthiasWM:
Um, wait a minute.

MatthiasWM:
It's merged now.

fab672000:
Thanks!
I already started to use the static analyzer reporting there is a lot there and probably a fraction of the warnings we could decide to take actions on.

PR #17, opened 2019-01-21
Quote variables that may contain spaces

mid-kid:
This'll allow installing FLTK to directories that contain spaces.

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -79,59 +79,59 @@ depend:
 
 install: $(MANPAGES)
 	echo "Installing documentation files in $(DESTDIR)$(docdir) ..."
-	-$(INSTALL_DIR) $(DESTDIR)$(docdir)
+	-$(INSTALL_DIR) "$(DESTDIR)$(docdir)"
 	if test -f html/index.html ; then \
 	  for file in html/* ; do \
-		$(INSTALL_DATA) $$file $(DESTDIR)$(docdir); \
+		$(INSTALL_DATA) $$file "$(DESTDIR)$(docdir)"; \
 	  done \
 	fi
 	if test -f fltk.pdf ; then \
 	  echo "Installing fltk.pdf in $(DESTDIR)$(docdir) ..."; \
-	  $(INSTALL_DATA) fltk.pdf $(DESTDIR)$(docdir); \
+	  $(INSTALL_DATA) fltk.pdf "$(DESTDIR)$(docdir)"; \
 	fi
 	echo "Installing man pages in $(DESTDIR)$(mandir) ..."
-	-$(INSTALL_DIR) $(DESTDIR)$(mandir)/cat1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fluid.$(CAT1EXT) $(DESTDIR)$(mandir)/cat1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-config.$(CAT1EXT) $(DESTDIR)$(mandir)/cat1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-options.$(CAT1EXT) $(DESTDIR)$(mandir)/cat1
-	-$(INSTALL_DIR) $(DESTDIR)$(mandir)/cat3
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk.$(CAT3EXT) $(DESTDIR)$(mandir)/cat3
-	-$(INSTALL_DIR) $(DESTDIR)$(mandir)/man1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fluid.man $(DESTDIR)$(mandir)/man1/fluid.1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-config.man $(DESTDIR)$(mandir)/man1/fltk-config.1
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-options.man $(DESTDIR)$(mandir)/man1/fltk-options.1
-	-$(INSTALL_DIR) $(DESTDIR)$(mandir)/man3
-	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk.man $(DESTDIR)$(mandir)/man3/fltk.3
+	-$(INSTALL_DIR) "$(DESTDIR)$(mandir)/cat1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fluid.$(CAT1EXT) "$(DESTDIR)$(mandir)/cat1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-config.$(CAT1EXT) "$(DESTDIR)$(mandir)/cat1"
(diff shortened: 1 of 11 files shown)


Albrecht-S:
@mid-kid: Hmm, why would you want to do this (install in paths that contain spaces)? It is common sense not to do this because lots of software tool, libs, etc. have issues with this, and trying to do this is asking for troubles...

That, said, I checked your patch and it looks not bad; maybe a few pieces are missing, but I can't tell if this would really enable FLTK to be installed (and used) in paths with included spaces. I'm wondering if it's worth the effort...

mid-kid:
Well, I've encountered someone trying to install a tool that tells you to install a static version of fltk manually in its build instructions, who was trying to do so in a path that contained spaces (something like `/mnt/data/Jeux/GB ASM/etc...`), and refused to install it somewhere else, so we spent a while figuring this out, the result of that endeavour being this patch.

I've tested this, including with the `--enable-local{png,jpeg,zlib}` flags, installing and uninstalling, using `fltk-config` with a couple of different options, and I'm completely sure it works, at least with the autotools build system.

As for the patch itself, I've really just added quotes everywhere, which I believe is considered good practice™ anyway, and shouldn't be too hard to maintain, since people generally tend to mimic the style of surrounding code. That said, due to make only evaluating the command line after variable expansion (through `system()`, I reckon), there might still be issues with paths containing quotes, dollar signs, escape sequences, and/or whatever else the system's shell (and `sed` I guess?) chokes on. I don't think there's really a proper way around this, though I've found paths with spaces to appear more frequently than the other variants.

Whether it's worth the effort, I don't really know. Most of the changes won't break anything, at least, except maybe the fltk-config changes, which now quotes library paths in the output. I really have no clue if this is acceptable behavior. I have not checked if `pkg-config` does anything similar for paths with special characters. I don't mind splitting up/removing those changes.

MatthiasWM:
+1 on this one. I have seen al kinds of path names, and I wouldn't;t be surprised to find emoticons in file names, too (I just checked on MacOS, so, yes, it is possible, which means it's also happening). 

I saw that the OP also replaced some fixed paths `/usr/share/` to `$datadir`. I have not verified that, but that is of course a good idea, assuming the `$datadir` is set correctly.

mid-kid:
Thankfully emoji and any unicode characters aren't interpreted (have a special meaning) in any of the scripting languages that are used in the build system, so there's no problem in using those, and in fact it's safer to use some unicode spaces ([example](https://www.fileformat.info/info/unicode/char/00a0/index.htm)) rather than regular ascii spaces, for this reason.

The patch in OP would specifically fix regular space and tab characters, and depending on the system shell, question marks and pound signs. Characters like single and double quotes and backslashes are still problematic, but that'd require properly escaping everything in the makefile, which is not trivial.

Albrecht-S:
FYI: I added this PR to the "_Release FLTK 1.4.0_" milestone and will check this thoroughly when time permits and when I'm going to change other CMake related stuff. Promised.

ManoloFLTK:
Conflicts after recent changes in the master branch as of today have been resolved.

documentation/Makefile
```diff
@@ -79,59 +79,59 @@ depend:
...
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fluid.$(CAT1EXT) "$(DESTDIR)$(mandir)/cat1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-config.$(CAT1EXT) "$(DESTDIR)$(mandir)/cat1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-options.$(CAT1EXT) "$(DESTDIR)$(mandir)/cat1"
+	-$(INSTALL_DIR) "$(DESTDIR)$(mandir)/cat3"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk.$(CAT3EXT) "$(DESTDIR)$(mandir)/cat3"
+	-$(INSTALL_DIR) "$(DESTDIR)$(mandir)/man1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fluid.man "$(DESTDIR)$(mandir)/man1/fluid.1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-config.man "$(DESTDIR)$(mandir)/man1/fltk-config.1"
+	$(INSTALL_MAN) $(SRC_DOCDIR)/fltk-options.man "$(DESTDIR)$(mandir)/man1/fltk-options.1"
+  -$(INSTALL_DIR) "$(DESTDIR)$(mandir)/man3"
```
mid-kid, line 103: watch the indents

MatthiasWM:
That's great! Thank you!

Albrecht-S:
This issue is about autotools/configure and Makefiles only. CMake is not involved, hence I removed the labe CMake.

Also, this request is not urgent. Therefore I removed it from the release milestone. I'm not sure if this will (or should be) considered in the future after 1.4.0 got released because our mid-term plans are to drop autotools builds anyway.

I'm leaving this PR open anyway, just in case, but it has conflicts that need to be resolved anyway. Sorry to say, it's probably not worth trying to do this.

mid-kid:
Happy 5th anniversary!

Albrecht-S:
@mid-kid Sorry for the long delay - I was not sure what to do with your PR. Finally I decided to update and merge it.

_Reasoning:_ FLTK 1.4.x is the last release series that supports autotools (autoconf/configure/make), hence I decided to merge it in 1.4.0 although we recommend to use CMake starting with this release.

Albrecht-S:
FTR: unfortunately I noticed after the merge that this PR broke `fltk-config`, hence I reverted the quoting in `fltk-config.in` partially (commit 4077dfa629d43aedf445d828e809d1732205c6f7).

Anyway, I consider this PR resolved. If anybody feels inclined to fix quoting in `fltk-config` please open another issue with a patch or open a new PR. TIA.

Note: the [broken version](https://github.com/fltk/fltk/blob/b83a6abe01664245a2319c58b02063ba924fe912/fltk-config.in) of `fltk-config.in` can be found in commit b83a6abe01664245a2319c58b02063ba924fe912 as a starting point.

Hint: use `git show b83a6abe01664245a:fltk-config.in > fltk-config.in.broken` to check the broken file out.

mid-kid:
Whoa, thanks! I had long given up on using spaces in file paths since, but still cool that it made it in!

Albrecht-S:
FTR: commit f74f66e507569d93d5c9a289a080b951d2f10bf7 finally fixes quoting in fltk-config(.in).

Albrecht-S:
> this commit breaks something ...

@wcout Yes, indeed. I noticed (and fixed) it for `fltk-config --compile ...` by prepending `eval` which does an extra parser step. This can be a **workaround** for your command as well:
```
eval g++ -o demo `../fltk-config --cxxflags` demo.cxx `../fltk-config --ldstaticflags`
```
... works, but it's only a workaround as said above, not a real solution (we don't wanna break any users' working shell scripts or Makefiles).

I have no idea how to fix it if we want to apply quotes for paths with spaces.
FTR: I did a quick test by replacing single quotes with double quotes in `fltk-config --compile ...` but this didn't fix it.

Is there anybody with deeper shell knowledge who can propose a solution? This would be very much appreciated! TIA.

PR #28, opened 2019-05-11
Correct config installation path on macOS

Lasander:
Before:
install/FLTK/.framework/Resources/CMake/FLTKConfig.cmake
After:
install/FLTK.framework/Resources/CMake/FLTKConfig.cmake

See https://cmake.org/cmake/help/v3.0/command/find_package.html
for definitions of expected installation paths.


Albrecht-S:
Thanks for your PR. 
Looks correct, according to the CMake docs.

@ManoloFLTK, what do you think?

ManoloFLTK:
This change can be pushed in, because a directory hierarchy containing
      FLTK/.framework/Resources 
makes no sense. The intention is most certainly to create a directory called
FLTK.framework
and to put something in it.

However, I don't know what is the general intention with CMake creating FLTK.framework.
With the patch, and after
CMake
make; make install
one gets these files in that organization:
FLTK.framework
  Resources
    CMake
      FLTKConfig.cmake
      UseFLTK.cmake
      and 2 other .cmake files

A normal macOS framework is expected to contain a dynamically linked shared library
and the header files useful to build with that library. That's not at all what CMake + make install
produces at present for the content of directory FLTK.framework.
Thus, I'm not sure what to think.

ManoloFLTK:
`FLTK.framework
  Resources
    CMake
      FLTKConfig.cmake
      UseFLTK.cmake
      and 2 other .cmake files
`
Resources
CMake
FLTKConfig.cmake
UseFLTK.cmake
and 2 other .cmake files

ManoloFLTK:
The post above is unclear because it has lost its necessary indentation.
Here is a figure showing the content of FLTK.framework:

image

Albrecht-S:
@ManoloFLTK: I reopened this PR and merged the commit. No matter what contents the `FLTK.framework` directory would have now when populated by CMake `install` this seems to be at least the correct directory name (as you confirmed).

The intent of the correction was to make find_package() work better as it expects the named directory as given. Filling the framework directory with the full contents expected to be able to build an application is another task.

Thanks to @Lasander for the patch and to @ManoloFLTK for the comments.

PR #30, opened 2019-06-06
STR #275 fl_ask, fl_alert, fl_file_chooser, fl_* positioning

airbrett:
Added new positioning mode based on x,y point. add new function to set/get positioning modes.


Albrecht-S:
Thanks for this PR. Looks technically correct at a first glance but needs more thorough review(s).

It's an API change (extension) and as such it needs to be consistent with the FLTK API. For instance, setting a title for a common dialog works intentionally only for the next call of one of the general dialogs and resets itself after that. I'm not sure if your PR does this as well, but I think it should probably do it to be consistent (i.e. reset to`'hotspot'` after the next call).

As said above, needs review...

airbrett:
Thanks for considering my PR. You are right, the mode was not reset to hotspot after innards() was called. I reviewed the original code and I don't see a reset for the enableHotspot variable after innards(). I went ahead and made the change to reset as suggested (makes sense to me) but if the original code did not reset, this change could change the behavior of existing programs. What do you think?

airbrett:
Anything I can do to make this PR more palatable?

Albrecht-S:
I apologize for the long delay. Taking your comment of Jun 11 2019 into account I think that we should have an internal (static) variable that stores the `fl_message_hotspot(int)` setting (aka `enableHotspot` variable) and reset to this stored mode after calling innards() with one of your settings. This would be something like:
- `fl_message_hotspot(int)` is used as a permanent mode switch
- `fl_message_position(int,int,int)` is used as a one-shot function

But honestly, I'm not sure about the best API at all. The idea given above would mean that we'd have to keep the old `fl_message_hotspot(int)` API which would be bad. What do you think?

Also, we don't use `#include <cstddef>` - please use `#include <stddef.h>` or use code like `if (x)` instead of `if (x != NULL)` to avoid using `NULL` at all.

Albrecht-S:
PS: one question: how are you using the x/y position setting? Is it a fixed position, centered, related to the current window, anything else? I'm asking to get a better feeling what you would like to achieve and what would be the benefits of your proposal. TIA.

airbrett:
Thanks for the feedback. My motivation for this change is to tie dialog boxes more tightly to my application. For example when a fl_alert/fl_ask/etc dialog is shown, the dialog could be no where near my application and it may not be obvious to the user which application issued the dialog. I think this is exacerbated when in environments when there are multiple applications using FLTK for it's user interface or on a multi-monitor setup (dialog could be on an entirely different screen). Ideally I would position the fl_alert/fl_ask/etc dialog centered on my main window or on the middle of the monitor which my application is visible on.

My initial attempt of this change I had created an overridden function for each for position ex:
```
int fl_ask(const char *fmt, ...)
int fl_ask(const int x, const int y, const char *fmt, ...)
```
The nice part about that approach is that it's obvious that the position is temporary. This was a much more intrusive change, I wasn't sure how well that would be received so I reverted. What's missing from this (and the current PR) is there is no way to get the dialog size so centering would be tough.  Maybe a function similar to resizeform() could be written to get the size before calling fl_alert/fl_ask/etc?

Another idea I had was add a parameter for a window to center the dialog on. Ex
```
int fl_ask(const char *fmt, ...)
int fl_ask(Fl_Window* CenteredOn, const char *fmt, ...)
```

What do you think?

Albrecht-S:
A **quick** thought, what about this: to make things easier, let's keep the old API and add only one new call to set the preferred window position - to be called before the fl_*ask* (etc.) dialog and concerning only the next dialog popup. The API could be as proposed by you but w/o the `mode` argument (just `x` and `y`):
```
void fl_message_position(const int x = 0, const int y = 0);
int fl_message_position(int* x = NULL, int* y = NULL);
```
The next dialog would use the given x/y position unless ... _thinking_ ... both `x` and `y` are `-1` (or negative in general, or any magic value?). If the latter would be the case it would use the `enableHotspot` setting and - always - reset the `x/y` values to `-1` after the call. Or something like that, we might also use another status variable (position_set: 0|1).

This way `fl_message_position()` would always be temporary, have priority over `enableHotspot` and would not interfere with the `enableHotspot` setting at all.

FYI: I'm a little hesitant because there are other plans to modify the existing dialogs, see [STR 334](https://www.fltk.org/str.php?L334). It's been a long time since this STR has been posted and that I was working on it (2015) but I'm sure it can be done with an Fl_Message, Fl_Message_Box or Fl_Dialog class because I implemented such a dialog class in another project. The idea is to change the internal code but keep the fl_*ask* (etc.) dialog interface for compatibility. However, there would (should) also be a new API that would give the user more freedom to modify the dialogs as needed and build one's own dialogs easily (more or less). The problem with new functions as proposed in this PR is that they would all need to be supported by the new dialog classes and their API as well.

airbrett:
Sounds like a good approach to me. I attempted to implement the changes you described. If this isn't the time to make this change, that's fine.

Albrecht-S:
Thanks for the latest update. Looks good, I'm going to check it further and do some tests. I'll let you know about my progress.

Albrecht-S:
@airbrett Hi again, I took the freedom to use your PR, rebase and squash the commits, and add options to center the message box over
- the given coordinates
- a given widget or window
in another commit which is visible in the history.

The merge is in commit 4ab49d30d3, your commit(s) have been squashed and edited slightly by me. I hope you don't mind.

This PR is no longer directly applicable. Please test and comment if there's anything that doesn't work for you with this update - and feel free to close this PR yourself.

airbrett:
@Albrecht-S I think using -1 as the "not set" value for X and Y in
`int fl_message_position(int *x, int *y)`
could be a misleading. I don't know about other platforms, but in a multi display configuration Windows will use negative positions for windows located to the left and/or above of the primary screen. I found this great description https://stackoverflow.com/a/53026765 . For example if the programmer did this:
```
fl_message_position(-1,0);
```
Then called:
```
int x;
int y;
fl_message_position(&x, &y);

if (x == -1)
//Is X position not set or is it actually -1?
```
The X value would indicate that the position is not set(according to documentation) but in reality it was set to -1. This would apply to Y values also. I think the window would be positioned properly and the X/Y value returned from the function above would be correct however -1 would have dual meaning in this case. I don't know if this edge case is worth the effort to fix but I had a couple ideas:
A) Use INT_MAX as the "not set" value.
B) Not use a "not set" value since the return value indicates if X or Y contain valid data or not. Something like this:
```
  int fl_message_position(int *x, int *y) {
  if (form_position)
  {
       if (x)
         *x = form_x;
       if (y)
          *y = form_y;
     }
    return form_position;
  }
```
Regardless, for my use case your changes work great, I really like the center option!

Albrecht-S:
Hi @airbrett, thanks for your feedback. I'm aware of negative window positions, and that's one of the reasons why I added the **return value** on the getter function
```
  int fl_message_position(int *x, int *y)
```
Users should always use the _function return value_ to determine if the value is set (as [documented](https://fltk.gitlab.io/fltk/group__group__comdlg.html#gacc72f88c9c6ca5db5156e8dfff0cea41)). The additional "_returns -1 if not set_" for *x and *y are only so we have a defined value and not to say "_the x and y values are 'undefined'_" if not set. I could also have chosen 0 but I decided deliberately to use -1. I definitely wouldn't like using `INT_MIN` since this would require the user to `#include <limits.h>` which I wanted to avoid.

That said, I doubt that this getter function would be used much since setting the window position is a one-shot function (set position and call the dialog function), because it is reset after the call anyway. I can't imagine a real reason to call it at all...

PS: in the code (B) you posted the values of x and y would be "undefined", i.e. whatever these variables were initialized with. If the user used stack variables that would be random garbage, which I'd also like to avoid.

airbrett:
Sounds good to me, thanks for your help getting this change through!

silverduner:
I think this PR will not allow us to position a Fl_File_Chooser right? For Fl_File_Chooser positioning, IMHO, we can just expose `position()`, `x()`, `y()`, `w()` and `h()` from `Fl_File_Chooser.fl` for users to do the rest.

PR #35, opened 2019-09-19
enabling static compiling exe with mingw under linux

tpanj:
This is proposed solution.

Otherwise you will get error: 
> This application has failed to start because **libgcc_s_sjlj-1.dll** was not found. Re-installing the application may fix this problem.

### DIFF
--- a/CMake/options.cmake
+++ b/CMake/options.cmake
@@ -75,6 +75,9 @@ endif (UNIX)
 ##  May be removed once high-DPI support under Windows is complete.
 #######################################################################
 if (WIN32)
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Thanks for the proposal, but sorry, this patch is too drastic (linking everything static without options).

The correct way to fix your build is to specify -D"static-libgcc -static-libstdc++" on the CMake commandline or to edit `CMAKE_EXE_LINKER_FLAGS` accordingly.

We are considering an option to make this more user friendly or to include the options as written above as default, but not the entire static linking.

That said, I'm sorry but we can't accept your PR and I will close it now. Thanks for proposing it anyway.

PR #36, opened 2019-09-19
sudoku demo - cmake modification for setting icon of exe and window

tpanj:


erco77:
Adjusted the title to be more descriptive of the change being suggested.
Albrecht, I know you're working on the cmake stuff -- this seem OK syntax wise?

I can confirm the sudoku built with both mingw64 and VS 2017 **both don't show the sudoku icon in a file browser**,
but do show the window icon correctly (when you run the app, upper left corner).

I haven't tried the patch, but will in a bit. Just wanted to report the above.

Albrecht-S:
@erco77 Thanks for looking into it. Note that the given patch is hopelessly outdated. You may spend lots of time to get it working, I don't know if it's worth it. I can take a closer look tomorrow. 

Some hints though:
- CMake/macros.cmake does no longer exist, the macro is now in fl_create_example.cmake
- the macro create_example() MUST NOT include special cases based on target names (`${tname}`) like "sudoku"
- I don't think that the current macro would support ".rc" files but it could be extended (there's some parsing of file types which are given as source files, currently four types). That's where the ".rc" files would come into play
- the (contents of the) given (new) file `test/sudoku2.rc` should **replace** the (contents of the) existing file `sudoku.rc` (which is unnecessary large - lots of unused stuff), rather than adding another file
- `sudokurc.h` could supposedly be deleted

Still interested in doing it? If not I can take a stab at it tomorrow...

erco77:
I'll take a look -- if it's an easy fix I'll do it, otherwise I'll leave it to you.
Will follow up.

Albrecht-S:
One more note: the code
```
add_executable(${tname} WIN32 ${srcs} sudoku2.rc ${FLUID_SOURCES})
```
looks basically fine with the exception of the explicit file name. As soon as parsing of `.rc` files works they could simply be added to the `${srcs}` list variable instead of adding yet another variable. For CMake the `.rc` file is just another _source file_.

I'd appreciate if you could do it and test on Windows.

erco77:
I took a stab, and was surprised to find it worked, at least for mingw.
Ran out of time today to test with VS 2017, it's late here.

The window titlebar icon has already been correct, so no other changes needed for that AFAICT.
Just getting the .exe's icon is all that needs solving.

Attached a patch: [sudoku-patch.txt](https://github.com/fltk/fltk/files/4929818/sudoku-patch.txt)
Please check the cmake mod; no idea if it's proper. Seemed to make mingw work.

I don't know cmake file's syntax at all, just guessing at how the file worked, and modified accordingly.
Kinda surprised it worked the first time (makes me suspicious)

_[digression] I can't believe the cmake if/endif syntax needs the condition repeated verbatim in the endif,
what the heck is that?_

Also attaching "proof": a win32 browser screenshot showing the `sudoku.exe` with the proper icon.
sudoku-icon-in-win32-browser

erco77:
BTW, my patch works for VS2017 too. Beginner's luck?

Albrecht-S:
> Attached a patch: sudoku-patch.txt
Please check the cmake mod; no idea if it's proper. Seemed to make mingw work.

@erco77 I see your mods and I see it works, but this is the "old style" change, as the OP proposed, and what I wanted to avoid, see my previous post with hints:
> the macro create_example() MUST NOT include special cases based on target names (${tname}) like "sudoku"

There's nothing to say against @tpanj 's work, this was the only way at that time. We can still use their simplification of the `sudoku.rc` file.

Thanks to @tpanj, BTW.

Explanation: The "old style" macro was overloaded with lots of special cases like this one, the "new style" is to avoid all special cases in the macro, putting all necessary info on the command line. If you don't mind I'll go ahead and fix this issue which **needs** (as I found out after looking at the code) only a tiny change:
```
$ git diff -U0
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index df5903cd0..bd8123fcd 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -103 +103 @@ CREATE_EXAMPLE (subwindow subwindow.cxx fltk)
-CREATE_EXAMPLE (sudoku "sudoku.cxx;sudoku.icns" "fltk_images;fltk;${AUDIOLIBS}")
+CREATE_EXAMPLE (sudoku "sudoku.cxx;sudoku.icns;sudoku.rc" "fltk_images;fltk;${AUDIOLIBS}")
```
I didn't even need to change the `create_example` macro because the `else` case in the parser of the source files _does the right thing_:
```
  # filter input files for different handling (fluid, icon, plist, source)

  foreach (src ${SOURCES})
    if ("${src}" MATCHES "\\.fl$")
      list (APPEND flsrcs ${src})
    elseif ("${src}" MATCHES "\\.icns$")
      set (ICON_NAME "${src}")
    elseif ("${src}" MATCHES "\\.plist$")
      set (PLIST "${src}")
    else ()
      list (APPEND srcs ${src})
    endif ("${src}" MATCHES "\\.fl$")
  endforeach (src)
```
This was one of the **main goals** when I rewrote this macro: _No special cases, everything only to be changed in the arguments of the calling CMake file `test/CMakeLists.txt`_. I'm happy that this works so smoothly.

There's one thing to consider though: we add the `.rc` files unconditionally, we might have to think about other platforms that don't need nor can build that file. However, CMake should take care of this and ignore it on - say - the macOS platform. This needs to be tested, but I'm pretty sure this is true. It worked fine on Linux.

Still need to test this with VS 2019 in my VM or maybe even VS 2010 on my old Windows box.

Albrecht-S:
@erco77 @tpanj I committed c380a91361db381747339a02e808e1b6850f7ce2 which fixes the issue properly in our new CMake system (as described above).

@tpanj Thank you very much for your proposal, and sorry for the long delay to fix it. I could not use your PR directly, so I committed my own fixes.

FTR: tested on Windows 10 with VS 2019 as well.

Albrecht-S:
@erco77 @tpanj Can you confirm the fix?

This PR will be closed either when confirmed or after 5 days, whichever comes first.

Albrecht-S:
Addendum, reply to this comment: https://github.com/fltk/fltk/pull/36#issuecomment-659217358
> BTW, my patch works for VS2017 too. Beginner's luck?

No, probably not. CMake is a tool to write **cross platform** build files. In most cases one CMake file can be used w/o changes on many platforms.

Some platform specific code may be necessary though, but I believe that we can remove some of the platform specific `if | then | else` clauses because CMake would simply ignore the code on other platforms. For instance we could presumably use `macOS bundle` code unconditionally on all platforms, and I know that the `WIN32` attribute for `add_executable()` is ignored on other platforms (it's documented).

erco77:
Glad you were able to fix it up.

Confirmed your mod:  I pulled the latest from git, see your new sudoku commit, rebuilt vs 2017,
and the sudoku executable has the icon, so looks good to me.

Also confirmed with mingw64.

So I think you can close if you're just waiting for my confirmation.

> @erco77 I see your mods and I see it works, but this is the "old style" change,
> as the OP proposed, and what I wanted to avoid, see my previous post with hints:

I didn't understand most of the hints (as I don't understand cmake yet), and you were asleep so I couldn't ask,
but I did see your followup which made sense:
```
One more note: the code
add_executable(${tname} WIN32 ${srcs} sudoku2.rc ${FLUID_SOURCES})
looks basically fine with the exception of the explicit file name.
```
..so I took that "exception of the explicit filename" to mean the "sudoku2.rc" was wrong, so I removed the "2",
and just tried to figure out where to insert that line. Didn't know that could all be avoided, but was just surprised
that it worked anyway.

Albrecht-S:
Yep, first steps with CMake are not easy, it takes some time to understand the concept, and then to work with its features and syntax. Not always intuitive, and the development of CMake is quick. Every now and then new versions and features.

I'd like to wait for @tpanj  's confirmation as well.

tpanj:
I am glad you found this useful!
Currently enhanced modification at **fltk:master** works great on Windows 7 with VS 2019.

Albrecht-S:
Thanks for your feedback, it's appreciated

PR #47, opened 2020-01-07
Fluid: when linking include LDFLAGS last

michaelld:
LDFLAGS often contains "-L*" flags to tell the linker where to search for libraries. If these are included -before- the ones used for internal linkage, and FLTK is already installed on the system, then the already-installed FLTK libraries can be found before those internal to the build. Moving the LDFLAGS to the end generally solves this issue. Worst case the move won't hurt linking since all of the flags are still included in the link command.

### DIFF
--- a/fluid/Makefile
+++ b/fluid/Makefile
@@ -55,7 +55,7 @@ all:	$(FLUID) fluid$(EXEEXT)
 fluid$(EXEEXT):		$(OBJECTS) $(LIBNAME) $(FLLIBNAME) \
 			$(IMGLIBNAME)
 	echo Linking $@...
-	$(CXX) $(ARCHFLAGS) $(CXXFLAGS) $(LDFLAGS) -o $@ $(OBJECTS) $(LINKFLTKFORMS) $(LINKFLTKIMG) $(LDLIBS)
+	$(CXX) $(ARCHFLAGS) $(CXXFLAGS) -o $@ $(OBJECTS) $(LINKFLTKFORMS) $(LINKFLTKIMG) $(LDFLAGS) $(LDLIBS)
 	$(OSX_ONLY) $(RM) -r -f fluid.app/Contents/MacOS
 	$(OSX_ONLY) mkdir fluid.app/Contents/MacOS
 	$(OSX_ONLY) $(INSTALL_BIN) fluid fluid.app/Contents/MacOS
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Thanks for the proposal, and please apologize the long delay.

I believe I can follow you - partially.

> LDFLAGS often contains "-L*" flags to tell the linker where to search for libraries.

What do you mean with "often"? Where do these flags "often" come from? Are you maybe referring to the environment variable LDFLAGS as set by the user and applied to the build by means of `configure` and generated Makefiles or, in the case of FLTK, the `makeinclude` file? Please elaborate.

In a default build (on my Linux system) I don't see any "special" flags in LDFLAGS, but that is obviously caused by not setting anything.

Back to that _"I can follow you - partially"_ statement: I learned that link command and flags order matters, and that's the part I can follow. However, `./configure --help` says:
```
  LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
              nonstandard directory <lib dir>
  LIBS        libraries to pass to the linker, e.g. -l<library>
```
I believe (without actually verifying) that `LIBS` would be propagated to the internal (makeinclude) variable LDLIBS and should be placed in the linker command last (after LDFLAGS).

Conclusion: I think that your patch is not bad but should probably invert the order of LDFLAGS and LDLIBS so that LDFLAGS comes first. This is an excerpt from `makeinclude` of an "empty" (default) configuration:
```
OPTIM		=  -Os -Wall -Wunused -Wno-format-y2k  -fno-exceptions -fno-strict-aliasing -ffunction-sections -fdata-sections
LDFLAGS		= $(OPTIM)  
LDLIBS		= -lXrender -lXcursor -lXfixes -lXext -lXft -lfontconfig -lXinerama -lpthread -ldl -lm  -lX11 
```

I included `OPTIM` as well since it is used to initialize `LDFLAGS`. For me it would make perfectly sense to have `$(LDFLAGS) $(LDLIBS)` at the **end** of the linker command line but not, as you propose `$(LDLIBS) $(LDFLAGS)`.

If you could change this I think your proposal can be accepted - or did I miss anything?

If yes, please explain what I missed and why.

If not, could you update your PR, or should I just commit your proposal as modified by me above?

michaelld:
Thanks for working on this PR! It’s been long enough that I have to do a build of fltk to re-verify both the need as well as the fix. I will strive to do that tomorrow (US-Eastern).

Albrecht-S:
Thanks for your feedback, and TIA for testing again.

michaelld:
Yes what you propose works and makes sense. Either way works, as it turns out. This is because of how the variables are determined during configuration. I'll update the PR accordingly.

michaelld:
Done. Thanks for the suggestion! I tried to clarify the commit message to be a bit more precise, too.

michaelld:
Rebased to the current master HEAD.

Albrecht-S:
Thanks. Merged.

PR #48, opened 2020-01-16
Fix JIS Keyboard dead keys

informationsea:
Underscore key and Yen key do not work properly in Japanese keyboard. This patch fixes the problem.

### DIFF
--- a/src/drivers/Cocoa/Fl_Cocoa_Screen_Driver.cxx
+++ b/src/drivers/Cocoa/Fl_Cocoa_Screen_Driver.cxx
@@ -249,6 +249,7 @@ void Fl_Cocoa_Screen_Driver::insertion_point_location(int x, int y, int height) 
 int Fl_Cocoa_Screen_Driver::compose(int &del) {
   int condition;
   int has_text_key = Fl::compose_state || Fl::e_keysym <= '~' || Fl::e_keysym == FL_Iso_Key ||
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Thanks for this PR and for PR #49 as well. The patches are looking sensible, but I have one question:
What about FL_Kana and FL_Eisu? Would these keys have to be handled as well?

See Enumerations.H:
```
FL/Enumerations.H-459-#define FL_Kana         0xff2e  ///< The Kana key of JIS keyboards.
FL/Enumerations.H-460-#define FL_Eisu         0xff2f  ///< The Eisu key of JIS keyboards.
FL/Enumerations.H:461:#define FL_Yen          0xff30  ///< The Yen key of JIS keyboards.
FL/Enumerations.H-462-#define FL_JIS_Underscore 0xff31 ///< The underscore key of JIS keyboards.
```
If yes we could maybe handle this entire range (0xff2e .. 0xff31). Just a thought.

informationsea:
FL_Kana and FL_Eisu are used to switch between input method mode instead of entering text. These key works correctly in current code, so should not be handled in `compose` function.

Albrecht-S:
@informationsea: Thanks for clarifying this.

I believe we can merge both PR's (#48 and #49).
@ManoloFLTK, what do you think?

ManoloFLTK:
This contribution from a Japanese user is precious and must be committed.

Albrecht-S:
Done, thanks to @informationsea and @ManoloFLTK.

PR #49, opened 2020-01-16
Fix JIS Keyboard dead keys (branch 1.3)

informationsea:
Underscore key and Yen key do not work properly in Japanese keyboard. This patch fixes the problem in branch 1.3.


Albrecht-S:
See also related PR #48.

Albrecht-S:
Done, thanks to @informationsea and @ManoloFLTK.

PR #54, opened 2020-02-09
New Fl_ICO_Image class

darealshinji:
A class to read Windows icons. Unfortunately I had to do some changes in Fl_BMP_Image and Fl_PNG_Image.


darealshinji:
This one might also be useful as a resource for the format: http://vitiy.info/manual-decoding-of-ico-file-format-small-c-cross-platform-decoder-lib/
Or perhaps some MSDN documentation or MinGW header file.

I'll see if I can make it a bit more flexible, allowing to manually parse the ICONDIRENTRY stuff through some class methods. Though I thought it was most useful to pick the highest resolution icon.

darealshinji:
I've added the ability to load a specific image from the icon file and to only load the `ICONDIRENTRY` stuff, so you can access and parse it yourself. Setting `id` to 0 will load the first image, 1 will load the next, etc. Setting it to -1 will load the biggest/best image (default). -2 will only load the `ICONDIRENTRY`.

Here's an example:
``` C++
#include <FL/Fl.H>
#include <FL/Fl_Box.H>
#include <FL/Fl_ICO_Image.H>
#include <FL/Fl_Double_Window.H>

int main()
{
  Fl_ICO_Image *ico = new Fl_ICO_Image("icon.ico", NULL, -2);
  Fl_Double_Window *win = new Fl_Double_Window(300, 400, "Test");
  Fl_Box *b = new Fl_Box(0, 0, win->w(), win->h());

  printf("icon count: %d\n", ico->idcount());

  int id = 3;
  printf("icon resource #%d offset: %d\n", id, ico->icondirentry()[id].dwImageOffset);

  delete ico;
  ico = new Fl_ICO_Image("icon.ico", NULL, id);

  if (!ico->fail()) {
    b->image(ico);
  }
  win->end();
  win->show();

  return Fl::run();
}
```

Here's an icon file to test: [icon.ico.zip](https://github.com/fltk/fltk/files/4181300/icon.ico.zip)

Albrecht-S:
Thanks for the update and the demo program. As far as I could test everything looks fine.

There's one question though: (why) do we need another private member variable Fl_BMP_Image::desired_h_ ? AFAICT there's only place where this is set and used in` load_bmp_()`:
```
void Fl_ICO_Image::load_ico_(Fl_Image_Reader &rdr, int id)
{
  ...
  rdr.seek(icondirentry_[pickedID].dwImageOffset);
  desired_h(h());
  load_bmp_(rdr, 1);
}
```
I think that one more (optional) parameter in load_bmp_() would be a better choice because this does not add anything to the memory footprint of `Fl_BMP_Image`. But maybe I'm missing something. Opinions?

darealshinji:
Like I wrote here in the comment:
``` C++
    // When reading from a .ICO file we must be able to force a specific height.
    // I guess the mask is sometimes not correctly recognized. There must be a better way...
```
Yeah, maybe add another parameter to load_bmp_().

darealshinji:
Added is as optional parameter to `load_bmp_()`.

src/Fl_BMP_Image.cxx
```diff
@@ -164,7 +166,7 @@ void Fl_BMP_Image::load_bmp_(Fl_Image_Reader &rdr)
     // If the height is negative, the row order is flipped
     temp = rdr.read_long();
     if (temp < 0) row_order = 1;
-    h(abs(temp));
+    w(ico_height > 0 ? ico_height : abs(temp));
```
darealshinji, line 169: I guess this should be `h()` instead of `w()`? I think for ICO files we need to rely on the width and height data that was read from the ICO header and ignore the info from the bitmap data. So pass the width data as an argument too?
``` C++
if (ico_height > 0 && ico_width > 0) {
  rdr.read_long();
  rdr.read_long();
  w(ico_width);
  h(ico_height);
} else {
  // New BMP header...
  w(rdr.read_long());
  // If the height is negative, the row order is flipped
  temp = rdr.read_long();
  if (temp < 0) row_order = 1;
  h(abs(temp));
}

File: src/Fl_ICO_Image.cxx [hunk 2]
```diff
@@ -0,0 +1,189 @@
...
+    icondirentry_[i].wPlanes = (int)rdr.read_word();
+    icondirentry_[i].wBitCount = (int)rdr.read_word();
+    icondirentry_[i].dwBytesInRes = (int)rdr.read_dword();
+    icondirentry_[i].dwImageOffset = (int)rdr.read_dword();
+
+    if (icondirentry_[i].bWidth == 0) icondirentry_[i].bWidth = 256;
+    if (icondirentry_[i].bHeight == 0) icondirentry_[i].bHeight = 256;
+  }
+
+  if (id <= -2) return;
```
darealshinji, line 98: Adding a note that this return here is done to only read the `icondirentry_` data without the actual image data would be useful.

src/Fl_PNG_Image.cxx
```diff
@@ -131,6 +132,9 @@ void Fl_PNG_Image::load_png_(const char *name_png, const unsigned char *buffer_p
       ld(ERR_FILE_ACCESS);
       return;
     }
+    if (offset > 0 && fseek(fp, (long)offset, SEEK_SET) == -1) {
+      rewind(fp);
+    }
```
darealshinji, line 137: Should the function return here instead of calling `rewind()` and ignoring the `fseek()` error?

Albrecht-S:
@darealshinji Thanks for your review, the comments look correct. ISTR that I found the w/h mismatch when I initially worked on this as well but I need to look closer into it again. When time permits. Sorry for the delay.

ManoloFLTK:
I've taken this code, updated it relatively to today's FLTK state (401b3dc), included remarks given above dated Sept 13, 2021, added Doxygen comments, added the required Fl_ICO_Image destructor, made struct ICONDIRENTRY a member struct of class Fl_ICO_Image, and made a new, private constructor of class Fl_PNG_Image used by class Fl_ICO_Image, rather than changing the public Fl_PNG_Image constructor.

Function fl_check_images()  has been slighlty enhanced to recognize .ico files, and demo program test/pixmap_browser has been changed to show .ico files by default. I've seen the demo icon.ico file mentionned above is correctly processed. I've also seen another, attached, .ico file which internally uses the BMP format as opposed to the PNG format used by icon.ico, is processed OK.

The result is in branch ICO_Image of [my FLTK fork](https://github.com/ManoloFLTK/fltk).

@Albrecht-S could you, please, review this code and, possibly give your OK for commit ?

@darealshinji  What would be the most correct way to identify a .ico file from its first bytes as needed by function fl_check_images() to recognize such files? I've used this condition
`if (memcmp(header, "\0\0\1\0", 4) == 0 && header[5] == 0) ` where `header` contains the first few bytes of the file.
Also, could you, please, update your FLTK fork and Fl_ICO_Image branch with this code, and give any comment of yours? That would allow the final commit to be done with you as author.

[seaview.ico.zip](https://github.com/fltk/fltk/files/10105873/seaview.ico.zip)

Albrecht-S:
Thanks, @ManoloFLTK, I read your message and will try to test this tomorrow.

PR #55, opened 2020-02-10
Fl_Window::default_icons(): set icons automatically from exe

darealshinji:


src/Fl_win32.cxx
```diff
@@ -2424,6 +2424,22 @@ void Fl_Window::default_icons(HICON big_icon, HICON small_icon) {
...
+  char path[FL_PATH_MAX];
+
+  big_icon_array[0] = NULL;
+  small_icon_array[0] = NULL;
+  big_icon = big_icon_array;
+  small_icon = small_icon_array;
+
+  // use exe's/module's icon resource to set window default icons
+  GetModuleFileNameA(NULL, path, FL_PATH_MAX);
+  ExtractIconExA(path, 0, big_icon, small_icon, 1);
```
Albrecht-S, line 2438: Wouldn't it better to use `GetModuleFileNameW()` and `ExtractIconExW()` for international filename support? I'm not sure if it would work anyway, but I'd feel more comfortable with the wide character (`W`) versions.

Albrecht-S: ... and isn't there another way to access the resource directly (in memory) w/o accessing the file on disk as `ExtractIconExA(path, ...)` seems to do (I may be wrong though, please correct me).

Albrecht-S:
> No description provided.

@darealshinji Thanks for proposing this. A better description would likely have helped to check this proposal earlier. ;-)

I looked at the source code and I believe I understood the intention, please correct me if I'm wrong:

- this function (your PR) is for Windows only
- a programmer (below: user) needs to add a **Windows icon resource** by the usual means (.rc file etc.) to the program
- this icon would be displayed in the Windows explorer, right?
- the user must also call `Fl_Window::default_icons()` to activate this feature
- would this automatically display the icon?
- the user can still call any of the other Windows icon setting methods afterwards
- if done so, the icon established last would be displayed

With "display the icon" I mean in the window frame and in the task bar.

Note that I didn't test your PR (yet). Please elaborate if anything of the above statements is wrong and answer the questions.

It would also be helpful if you could post a small demo program that would utilize this feature. No need to post the icon or the resource file, just the main program file. You can start with test/hello.cxx or post a patch or instructions on how to modify it. TIA.

src/Fl_win32.cxx
darealshinji: Using the wide character versions might be better, I guess. I don't know if there's a better way to access the resource.

darealshinji:
To answer the questions:
- yes
- yes
- the icon visible in Windows Explorer (file browser) for the .exe gets used in taskbar and window bar
- yes
- icon is displayed if `Fl_Window::default_icons()` was called and a resource was found
- yes, for example you change the icon resource of the compiled .exe with a tool and the new icon will be used on next startup
- not sure what you mean; it works like all the other variants of `Fl_Window::default_icon`

An icon added as a resource gets displayed in the Explorer but it's not used in taskbar and window bar. In FLTK programs I always had to manually add an icon with `Fl_Window::default_icon()` which meant I've added another resource (usually a PNG file). For convenience and to reduce size I came up with this code, which picks the .exe resource icon and uses it as default icons:
``` c++
/* use exe's icon resource to set window default icons */
wchar_t mod[MAX_PATH_LENGTH];
HICON hIconL[1] = { 0 };
HICON hIconS[1] = { 0 };
HICON *phIconL = hIconL;
HICON *phIconS = hIconS;
GetModuleFileNameW(NULL, mod, MAX_PATH_LENGTH);
ExtractIconExW(mod, 0, phIconL, phIconS, 1);
Fl_Window::default_icons(hIconL[0], hIconS[0]);
```
And I thought it would be neat if this could be simplified to just calling `Fl_Window::default_icons()`.

Albrecht-S:
Thanks for your answers and the explanations. We need to ensure that we keep cross platform conformity, hence we need a default action (likely do nothing) on other platforms.

Maybe we can add a similar function on macOS where icons can exist inside bundles, but I don't know enough of macOS.

I'll do some tests and look at the internals. I'll keep you updated.

darealshinji:
I've changed it to use the wide character versions of the Windows API functions. I've also changed `sudoku.cxx` to use the resource instead of the bitmap, if you don't mind.

darealshinji:
@MatthiasWM The icon is linked to the .exe file through `sudoku.rc` at this part of the CMake code:
``` cmake
if(WIN32)
CREATE_EXAMPLE(sudoku "sudoku.cxx;sudoku.rc" "fltk;fltk_images;${AUDIOLIBS}")
else()
CREATE_EXAMPLE(sudoku sudoku.cxx "fltk;fltk_images;${AUDIOLIBS}")
endif(WIN32)
```

FL/Fl_Window.H
```diff
@@ -351,6 +351,7 @@ public:
   // These 2 member functions break the driver model but are kept for back compatibility.
   // They are implemented in Fl_win32.cxx
   static void default_icons(HICON big_icon, HICON small_icon);
+  static void default_icons();
   void icons(HICON big_icon, HICON small_icon);
```
Albrecht-S, line 355: It is clearly said in the comment above: "These 2 member functions break the driver model but are kept for back compatibility."
Changing 2 to 3 could NOT be accepted, we should definitely not add new methods to those "kept for back compatibility".

Albrecht-S:
> Since the Fl_Window::default_icons(icon, icon) interface is already limited to WIN32, I don't see an issue with adding this function to only one platform.

@MatthiasWM Please see my comment above. The existing functions are only for backwards compatibility. We should not add more overloaded versions of those.

I'm not sure how to do this in a consistent way but I would suggest to add a new method name (maybe `use_default_icons()` ?) that could be used on all platforms to avoid '#if ...' in user code. As I wrote [before](https://github.com/fltk/fltk/pull/55#issuecomment-665137807) I can imagine that such a method would use/load a macOS icon file from the bundle. A first version could be a no-op on other platforms.

Albrecht-S:
@darealshinji The code regarding sudoku.rc has been significantly changed, even in branch-1.3. From which version is the code snippet you posted?

Anyway, a quick search in current 1.4 (git) reveals (CMake + Makefile):
```
test/CMakeLists.txt:107:CREATE_EXAMPLE (sudoku "sudoku.cxx;sudoku.plist;sudoku.icns;sudoku.rc" "fltk_images;fltk;${AUDIOLIBS}")
...
test/Makefile:477:sudoku.exe: sudoku.o sudoku.rc
test/Makefile-478-	echo Linking $@...
test/Makefile:479:	$(RC) sudoku.rc sudokures.o
test/Makefile-480-	$(CXX) $(ARCHFLAGS) $(CXXFLAGS) $(LDFLAGS) sudoku.o sudokures.o -o $@ $(AUDIOLIBS) $(LINKFLTKIMG) $(LDLIBS)
```

Hence I agree that no code changes would be required if it works as described.

Albrecht-S:
> @darealshinji The code regarding sudoku.rc has been significantly changed, even in branch-1.3. From which version is the code snippet you posted?

Sorry, answering my own question: I pulled your branch and now I see that this code is from your patch. Looking into it...

darealshinji:
A new method name sound okay. On Mac you'd have to use the CF* functions to find out if you're in a bundle, then look for a default png icon in the App bundle resources. But I don't have a Mac system to test this.

PR #61, opened 2020-03-13
Modified the dist-clean target.

MichaelTiernan:
Added a couple of small items that were left behind from the configure/build process which kept the dist-clean target from actually returning you to the distribution "clean" state.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,10 @@ distclean: clean
 		$(RM) test/`basename $$file .fl`.cxx; \
 		$(RM) test/`basename $$file .fl`.h; \
 	done
+	$(RM) -rf autom4te.cache/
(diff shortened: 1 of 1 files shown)


erco77:
Mods look OK to me, but I don't think it's been applied yet.  Why was it closed?
Not sure why the automatic build failed; seems unrelated to the code changed.
Other devs: comments?

MichaelTiernan:
I have to double check my work.
I did a few things yesterday and I think I killed this 'cause I did the merge incorrectly.

Albrecht-S:
The build failed because the author @MichaelTiernan deleted the branch in his fork.

See also the new PR #63. Note: the new commit message indicates that the `make clean` target was changed but indeed it was the `make distclean` target.

PR #63, opened 2020-03-14
Modified make clean target.

MichaelTiernan:
Added items that are left around by config/make process.
This is also a correction to an earlier pull request.


Albrecht-S:
Thanks for this PR. I merged it in commit d9f7a368a6e997106f17abdac247e682a3642f1c, but I edited the commit message (s/clean/dist-clean/) ... which, after checking again, is not quite correct: should read 'distclean', sorry.

PR #71, opened 2020-04-14
Fix int highlight() comment/doc in Fl_Text_Buffer

MoAlyousef:
The function might also be renamed to highlighted, like selected.


Albrecht-S:
Thanks for the fix.

I merged it after rebasing. I hope you don't mind.

PR #72, opened 2020-04-18
Make draw method protected in Fl_Wizard

MoAlyousef:
It is currently private, making it protected would allow custom drawing on the wizard widget


Albrecht-S:
Thanks for the patch, you're absolutely right. All (widget) draw methods should be protected rather than private.

PR #78, opened 2020-05-13
Fix "misleading indentation" in Fl_Text_Buffer.cxx

newtonallen3:
Compiling with clang produced the following warning on this line:

src/Fl_Text_Buffer.cxx:1292:5: warning: misleading indentation; statement is not part of the previous 'if' [-Wmisleading-indentation]
    replace(start, end, text);
    ^
src/Fl_Text_Buffer.cxx:1288:3: note: previous statement is here
  if (!sel->position(&start, &end))
  ^


newtonallen3:
Yes, line numbers were from an old version (1.3.5). Thanks!

PR #80, opened 2020-05-17
expose position(int, int), x(), y(), w() and h() from Fl_File_Chooser

silverduner:
To give users the ability to position the dialog.

### DIFF
--- a/FL/Fl_File_Chooser.H
+++ b/FL/Fl_File_Chooser.H
@@ -168,6 +168,11 @@ public:
   const char *value(int f = 1); 
   void value(const char *filename); 
   int visible();
+  void position(int x, int y);
+  int x();
+  int y();
+  int w();
+  int h();
   /**
    [standard text may be customized at run-time]
   */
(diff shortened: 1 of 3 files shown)


MatthiasWM:
I'd be happy to accept this PR, but I would like to ask for the addition of size(w, h) and resize(x, y, w, h). Otherwise, asking for w() and h() does not make much sense. Also, when explicitly setting the position, the hotspot() should be disabled, or FLTK will position the File Chooser under the mouse pointer.

gvnnz:
Also, should all the getter methods be marked as `const`? E.g. `int x() const;`, `int y() const;`, ...

Albrecht-S:
> should all the getter methods be marked as const?

Yes, please!

@gvnnz Thanks for the heads-up.

silverduner:
Hi, glad to see some responses on this PR. Just updated as suggested.

Albrecht-S:
Thanks for the update. Patch (PR) looks good but I didn't test. @MatthiasWM ?

MatthiasWM (approved):

MatthiasWM:
Thanks for the patch. Finally I am no longer limited to the puny little default file chooser ;-)

silverduner:
> Thanks for the patch. Finally I am no longer limited to the puny little default file chooser ;-)

Indeed :). Thanks for merging it.

PR #86, opened 2020-05-31
fix mousewheel simultaneous X and Y scrolling under OS X

elaforge:
Previously, fltk would only emit either a X mousewheel event or a Y one,
never both at once, and preferring X.  This meant any horizontal scrolling
would cause vertical scrolling to jam until the horizontal momentum stopped.

### DIFF
--- a/src/Fl_cocoa.mm
+++ b/src/Fl_cocoa.mm
@@ -929,28 +929,16 @@ static void cocoaMouseWheelHandler(NSEvent *theEvent)
     return;
   }
   Fl::first_window(window);
-  
-  // Under OSX, single mousewheel increments are 0.1,
-  // so make sure they show up as at least 1..
-  //
-  float dx = [theEvent deltaX]; if ( fabs(dx) < 1.0 ) dx = (dx > 0) ? 1.0 : -1.0;
-  float dy = [theEvent deltaY]; if ( fabs(dy) < 1.0 ) dy = (dy > 0) ? 1.0 : -1.0;
-  if ([theEvent deltaX] != 0) {
-    Fl::e_dx = (int)-dx;
-    Fl::e_dy = 0;
-    if ( Fl::e_dx) Fl::handle( FL_MOUSEWHEEL, window );
-  } else if ([theEvent deltaY] != 0) {
-    Fl::e_dx = 0;
-    Fl::e_dy = (int)-dy;
-    if ( Fl::e_dy) Fl::handle( FL_MOUSEWHEEL, window );
-  } else {
-    fl_unlock_function();
-    return;
+
+  // Under OSX, mousewheel deltas are floats, but fltk only supports ints.
+  int dx = roundf([theEvent deltaX]);
+  int dy = roundf([theEvent deltaY]);
+  if (dx || dy) {
+    Fl::e_dx = -dx;
+    Fl::e_dy = -dy;
+    Fl::handle(FL_MOUSEWHEEL, window);
   }
-  
   fl_unlock_function();
-  
-  //  return noErr;
 }
 
 /*


Albrecht-S:
Hmm, besides fixing the x/y mousewheel issue as documented in the commit message, this commit does also change the behavior of the mousewheel handling in cases where the absolute value of mousewheel increments is smaller than 0.5, i.e. the old code raised this up to 1 (or down to -1) instead of rounding it to 0 as the proposed code would do.

I'm explicitly not saying that one or the other is wrong, it's just an observation, leaving this to discuss and decide by people who know more about macOS behavior and can test (@ManoloFLTK ?).

Question: what are "usual" mousewheel values, are they in the range < +/- 1 or are they (much) larger? This might make a difference for the decision what to use.

Other than that: the proposed code looks good and pretty simple.

One more note: ISTR that I saw  (probably under Windows) that a static variable was used that accumulated all mousewheel increments and used only the usable (maybe integer) part, or something like that. Would be interesting to think about such a solution which would (a) round and (b) not lose fractional parts. Just a thought...

elaforge:
That's true, I should have mentioned that difference.  I tried it out and it still feels fine to me... it means that the <0.5 tail is clipped off.  In theory it should lead to a more natural stop because it more closely approximates what the OS is trying to send rather than rounding up even at 0.001, but I honestly can't feel a difference.

Mousewheel values will go up to arbitrary numbers depending how fast you scroll.  I saw peaks at 25 or so for pretty quick scrolling.  It's common to do a flick to jump to the top/bottom of something, so this makes sense.

I suppose if we store the roundoff it would more closely approximate the integral of the curve, but it seems like you'd get weird behaviour.  I can imagine it would go to 0, but continue accumulating fractions, and then get another little pixel nudge after it stopped.  Whereas now it may stop a bit suddenly, but at least it stays stopped. Also the fraction would stick around forever unless you put some timeout, meaning your next scroll has to cancel out the tiny bit of momentum left over from the one 5 minutes ago, which makes no sense.  Probably none of it makes a difference you can feel though, because the deacceleration curve seems to be sharp enough that there isn't a very long tail of <1 fractions... maybe just 1 or 2 "frames" worth.  This also makes sense because you don't want your UI to keep jiggling for too long after you poke it.

Albrecht-S:
Thanks for the PR and the quick reply with details. I think I'm +1 on the PR - unless @ManoloFLTK has doubts regarding the rounding.

ManoloFLTK:
I agree with the intention to simultaneously process horizontal and vertical
wheel movements. However, that should be done consistently in FLTK; namely
it's necessary to consider how class Fl_Text_Display processes the 
FL_MOUSEWHEEL event:
    case FL_MOUSEWHEEL:
      if (Fl::event_dy()) return mVScrollBar->handle(event);
      else return mHScrollBar->handle(event);
That deserves close scrutiny because that code is not platform-specific and
because there's room for only one return value.

This PR makes me realize function cocoaMouseWheelHandler(NSEvent *theEvent)
has not been modified to reflect screen scaling of branch 1.4. I'll do that.

ManoloFLTK:
I propose this modified code :
`
static void cocoaMouseWheelHandler(NSEvent *theEvent)
{
  // Handle the new "MightyMouse" mouse wheel events. Please, someone explain
  // to me why Apple changed the API on this even though the current API
  // supports two wheels just fine. Matthias,
  fl_lock_function();
  Fl_Window *window = (Fl_Window*)[(FLWindow*)[theEvent window] getFl_Window];
  Fl::first_window(window);
  // Under OSX, mousewheel deltas are floats, but fltk only supports ints.
  //
  float s = Fl::screen_driver()->scale(0);
  int dx = roundf([theEvent deltaX] / s);
  int dy = roundf([theEvent deltaY] / s);
  if (dx) {
    Fl::e_dx = -dx;
    Fl::e_dy = 0;
    if (Fl::e_dx) Fl::handle( FL_MOUSEWHEEL, window );
  }
  if (dy) {
    Fl::e_dx = 0;
    Fl::e_dy = -dy;
    if (Fl::e_dy) Fl::handle( FL_MOUSEWHEEL, window );
  }
  fl_unlock_function();
}

`

elaforge:
Good catch, I hadn't considered the fact that client code may be written to only expect one or the other.  Your change looks good to me, but could you add a comment explaining the either/or behaviour?  So the next person isn't tempted to "fix" it by merging them.

ManoloFLTK:
@elaforge : please, close this PR if the committed fix (f9a63b5) is satisfactory.

Albrecht-S:
@elaforge wrote in reply to @ManoloFLTK's comment:
> Good catch, I hadn't considered the fact that client code may be written to only expect one or the other.

I believe that this is from the good old times when we didn't have devices that could generate simultaneous scrolling on both x and y axis. I think that we'll need to rewrite the widget internals in the long term to use both e_dx and e_dy in one FL_MOUSEWHEEL event. But that's not relevant for now and should not prevent closing this PR. Just a comment for future development...

elaforge:
Oh it's already merged outside of the PR.  Ok, looks good to me.

PR #94, opened 2020-06-21
Quick fix for inability to link with pango from fltk-config

Taeril:
Commit 6fe226cb804d000b29ea53e08acc505267fd44de introduced use of pkg-config which broke linking from fltk-config if FLTK was configured to use pango library.

This patch duplicates line from another if branch that just adds libraries assuming that if pkg-config found pangoxft that there are present all it's requirements.


Albrecht-S:
Thanks for this quick fix. This was something I had observed myself but didn't get to looking into it yet.

PR #96, opened 2020-07-09
Convert Fl_Group::array_ to union to better represent its behavior

ras0219:
While using b0e0c355edaa2e23148cb0260ada907aec930f05 with Visual Studio 2019, I encountered crashes in Release that would not reproduce during Debug. While debugging the crash in release mode, I saw that `Fl_Group::array_` had unexpected invalid pointer values. Based on things not reproducing in Debug, I assumed the crash to be related to some undefined behavior.

Reading through the code, I noticed the re-use of `array_` to sometimes be of type `Fl_Widget*`, but sometimes as `Fl_Widget**`. This is a common source of UB pitfalls, so I rewrote it using a union to make things clearer (similar to other parts of the code). This rewrite fixed the crashes I experienced.

My code that caused the crashes was quite innocuous:
```cpp
struct MyWidget : Fl_Group
{
  MyWidget(int x, int y, int w, int h, const char* label = 0)
    : Fl_Group(x, y, w, h, label)
    , inner_widget(x, y, w, h) // crashes during inner_widget's constructor while adding to the superclass' group
  { this->end(); }

  Fl_Output inner_widget;
};
```

Specific OS/Platform details:
- Visual Studio Community 2019 Version 16.6.3
- MSVC 19.26.28806 Native x64 -> x64 Tools
- Windows 10.0.18363

### DIFF
--- a/FL/Fl_Group.H
+++ b/FL/Fl_Group.H
@@ -41,7 +41,10 @@ class Fl_Rect;
 */
 class FL_EXPORT Fl_Group : public Fl_Widget {
 
-  Fl_Widget** array_;
+  union {
+    Fl_Widget** arrayN_;
+    Fl_Widget* array1_;
+  };
   Fl_Widget* savedfocus_;
   Fl_Widget* resizable_;
   int children_;
--- a/src/Fl_Group.cxx
+++ b/src/Fl_Group.cxx
@@ -38,7 +38,7 @@ Fl_Group* Fl_Group::current_;
         is added or removed.
 */
 Fl_Widget*const* Fl_Group::array() const {
-  return children_ <= 1 ? (Fl_Widget**)(&array_) : array_;
+  return children_ <= 1 ? &array1_ : arrayN_;
 }
 
 /**
@@ -317,9 +317,9 @@ int Fl_Group::navigation(int key) {
   int i;
   for (i = 0; ; i++) {
     if (i >= children_) return 0;
-    if (array_[i]->contains(Fl::focus())) break;
+    if (arrayN_[i]->contains(Fl::focus())) break;
   }
-  Fl_Widget *previous = array_[i];
+  Fl_Widget *previous = arrayN_[i];
 
   for (;;) {
     switch (key) {
@@ -342,7 +342,7 @@ int Fl_Group::navigation(int key) {
     default:
       return 0;
     }
-    Fl_Widget* o = array_[i];
+    Fl_Widget* o = arrayN_[i];
     if (o == previous) return 0;
     switch (key) {
     case FL_Down:
@@ -361,7 +361,7 @@ Fl_Group::Fl_Group(int X,int Y,int W,int H,const char *l)
 : Fl_Widget(X,Y,W,H,l) {
   align(FL_ALIGN_TOP);
   children_ = 0;
-  array_ = 0;
+  array1_ = 0;
   savedfocus_ = 0;
   resizable_ = this;
   bounds_ = 0; // this is allocated when first resize() is done
@@ -473,18 +473,18 @@ void Fl_Group::insert(Fl_Widget &o, int index) {
   }
   o.parent_ = this;
   if (children_ == 0) { // use array pointer to point at single child
-    array_ = (Fl_Widget**)&o;
+    array1_ = &o;
   } else if (children_ == 1) { // go from 1 to 2 children
-    Fl_Widget* t = (Fl_Widget*)array_;
-    array_ = (Fl_Widget**)malloc(2*sizeof(Fl_Widget*));
-    if (index) {array_[0] = t; array_[1] = &o;}
-    else {array_[0] = &o; array_[1] = t;}
+    Fl_Widget* t = array1_;
+    arrayN_ = (Fl_Widget**)malloc(2*sizeof(Fl_Widget*));
+    if (index) {arrayN_[0] = t; arrayN_[1] = &o;}
+    else {arrayN_[0] = &o; arrayN_[1] = t;}
   } else {
     if (!(children_ & (children_-1))) // double number of children
-      array_ = (Fl_Widget**)realloc((void*)array_,
+      arrayN_ = (Fl_Widget**)realloc((void*)arrayN_,
                                     2*children_*sizeof(Fl_Widget*));
-    int j; for (j = children_; j > index; j--) array_[j] = array_[j-1];
-    array_[j] = &o;
+    int j; for (j = children_; j > index; j--) arrayN_[j] = arrayN_[j-1];
+    arrayN_[j] = &o;
   }
   children_++;
   init_sizes();
@@ -518,11 +518,11 @@ void Fl_Group::remove(int index) {
 
   children_--;
   if (children_ == 1) { // go from 2 to 1 child
-    Fl_Widget *t = array_[!index];
-    free((void*)array_);
-    array_ = (Fl_Widget**)t;
+    Fl_Widget *t = arrayN_[!index];
+    free((void*)arrayN_);
+    array1_ = t;
   } else if (children_ > 1) { // delete from array
-    for (; index < children_; index++) array_[index] = array_[index+1];
+    for (; index < children_; index++) arrayN_[index] = arrayN_[index+1];
   }
   init_sizes();
 }


Albrecht-S:
Thanks for this interesting PR. `array_` is private, so thechnically spoken we could apply it. But I have some questions.

First (real, serious) question: **why** should this change anything? Did you find and fix any coding errors? I can't see significant changes in your code (except clarity), hence the machine code should be the same. Did I miss anything?

Second question: can you provide a short but full, compileable example program (source code) that shows the crash? I'd like to test it and see if I can reproduce the issue on my system (I can test on Win10 with VS 2019).

That said, I'm neither pro nor con right now. Your proposed changes can make the code clearer in some points, but the different names (of union members) can make things less obvious (for a human reader). WRT names: I'd probably have left the `array_` variable name (union member) as-is and added another name like `child1_` (since it is not an array in this case). But that are only minor details.

Last question: can you please explain what "UB pitfalls" are (what is a "UB")? I tried to find it with acronym finder, but there are too many irrelevant hits. TIA.

erco77:
On 2020-07-09 05:23, Albrecht Schlosser wrote:
> Last question: can you please explain what "UB pitfalls" are (what is a "UB")?
> I tried to find it with acronym finder, but there are too many irrelevant hits. TIA.


	Just a guess, "Undefined Behavior".
	FWIW I've never heard of this acronym either.

	https://en.wikipedia.org/wiki/Undefined_behavior

ras0219:
Apologies for the terse language -- in my professional circles 'UB' is a common acronym for Undefined Behavior[1]. Essentially, if per the C/C++ specifications your application has undefined behavior it means that the program can do _literally anything_. Unfortunately, it is extremely easy to accidentally introduce undefined behavior and compilers generally do not have any errors or warnings to diagnose these problems.

As an example, the following has undefined behavior:

```cpp
float x;
int* p = (int*)&x;
int y = *p; // undefined behavior to access x through pointer of type int*
```

Link [1] has many more examples of undefined behavior, and I highly recommend the series of three LLVM blog posts on the subject [2].

While I will try to provide a reduced program, these crashes are often difficult to reduce since they can rely on subtle optimizations that change randomly due to seemingly unrelated code changes (for example, introducing or removing another variable of a related type anywhere else in the program).

My hope is to author this PR such that it's unambiguously _clearer_ code, which happens to solve my issue as a bonus. I'd be happy to rename `arrayN_` and `array1_` however you'd like if you feel that this makes the code clear enough to be an unambiguous improvement.

[1] https://en.cppreference.com/w/cpp/language/ub
[2] http://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html

Albrecht-S:
Thanks for your further comments, I will reply to them in a follow-up. Meanwhile I created my own demo program using your example code combined with test/hello.cxx to make a working program. I replaced the `inner_widget` of type `Fl_Output` with `Fl_Box` and moved the initialization of the box widget (in `hello.cxx`) to the constructor. See attached file [pr96-demo.cxx.txt](https://github.com/fltk/fltk/files/4898452/pr96-demo.cxx.txt).

As expected (at least by me) this program behaved correctly in both Debug and Release builds on Windows 10 with VS 2019.

... which makes perfectly sense taking into account that it's all about **undefined behavior**. I know.

Anyway, we never had complaints about this, and I assume that this code has been in FLTK from the very beginning (1998) until today. Which does not mean that it's correct. ;-)

I wonder what other change in your program may have triggered the faulty behavior…

Albrecht-S:
That said, I agree that your proposal is correct (AFAICT). I didn't know the acronym, thanks for your explanations.

My personal take on the variable (union member) names is as I wrote above. I'd appreciate if you could rename `arrayN_` ("back") to `array_` and `array1_` to `child1_`. But that's (so far) only my personal feeling, other devs may have different opinions. We may want to disscuss this change in fltk.coredev or vote here.

I see your point to rename the variable in both cases to avoid accidental use of it in other code parts, but I think this is not necessary and I prefer having the old name where it used to be, and our CMP states that the accessor method for private members with trailing underscore (`array_`) should be `array()`, hence we should IMHO keep this name.

Albrecht-S:
@fltk/coredevs Please vote (two votes):

Vote 1: Accept this PR in general, i.e. introduce the union? Vote +1, 0, or -1

If vote 1 is accepted:

Vote 2: Rename the union members as I suggested (+1) or keep the names as in the PR (-1)? Please vote for vote 2 no matter what you voted for vote 1.

My votes are: +1 and +1.

Other variable names may be suggested as well.

imaclmaca:
> On 9 Jul 2020, at 19:08, Albrecht Schlosser wrote:
> 
> 
> @fltk/coredevs Please vote (two votes):
> 
> Vote 1: Accept this PR in general, i.e. introduce the union? Vote +1, 0, or -1
> 
> If vote 1 is accepted:
> 
> Vote 2: Rename the union members as I suggested (+1) or keep the names as in the PR (-1)? Please vote for vote 2 no matter what you voted for vote 1.
> 
> My votes are: +1 and +1.
> 
> Other variable names may be suggested as well.


OK - I just spent some time reviewing the PR. 
And yesterday I spent a fair bit of time in Bitbucket, reviewing PR’s. This is a lot like work!

Anyway; results. I think the proposed change is OK so vote 1 is +1 from me. 
For vote 2... I’m ambivalent, I didn’t greatly like the array1_ / arrayN_ names so I guess child1_ / array_ is better? But in either case a little comment would resolve the ambiguity. I guess on balance I’m +1(ish.)


<potential rant approaching>

As regards the discussion, this looks a lot like the aliasing issues I’ve been seeing with modern compilers recently. Undefined Behaviour, as the OP said.

The current fashion amongst compiler writers is that if the spec does not define a specific behaviour, and you can optimise a chunk of code by taking some interpretation that suits you, even if it will surprise the hell out of the coders, then so be it, optimise away!

A lot of code that has worked “correctly” since forever, breaks when recompiled with modern compilers because they take a much more aggressive stance on optimising the code, and anything that is not nailed down is fair game. It is a major PITA. I’ve been hit by issues “caused” by gcc, icc, MS cl, diab...


My all time favourite undefined behaviour optimisation (though not even a particularly recent one, and not an aliasing issue) is what happens when you add two positive signed complement integers together. 
Now, we know that at the hardware layer there comes a point where the result will overflow and become negative. But the C (and C++) spec are silent on this, so it is *undefined*. (Java, note, explicitly defines overflow as the expected behaviour, FWIW.)

So in my embedded code, I have a lot of fixed point representations, and I’m adding lots of numbers that are potentially close to saturation, so I do... (massively simplified pseudocode example)

a = positive_value;
b = another_positive_value;
c = a + b;
if (c < 0)
{
  // deal with overflow
}

And this code now fails in many places and the overflow is never caught. 

Why? Because the compiler can reliably infer at compile time that a and b must be positive, and two positive values added together return a positive result, therefore the check of c is useless and can be optimised away. 

Now we all know that reasoning is bollocks, and the compiler authors know it is bollocks, and everybody knows it is not the behaviour that coders expect, but the compilers do it anyway because the behaviour is undefined, and they can speed their benchmarks a little bit more with this (or something, I do not know *why* they do this...)

So, now with gcc I’m a big fan of -fwrapv as an option, or at least -fno-strict-overflow.

OK, my code may be slightly less optimised; but on the other hand it will actually work. Which is always a nice feature.

Albrecht-S:
I just added the label `bug`: if we agree it's undefined behavior, then it's a **bug** to use the code in its old form.
@imaclmaca Thanks for your votes.

abrolag:
On Thu, 09 Jul 2020 13:53:20 -0700
Ian M <notifications@github.com> wrote:

>My all time favourite undefined behaviour optimisation (though not even a particularly recent one, and not an aliasing issue) is what happens when you add two positive signed complement integers together. 
>Now, we know that at the hardware layer there comes a point where the result will overflow and become negative. But the C (and C++) spec are silent on this, so it is *undefined*. (Java, note, explicitly defines overflow as the expected behaviour, FWIW.)
>
>So in my embedded code, I have a lot of fixed point representations, and I’m adding lots of numbers that are potentially close to saturation, so I do... (massively simplified pseudocode example)
>
>a = positive_value;
>b = another_positive_value;
>c = a + b;
>if (c < 0)
>{
>  // deal with overflow
>}
>
>And this code now fails in many places and the overflow is never caught. 
>
>Why? Because the compiler can reliably infer at compile time that a and b must be positive, and two positive values added together return a positive result, therefore the check of c is useless and can be optimised away. 

This puzzles me.
If they don't treat it as going negative, what on earth *do* they do with it?


-- 
Will J Godfrey
http://www.musically.me.uk
Say you have a poem and I have a tune.
Exchange them and we can both have a poem, a tune, and a song.

imaclmaca:
This is way off-topic for this thread now. Sorry!


> On 9 Jul 2020, at 22:41, Will J Godfrey wrote:

> >So in my embedded code, I have a lot of fixed point representations, and I’m adding lots of numbers that are potentially close to saturation, so I do... (massively simplified pseudocode example)
> >
> >a = positive_value;
> >b = another_positive_value;
> >c = a + b;
> >if (c < 0)
> >{
> > // deal with overflow
> >}
> >
> >And this code now fails in many places and the overflow is never caught. 
> >
> >Why? Because the compiler can reliably infer at compile time that a and b must be positive, and two positive values added together return a positive result, therefore the check of c is useless and can be optimised away. 
> 
> This puzzles me.
> If they don't treat it as going negative, what on earth *do* they do with it?


Nothing.
The compiler never emits any code for the conditional check at all, as it has “proved” that it is not necessary, and because the behaviour is undefined it can do what it likes. The compiler only has to do what is defined, never mind what actually makes sense...
So the code breaks at runtime.
Can you detect my annoyance at this “improvement” to the compiler? 

And to make it worse, it doesn’t happen every time that construction appears in the code, since it depends on whether the compiler reckons, from its analysis of the flow and context and so on, whether it knows that the values are both positive (or both negative) before deciding to remove the check. So you can do the “same thing” in several places and see some places work “right" and some fail. So -fwrapv, always.

erco77:
Ian, in the a/b/c/ case, did the compiler warn about optimizing this code away (I hope),
or did it just silently optimize it away? If the latter, I'm never using the optimizer again, lol

ras0219:
After trying to reduce my use case, I've come to the conclusion that I must have mixed up the 1.3 and 1.4 headers at some point when flipping between them. The exact crash I saw was during `delete[] sizes_` in `Fl_Group::init_sizes()`, with mysterious values. I think it fully explains my scenario if my `main.cpp` believed that `sizeof(Fl_Group)` was exactly one pointer shorter than `fltk`'s binary. Due to inlining, the vtable pointer of my first data member was being assigned over the `::sizes_` field, thus giving it the bogus value. It seems that my Debug configuration somehow used the right headers :)

I'm still not sure how I flipped the headers around and I'm very sorry for the false alarms. I hope the catharsis of ranting about UB makes up for it :)

Finally, I am still happy to see this PR through because as I said above, I hope it has value (cleaner code) independent of my particular problem.

Albrecht-S:
@ras0219 Thanks for letting us know what the actual cause was. Mixing headers happens pretty often. Anyway, thanks for your report, we appreciate all improvements like this.

Albrecht-S:
@ManoloFLTK, @erco77 your votes, please? See this https://github.com/fltk/fltk/pull/96#issuecomment-656273797 in case the long UB discussion distracted.

ManoloFLTK:
My understanding is that it's no longer a bug (even one that would appear only in certain
optimization conditions) that is being fixed. It's a matter of rewriting the code to make
conspicuous that a pointer variable is sometimes assigned a widget* value and 
other times a widget** value. Is that correct?

My vote is +1 to create the union.
It's also +1 to rename that union's members as Albrecht proposes, because only
one of these members is an array.

Albrecht-S:
Current status: +3 and +3, i.e. our decision is clear to add the union with renamed members as proposed.

@ras0219: Would you like to amend your PR? If you do, you can just add another commit and we can rebase/squash as needed. Please confirm or let me know (if not), then I'll do it. Thanks for your contribution and all the input about undefined behavior (UB).

Albrecht-S:
@ManoloFLTK : thanks for your vote. WRT bug or not: although @ras0219 confirmed that the crash was caused by mixing headers, all **undefined behavior** in our code should IMHO be considered a bug. Hence it's still a bug, but the other point is as you say to make the code more explicit and clear.

ManoloFLTK:
@Albrecht-S: I fully agree FLTK should not rely on any undefined behavior. I wonder, though, whether storing in the same pointer-sized variable either a Fl_Widget* or a Fl_Widget** value depending on the value of a well defined other variable is undefined behavior.

Albrecht-S:
Just some **thoughts** (not backed by specifications, others may know better):

- I **believe** that the standard doesn't tell us what the pointer size of any type is or has to be, hence different types might have implementation specific sizes (in practice this is probably never the case, although early Windows implementations had short and long pointers IIRC, or whatever it was called)
- if we'd declare `void * array_` instead and cast all pointers when accessing this `array_' member it would probably not be undefined behavior, but even more confusing (you can store any pointer type in `void *`)
- I also think that the behavior would be OK for all practical implementations, but that wouldn't change that it's _undefined_ and _"the compiler can do anything it wants"_.

But this is, as Ian already said, OT.

erco77:
> @ManoloFLTK, @erco77 your votes, please? See this [#96 (comment)](https://github.com/fltk/fltk/pull/96#issuecomment-656273797) in case the long UB discussion distracted.

Of importance if TLDR, here's the diff:
https://github.com/ras0219/fltk/commit/d6d91c94624655ffc91ef01f3de3d08c36d1244f

I'll weigh in when I have some time today.

Albrecht-S:
@ras0219 : I should have mentioned that I prepared a commit which I'm ready to push if you agree:
[pr96-patch.txt](https://github.com/fltk/fltk/files/4904126/pr96-patch.txt). It's basically your patch with renamed union members and two additional comments, then squashed to one commit and signed-off by me.

Albrecht-S:
@erco77 and others:

Please see my attached patch in https://github.com/fltk/fltk/pull/96#issuecomment-656751317 which simplifies the changes since the original `array_` variable kept its name (or got it back).

erco77:
Approach seems sound to me -- good use of a union.

Let's not use those varnames though. Maybe "`array_wp_`" and "`array_wpp_`" instead?

Or perhaps better/clearer, name the union "`array_`" and make the two members `wp` and `wpp`, e.g.
```
union {
    Fl_Widget *wp;     // widget ptr
    Fl_Widget **wpp;   // widget ptr ptr
} array_;
..
array_.wp[i] = ..
array_.wpp[i] = ..
..
```

erco77:
Oh, and regarding Albrecht's proposed name change "array_" vs "child1_".. I guess only if it makes sense.
For me, "child1" implies there's a "child2", so not a fan.
Also, it seems very important to make clear to the readers of the code one var is a ptr, the other's a ptrptr, so I think it helps to have that distinction in the name, hence my suggestion of wp/wpp.
I refer to the original unix system source code (1970's) for the p vs pp naming convention.

Albrecht-S:
Greg, we did already vote +3 and +3, you're a little late to the party. Of course, your vote is still appreciated.

Which varnames did you mean? Those in the original PR (which you linked above), or those after the rename we discussed (poposed by me) which essentially keeps the original `array_` name for two children or more and adds `child1_` for the case that there's exactly one child?

Note: I missed one rename in my previously posted commit which I added in an amended commit: [pr96-patch-v2.txt](https://github.com/fltk/fltk/files/4904229/pr96-patch-v2.txt). I think this is a good choice: using the union with minimal diffs. Any concerns with this?

Albrecht-S:
@erco77 Sorry, our posts crossed in time.

> I refer to the original unix system source code (1970's) for the p vs pp naming convention.

We don't use such naming conventions in FLTK though, IMHO we shouldn't start it now. But if you feel better with `child0_` instead of `child1_` (because it's the "0'th" child in C convention) I wouldn't mind changing it...

Albrecht-S:
BTW, another possible name to avoid confusion would be `single_child_` but I think the comments in the code are explicit enough. Anyway, if that's a better option, I wouldn't mind.

erco77:
> Which varnames did you mean? Those in the original PR (which you linked above), or those after the rename we discussed (poposed by me) which essentially keeps the original array_ name for two children or more and adds child1_ for the case that there's exactly one child?

I was referring to the names in the original PR. I saw your proposals later.
My initial hook into this thread was this msg:

> @ManoloFLTK, @erco77 your votes, please? See this #96 (comment) in case the long UB discussion distracted.

..so I just read comment 96 and replied to that first.
I'm OK with whatever you all come up with I guess. Still, I do like the array_.<names> approach.

erco77:
> Note: I missed one rename in my previously posted commit which I added in an amended commit: pr96-patch-v2.txt. I think this is a good choice: using the union with minimal diffs. Any concerns with this?

Ah, the comments on the union members helps a bit.
So hmm, so I take it all this is about that Fl_Group handle a single child differently from more than one.
Is there a reason not to just treat everything as an array, including a single child? I can't imagine there's a big performance gain by having lots of if's everywhere. But there must be sound reasons, I haven't dived into the whole code, cause I'm working on issue99 ATM which is kinda deep.

Albrecht-S:
The strategy is obviously all about saving memory and to avoid memory allocation entirely if there is only one child (or none).
We have only one pointer variable which is (names after suggested patch applied) either:
- `not used`: no children (`array_` initialized with 0)
- `Fl_Widget *child1_`: a pointer to the only child of the group
- `Fl_Widget **array_`: a pointer to an allocated array of pointers to children of the group

Another memory saving ___trick___ is that we don't have a variable for the allocated size of this array. I won't elaborate on this **very tricky thing**; read the source. ;-)

So if we always wanted to treat it as an array we'd have to allocate an "array" of 1 pointer if we have 1 child. It would work, but it's useless. Currently this **very tricky thing** above would imply that we'd `realloc()` the array when child no. 2 is added because a new array is `realloc()`ed whenever `children_ == (2**n); n > 0`.

I'd guess that this is Bill's code from the very beginning of FLTK (never checked though). Like to change it? ;-)

Note that this is all in Fl_Group which might be used very often, but I don't know what we'd gain or lose if we'd initially allocate an array of, say, four pointers, and start `realloc()`ing when we need five or more. It's performance vs. memory (and memory fragmentation as well).

We could indeed think about changing this in the year 2020 where (even embedded?) systems tend to have more memory than in 1998. A variable for the allocated array size would add another (unsigned) `int` but could simplify the code even more.

imaclmaca:
> On 10 Jul 2020, at 17:06, erco77 wrote:
> 
> Approach seems sound to me -- good use of a union.
> 
> Let's not use those varnames though. Maybe "array_wp_" and "array_wpp_" instead?
> 
> Or perhaps better/clearer, name the union "array_" and make the two members wp and wpp, e.g.
> 
> union {
>     Fl_Widget *wp;     // widget ptr
>     Fl_Widget **wpp;   // widget ptr ptr
> } array_;
> ..
> array_.wp[i] = ..
> array_.wpp[i] = ..


I was going to say I wasn’t at all keen on the union being named as “array_” because in the single-child case there is no array (which is kinda the point of Albrecht’s proposal to rename them anyway...)

But I now see that by the time I got to this, it had all been sorted out anyway. I was late...

ras0219:
@Albrecht-S the patches LGTM, I'd also be happy to rename things as requested. Just let me know how you'd like me to proceed.

Albrecht-S:
@ras0219 Thanks for confirmation, I committed the patch as posted above.

PR #99, opened 2020-07-09
Add error message box to Fl_File_Chooser (STR 1086)

Albrecht-S:
This commit adds an error message to Fl_File_Chooser so the user
can see that there are no files.

To do: provide the cause of an error (if any) or document how to
find it.


Albrecht-S:
@erco As discussed in private mail: here's a PR for STR 1086 to add an error message to Fl_File_Chooser. Please review, discuss, merge,... (whatever).

For more info see [STR 1086](https://www.fltk.org/str.php?L1086).

I suggest to continue discussion (if any) here and to close STR 1086 with a link to this PR.

Albrecht-S:
PS: @erco77 testing PR handling, hence @ mentioning you and requesting review.

erco77:
OK, I'm cloning your fork and will give it a try on some odd cases.

erco77:
Hmm, I can't seem to trigger the new error, maybe I'm using the wrong test program:

As root:
```
    mkdir /var/tmp/closed
    chmod 0 /var/tmp/closed
    touch /var/tmp/closed/foo.pdf

```
As normal user:
```
    > Run your fork's  test/file_chooser app
    > Hit ".." button, opens browser
    > Enter for Filename: /var/tmp
    > Click on 'closed' dir
    > Sees no errors and an empty window <-- EXPECTED PERMISSION DENIED ERROR
```

erco77:
Oh wait, I must need to check out the branch "str-1086_file-chooser".
I hate how 'git branch' doesn't show all the branches, and you have to use 'git branch -a'. I know you mentioned that, but it's easy to forget!
Retrying..

erco77:
OK, now it says "No files found..." for a second and then goes away.
Easy to miss IMHO, and it should try to show the actual OS error ("Permission Denied")
so that the user knows it's not a typo or something else.
I'll see if I can dive into the patch and see if I can recover the error reliably..

erco77:
It looks like this could work, but I'm gonna dive down into fileList->load() and make sure it returns errors properly.
```
--- a/src/Fl_File_Chooser2.cxx
+++ b/src/Fl_File_Chooser2.cxx
@@ -1058,7 +1058,8 @@ Fl_File_Chooser::rescan()
   int nfiles = fileList->load(directory_, sort);
   if (nfiles <= 0) {
     // fprintf(stderr, "No files found\n");
-    errorBox->label("No files found ...");
+    errorBox->label(strerror(errno));
+    // errorBox->label("No files found ...");
     errorBox->show();
   } else {
     errorBox->hide();
```

Albrecht-S:
It's only a proof of concept, and now when I tested it again, I noticed that the error message goes away whenever you click somewhere else. Hard to catch it in a screenshot, but I got it. I was not aware of this but it was intentional that the error message would be cleared whenever you click on another dir. Note that the real cause of the error, an error number or error message, would be difficult to get and present via the current API. I stopped where I got and left the "real error message stuff" for later. I thought of "you" because I (a) was too lazy to change the API and (b) feel not really comfortable with cross-platform error message stuff.

Anyway, screenshot 
file-chooser-error attached.

Albrecht-S:
I can confirm that `errorBox->label(strerror(errno));` works on my Linux box.

erco77:
I'd like to suggest we improve Fl_File_Browser a bit.
```
1) Currently Fl_File_Browser::load()'s return value is undocumented, 
   but seems to return the number of files found, or 0 in the case of either
   empty or errors.

2) load()s doxgen docs should be moved into the .cxx code,
   where the implementation is.

3) Add a way for load() to return actual error messages from the OS
   across all platforms.
```
One way to do (3) would be to add an error message string:

```
class FL_EXPORT Fl_File_Browser : public Fl_Browser {
    const char *errmsg_;   // ctor inits to NULL
[..]
public:
    const char *errmsg() const { return errmsg_; }
    void errmsg(const char *s); // > handles making strdup() of 's'
                                // > frees old errmsg first
                                // > dtor free()s if any
```
..and ensure the lower level code passes any OS errors back up, and the main app can check for an error message by testing errmsg() for NULL, and displaying it if any. Each new call to load() would first set errmsg(NULL) to clear any old errors.

erco77:
BTW, I'd be happy to do this, unless you want to take it.

Albrecht-S:
Don't worry, be happy ;-)

Please go ahead, all looks good. I didn't check the details in source code, but I believe you and it's bed time now here in Europe.

Albrecht-S:
PS: feel free to use my commit or not, i.e. create a complete new commit as you like. Mine was only intended as a POC.

Albrecht-S:
PPS: whenever you need to write new `.cxx` and `.h` files from `.fl` files (`fluid -c`) please note that fluid adds some trailing spaces. We need to fix this, but for now please:

- use `git diff --check` before `git add` or anything else to see if there are whitespace errors
- fix trailing whitespace in the generated files

Thanks.

erco77:
OK, I'll work on it.

Might touch a bit of code, as none of the low level driver code seems to reliably pass up
error messages, so a peer review will likely be needed (you/manolo).

I might make a separate fork, or maybe just branch FLTK -- noticed your repo's branch
has $Id$ tags, which must mean your fork is pre-big-$Id$/whitespace purge..?

Albrecht-S:
Nope, the current PR shouldn't have `$Id$` tags. See this comparison to current master:

https://github.com/fltk/fltk/compare/master...Albrecht-S:str-1086_file-chooser

I had previously pushed another commit with a similar branch name which I referred to in the STR comments, but the current branch should be "clean". The only `$Id$` tags I found are in the `zlib` folder.

Note the current branch is [str-1086_file-chooser](https://github.com/Albrecht-S/fltk/tree/str-1086_file-chooser) with a hyphen `'-'` in `file-chooser`, not the older one with an underscore. Commit is ad6566fad0a. Fortunately the relevant code **should** be the same (I think I only removed one `printf` statement which was commented out anyway).

Sorry for the confusion, I'll clean up tomorrow.

Albrecht-S:
WRT "fork or branch in fltk/fltk": I prefer to work in one's own fork until it's finished, then rebase/squash/merge and push to fltk/fltk only when it's done. You can create a PR or just point at your fork and branch so we can review it (the PR method is recommended because we can use GitHub for diffs and comparisons).

Another reason is that our **users** don't need to download WIP stuff and that the overall size of the FLTK repo will be smaller (without all that WIP stuff). We'll need to discuss this in fltk.coredev together with all the other GitHub handling discussions.

erco77:
OK, well I committed my fixes as a new branch "issue_99". (Next time I'll try a fork for large mods)
Give it a spin. I tested on Linux, mac 10.10.x, Windows VS2017 and Windows mingw64.

@Albrecht-S -- Big thing that needs testing that I can't do: 
**Check error messages in another language.**
Especially after building in Visual Studio. Win32 error messages in multiple languages are often a beast. Not sure if utf8 content gets back to the UI. To test I suggest:

1. Build FLTK in VS20xx
2. Create a subdir in c:\temp called "closed"
3. Change the perms so that closed is unreadable
4. Run the test file_chooser.exe app, browse to c:\temp and double-click on "closed"
5. You should see an error message (e.g. "Access denied"), hopefully in your local language
6. Verify special characters look ok. If not, problem is likely in get_ms_errmsg().

There are other ways to trigger errors, such as browsing to partial UNC paths, e.g. "//badhost/"
Note that resizing the chooser/preview border should word wrap error msgs properly.

Albrecht-S:
@erco77 Great job, works like a charm.

However, first tests with foreign language settings on my Linux box show **unexpected** results with an inaccessible directory as suggested:

1. The **good** one first: I built on Linux with MinGW-w64 cross tools and ran `pixmap_browser.exe` with `wine`:
```
LC_ALL=de_DE.utf8 wine mingw/bin/examples/pixmap_browser.exe
```

The result is correct with error message "Zugriff verweigert": issue-99-file-chooser-wine_DE

2. Now the bad news: no matter what locale I set and no matter whether I set it like shown above or with `export LC_ALL=de_DE.utf8` the error message is always "Permission denied". :-( I checked with `locale` and all looks fine.

Other **good news**: I see that the error message is kept alive now (better than in my POC). Thanks for the good work.

Looking forward to whatever you may find why the Linux version does not honor the locale setting. Thoughts: I know that we're setting the locale under some conditions in the source code...

I'll try testing on a real Win10 box, but if it works well with `wine` I think it should work on a real "German" Windows 10 box as well. Hmm, it's **Windows**, so I'll better test it. You never know. ;-)

erco77:
Regarding "bad news" case: so by setting the locale, which specific variable do you mean -- the `LANG` variable? And when doing anything to that variable, that changed the behavior of the `pixmap_browser/exe` to show English instead of the local language? If so, interesting.. maybe..

When I built FLTK with mingw, it compiled `src/scandir_win32.c` (and _not_ `src/scandir_posix.c`) which means mingw used the FormatMessage() stuff to get the error message, so the mingw binary _might_ _not_ respond to all POSIX techniques.

To me that's hopeful that the code I was concerned about, `get_ms_errmsg() + FormatMessage()`, is returning proper localized language (`"Zugriff verweigert"`), though that string appears to be all ASCII characters -- I was hoping there might be some non-ascii/Unicode in there, which would give me hope that utf8 is working (not sure from your example), so that other non-ASCII languages (e.g. Japanese) work OK.

Albrecht-S:
Some clarifications regarding foreign languages:
1. The Windows stuff seems to work fine, tested cross-compiled and run with `wine` on my Linux box only (so far)
2. Will test on a native Windows box later
3. The **surprising**, **unexpected** behavior (the "bad news" case) was on **native Linux**
4. I wrote that I used `LC_ALL` which sets **ALL** locale variables:
```
$ locale
LANG=
LANGUAGE=
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC="de_DE.UTF-8"
LC_TIME="de_DE.UTF-8"
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_PAPER="de_DE.UTF-8"
LC_NAME="de_DE.UTF-8"
LC_ADDRESS="de_DE.UTF-8"
LC_TELEPHONE="de_DE.UTF-8"
LC_MEASUREMENT="de_DE.UTF-8"
LC_IDENTIFICATION="de_DE.UTF-8"
LC_ALL=de_DE.UTF-8
$ ls -la
insgesamt 228                                                             <<< GERMAN <<<
drwxr-xr-x  6 albrecht albrecht   4096 Jul 11 14:54 .
drwxr-xr-x 19 albrecht albrecht   4096 Jul 11 14:54 ..
drwxr-xr-x 11 albrecht albrecht   4096 Jul 11 14:17 Debug
drwxr-xr-x 13 albrecht albrecht   4096 Jul 11 14:19 mingw
-rw-r--r--  1 albrecht albrecht 174556 Jul 11 14:16 .ninja_deps
-rw-r--r--  1 albrecht albrecht  32477 Jul 11 14:16 .ninja_log
drwx------  2 root     root       4096 Jul 11 14:53 xroot
drwxr-xr-x  2 albrecht albrecht   4096 Jul 11 14:53 xtemp
$ ls xroot/
ls: Öffnen von Verzeichnis 'xroot/' nicht möglich: Keine Berechtigung     <<< GERMAN <<<
$ bin/examples/pixmap_browser 
bash: bin/examples/pixmap_browser: Datei oder Verzeichnis nicht gefunden  <<< GERMAN <<<
$ Debug/bin/examples/pixmap_browser
$ LANGUAGE=de_DE Debug/bin/examples/pixmap_browser
```

Both commands show "Permission denied". Tested with "de_DE.utf8" as well w/o success.

I'll try to figure out a case where I can get special (non-ASCII) characters.

Anything else I could try?

BTW: maybe you could install the German or any other locale and test yourself?

erco77:
The trouble for me when testing other languages on Windows, IIRC, is that /everything/ changes into the new language to where you can't understand the menus or anything. I might know enough french to get around town, but the techno speak is usually otherworldly.
Basically just need to know if Unicode chars are getting through, and if not.. well, I might need advice, as converting Windows multibyte to utf8 is probably some serious insanity. I recall the "recommended" way to convert Windows strings into utf8 was to first convert to a CString (or some such), which I think was a class, which means C++, which is bad cause a lot of the low level code is in C not c++... meh

erco77:
If it were me, I'd just commit it to master, and let the world debug it, lol. English error messages are better than none at all, and if there's problems with local languages, those problems can be fixed in the future as time permits.
Basically, I just wanna mark it done and move on to the next bug, lol.

MatthiasWM:
Untested code to convert from utf8 to utf16 (Windows) and back without using any OS resources. Tested version will be available in a few days:

static uint32_t getUtf8Char(uint8_t const** sp)
{
    const uint8_t* s = *sp;
    uint32_t c = *s++;
    if (c < 0x80) {
        ;
    } else if (c < 0xE0) {
        uint32_t c1 = *s++;
        c = ((c & 0x1F) << 6) | (c1 & 0x3F);
    } else if (c < 0xF0) {
        uint32_t c1 = *s++;
        uint32_t c2 = *s++;
        c = ((c & 0x0F) << 12) | ((c1 & 0x3F) << 6) | (c2 & 0x3F);
    } else if (c < 0xF8) {
        uint32_t c1 = *s++;
        uint32_t c2 = *s++;
        uint32_t c3 = *s++;
        c = ((c & 0x07) << 18) | ((c1 & 0x3F) << 12) | ((c2 & 0x3F) << 6) | (c3 & 0x3F);
    } else {
        c = '?';
    }
    *sp = s;
    return c;
}


static void putUtf16Char(uint8_t** dp, uint32_t c)
{
    uint8_t* d = *dp;
    if (c > 0x10FFFF)
        c = '?';
    if (c < 0x10000) {
        *d++ = c >> 8;
        *d++ = c;
    } else {
        uint16_t c1 = ((c - 0x10000) >> 10) + 0xD800;
        *d++ = c1 >> 8;
        *d++ = c1;
        uint16_t c2 = ((c - 0x10000) & 0x03FF) + 0xDC00;
        *d++ = c2 >> 8;
        *d++ = c2;
    }
    *dp = d;
}


static uint32_t getUtf16Char(uint8_t const** sp)
{
    uint8_t const* s = *sp;
    uint32_t c = (s[0] << 8) | s[1]; s += 2;
    if ((c & 0xFC00) == 0xD800) {
        uint32_t c2 = (s[0] << 8) | s[1]; s += 2;
        c = 0x10000 + ((c - 0xD800) << 10) + (c2 - 0xDC00);
    }
    *sp = s;
    return c;
}


static void putUtf8Char(uint8_t** dp, uint32_t c)
{
    uint8_t* d = *dp;
    if (c > 0x10FFFF)
        c = '?';
    if (c < 128) {
        *d++ = c;
    } else if (c < 2048) {
        *d++ = 0xC0 | (c >> 6);
        *d++ = 0x80 | (c & 0x3F);
    } else if (c < 65536) {
        *d++ = 0xE0 | (c >> 12);
        *d++ = 0x80 | ((c >> 6) & 0x3F);
        *d++ = 0x80 | (c & 0x3F);
    } else {
        *d++ = 0xF0 | (c >> 18);
        *d++ = 0x80 | ((c >> 12) & 0x3F);
        *d++ = 0x80 | ((c >> 6) & 0x3F);
        *d++ = 0x80 | (c & 0x3F);
    }
    *dp = d;
}


/** 
 * Read utf-8 text and convert to utf-16
 *
 * @param src[in] utf-8 string to convert
 * @param srclen[in] length of string in bytes to convert
 * @param dstlenp[out] length of utf-16 string in bytes
 *
 * @return converted string
 *
 * @note Converted character string must be freed by the caller
 */
char* NewtUtf8To16(const char* src, size_t srclen, size_t* dstlenp)
{
    uint8_t* dBuffer = (uint8_t*)malloc(srclen * 4);
    uint8_t* d = dBuffer;
    const uint8_t* sBuffer = (uint8_t*)src;
    const uint8_t* s = sBuffer;

    while (s - sBuffer < srclen) {
        uint32_t c = getUtf8Char(&s);
        putUtf16Char(&d, c);
    }
    size_t n = d - dBuffer;
    *d++ = 0;
    *d++ = 0;
    *dstlenp = n;
    return realloc(dBuffer, n + 2);
}



/**
 * Read utf-16 text and convert to utf-8
 *
 * @param src[in] utf-16 string to convert
 * @param srclen[in] length of string in bytes to convert
 * @param dstlenp[out] length of utf-8 string in bytes
 *
 * @return converted string
 *
 * @note Converted character string must be freed by the caller
 */
char* NewtUtf16To8(const char* src, size_t srclen, size_t* dstlenp)
{
    uint8_t* dBuffer = (uint8_t*)malloc(srclen * 2);
    uint8_t* d = dBuffer;
    const uint8_t* sBuffer = (uint8_t*)src;
    const uint8_t* s = sBuffer;

    while (s - sBuffer < srclen) {
        uint32_t c = getUtf16Char(&s);
        putUtf8Char(&d, c);
    }
    size_t n = d - dBuffer;
    *d++ = 0;
    *dstlenp = n;
    return realloc(dBuffer, n + 1);
}

erco77:
@MatthiasWM Welcome to the party, lol
Is it possible that code is already in fltk somewhere? e.g. fl_utf8.cxx and/or the src/xutf8 directory?
I really don't know much about how Microsoft packages multibyte strings out of FormatMessage()..

MatthiasWM:
Silly me. Yes, I think these will do. Sorry that I am so quiet. Too many other projects ;-) . But I am still occasionally reading, as you can see.

/* F2: Convert a UTF-8 string into a wide character string - makes UTF16 on win32, "UCS4" elsewhere */
FL_EXPORT unsigned fl_utf8towc(const char *src, unsigned srclen, wchar_t *dst, unsigned dstlen);

/* F2: Convert a wide character string to UTF-8 - takes in UTF16 on win32, "UCS4" elsewhere */
FL_EXPORT unsigned fl_utf8fromwc(char *dst, unsigned dstlen, const wchar_t *src, unsigned srclen);

Albrecht-S:
@MatthiasWM Welcome from me, too.

@erco77 Greg, there's **no problem** with Windows. I repeat myself: the problem is with foreign locale's **under Linux**!!! Under Linux you should be able to run a program with another locale w/o changing the whole system. I understand your language issues with Windows though, but that's not relevant because Windows seems to work well.

That said, feel free to commit and push, but please use the "Rebase and Merge" (third option on GitHub) or rebase the branch manually and delete it when done, something like this:
```
$ git rebase master issue_99
$ git checkout master
$ git merge --ff-only issue_99
$ git push
```

This should do it cleanly.

We can indeed polish the foreign language stuff later. As you said: close the issue (and STR) which is important - an English error message is better than none.

Albrecht-S:
> Is it possible that code is already in fltk somewhere? e.g. fl_utf8.cxx and/or the src/xutf8 directory?
I really don't know much about how Microsoft packages multibyte strings out of FormatMessage()..

Sure, we're using this all the time for system functions under Windows, for instance directory and file names, environment variables, and more.

erco77:
> @erco77 Greg, there's no problem with Windows. I repeat myself: the problem is with foreign locale's under Linux!!! 

Oh, I think I must have read your report before your edit where you clarified the bad news case was under linux, not linux/wine. Or, I missed that comment. _(I thought you were running into the problem by invoking linux/wine with a locale variable that was breaking it. I see that's not the case)_

Hmm, well, under linux I kinda trust that strerror(errno) is "doing the right thing", whatever that is, so I'm not concerned for my commits. You can probably replicate the issue with a simple C program, (e.g. below) and get same results:

```
#include <stdio.h>
#include <string.h>
#include <errno.h>
int main() {
    if ( ! fopen("/no/way", "r") ) printf("fopen: %s\n", strerror(errno));
    return 0;
}
```
I know of no other way to get OS error messages under unix. Maybe see the `strerror()` man page for hints for other things to try.

Reading the `strerror` man page myself, sounds like I should maybe be using strerror_r() for safety in the context of threads, as some user apps might be threading while the browser is open. I should probably check FormatMessage() too, as it has some variants.

I'll be sure to do your git recommendations.

erco77:
@Albrecht-S BTW, if you /can't/ replicate with the above small C program, /let me know/ as that might need some investigation. Perhaps FLTK is doing something to affect the error messages, but I'll bet it ain't in my code, lol. strerror() is just too easy to use. FormatMessage() on the other hand...

BTW, you said "Windows seems to work well" -- I take that to mean wine/mingw, and not a native Windows machine yet compiled with VS 20xx. (You mentioned you were trying to find some time to try that)

I'll wait on the native Windows test because I'm not sure I trust that Wine+mingw handles Microsoft's convoluted multilang error message system the same way a native Windows environment does, so still looking forward to your (or anyone's) test building the issue_99 branch under native Windows + Visual Studio in a non-English environment. Extra credit if you can get actual non-ASCII characters in the error message..! (Recapping in case @MatthiasWM wants to try!!)

Albrecht-S:
**Got it!** You *need* to call setlocale() at least once during runtime of the program.

From `man setlocale`:
```
#include <locale.h>
char *setlocale(int category, const char *locale);

If  locale is an empty string, "", each part of the locale that should
be modified is set **according to the environment variables**.
```

So your test program doesn't work (doesn't honor the locale environment variables), but this modified version does:
```
$ cat err.c
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <locale.h>  // added
int main() {
    setlocale(LC_ALL, "");  // added
    if ( ! fopen("/no/way", "r") ) printf("fopen: %s\n", strerror(errno));
    return 0;
}
$ 
$ gcc -o err err.c
$ ./err
fopen: No such file or directory
$ LC_MESSAGES=de_DE.utf8 ./err
fopen: Datei oder Verzeichnis nicht gefunden
$ LC_MESSAGES=nb_NO.utf8 ./err
fopen: Fila eller mappa finnes ikke
```

FWIW, here are some `errno` examples with German Umlauts:
```
$ LC_MESSAGES=de_DE.utf8 ./errno 0 30 2>&1 | egrep -i 'ä|ö|ü|ß'
errno  1 = Vorgang nicht zulässig
errno  4 = Unterbrechung während des Betriebssystemaufrufs
errno  6 = Kein passendes Gerät bzw. keine passende Adresse gefunden
errno  9 = Ungültiger Dateideskriptor
errno 11 = Die Ressource ist zur Zeit nicht verfügbar
errno 12 = Nicht genügend Hauptspeicher verfügbar
errno 14 = Ungültige Adresse
errno 16 = Das Gerät oder die Ressource ist belegt
errno 18 = Ungültiger Link über Gerätegrenzen hinweg
errno 19 = Kein passendes Gerät gefunden
errno 22 = Das Argument ist ungültig
errno 25 = Unpassender IOCTL (I/O-Control) für das Gerät
errno 26 = Das Programm kann nicht ausgeführt oder verändert werden (busy)
errno 27 = Die Datei ist zu groß
errno 28 = Auf dem Gerät ist kein Speicherplatz mehr verfügbar
```

You can obviously reproduce the corresponding English versions with `strerror(n)`.

Now the question remains: when to call `setlocale(LC_ALL, "");` in (all?) FLTK programs and/or what side effects this would have. Or do we only call it when needed in the new code? I believe it would be better to call it along the initialization to always have a defined state - not only when an error occurs.

FWIW, we have already:
```
void Fl_X11_Screen_Driver::open_display_platform() {
  if (fl_display) return;

  setlocale(LC_CTYPE, "");
  XSetLocaleModifiers("");
```

but this initializes only the LC_CTYPE part of the locale and only on X11.

@erco77 : Do you have an idea how to proceed with this **on Linux** and maybe on other systems as well ?

For now I added the `setlocale()` call in `main()` of `pixmap_browser.cxx` and, TADA!, **it works**!

So, now that this is done (and I pushed my separate "Fluid" commit) I can try to start one of my Windows boxes to test on a native Windows system.

I'll let you know, but don't hold your breath, it's dinner time and I will likely not do anything essential tonight.

erco77:
Great that you figured out calling setlocale() in the app code!
I made that same change to the test/file_chooser.cxx app, and can confirm a Chinese language error by running (sorry, I use a tcsh):

`    ( setenv LC_ALL zh_CN.utf8 ; test/file_chooser )
`

chinese-file-chooser-error

__________________________________________________________________________

..which in google translate sounds right to me:
__________________________________________________________________________

chinese-translated
__________________________________________________________________________

To do that test I had to install the chinese language option, which on sci linux 7.x was:

`    yum langinstall zh_CN
`

> FWIW, we have already:
> 
> void Fl_X11_Screen_Driver::open_display_platform() {
>   if (fl_display) return;
>   setlocale(LC_CTYPE, "");
>   XSetLocaleModifiers("");
> [..]
> @erco77 : Do you have an idea how to proceed with this on Linux and maybe on other systems as well ?

Hmm, **I'm not sure.**

Offhand I'd leave calling locale() to applications (e.g. our test programs that invoke Fl_File_Chooser) should probably have this mod, but not set it in the library code, and just advise its use in the widget docs on a case-by-case basis.

Having us set it in the library code /might/ negatively effect existing applications that don't have it set, or might try to set it to something else, our library code sneakily un-doing their efforts.

The above use in open_display_platform() I have to wonder about: is that good practice?
Apparently this same code exists in 1.3.0, so it predates the new driver code. I'm not sure why it's there though.

I'll see if I can track down why that was added; might have to trawl through the old SVN commits (by checking out the 1.3.0 tag, and then using git blame?) Also, there might be hints in the newsgroup history too, not sure, will check.

IMHO we should make your setlocale() mods to our test + example apps that exercise the file chooser, and leave the library code as is, but document the technique in the widget doxygen docs.

I feel like we should get the opinion of others on this, perhaps on fltk.coredev, as my above opinions are purely instinctual on this.

Also: I'm not sure if we should set this in Visual Studio compiles of the apps, where it may not be needed or even used. Not sure about mingw builds either, if we're calling non-POSIX FormatMessage()/GetLastError().

We should probably take all this to fltk.coredev to get other opinions.

erco77:
> The above use in open_display_platform() I have to wonder about: is that good practice?
> Apparently this same code exists in 1.3.0, so it predates the new driver code. I'm not sure why it's there though.
> I'll see if I can track down why that was added; 

So `git checkout branch-1.3 && git blame` results:

```
f9039b2ae2 (Michael R Sweet    1998-10-06 18:21:25 +0000  670) void fl_open_display() {
f9039b2ae2 (Michael R Sweet    1998-10-06 18:21:25 +0000  671)   if (fl_display) return;
f9039b2ae2 (Michael R Sweet    1998-10-06 18:21:25 +0000  672)
b6bde2e456 (Matthias Melcher   2008-09-10 23:56:49 +0000  673)   setlocale(LC_CTYPE, "");     ⭨⭨⭨⭨
219ae5e455 (Albrecht Schlosser 2020-02-10 13:12:28 +0100  674)   XSetLocaleModifiers("");
b6bde2e456 (Matthias Melcher   2008-09-10 23:56:49 +0000  675)
```

Based on the commit log, sounds like its addition was related to the utf8 port:

```
commit b6bde2e456 [..]
Author: Matthias Melcher <fltk@matthiasm.com>
Date:   Wed Sep 10 23:56:49 2008 +0000

    Merging the UTF8 patch, consisting of O'ksi'd s original 1.1.6 patch and additions by Ian. 
    PLEASE BE AWARE that the patch in its current incarnation is a regression in many aspects
    and further work is required before we can announce Unicode support.
```

erco77:
Here's someone who seems to have reacted the same way I did, opening a github issue back in 2013 against a bluetooth library called "signal11": Subject:"Libraries should never call setlocale":

[](https://github.com/signal11/hidapi/issues/142)

It should be noted though, the library author never removed the calls to setlocale(), the OP never followed up, and the author closed the issue without making changes. I checked out the library, it's still actively maintained, and the setlocale() calls remained in there.

So still not sure how to weigh in on this.. I can see people who might use FLTK not expecting the overhead of internationalization to be invoked.. but I'm not sure the call does anything unless locale related environment variables are set to non-defaults, so maybe it's benign?

Perhaps we should provide an fl_internationalization() function (or some such, e.g fl_locale_init()) so that internationalized apps can call this up front in main() to be sure locale stuff is enabled, and move the setlocale() call we have into that (along with the one you recommended), just so apps that don't want i18n overhead can avoid it..?

Albrecht-S:
Yes, I came to the same conclusion that we should **not** call `setlocale()` in any form in an initialization process. I remember that I changed the instruction you pointed out, but it was only a change of `XSetLocaleModifiers("");` because the existing one led to problems (I fixed an STR, reverting a previous change).

I agree also that application programmers must be able to call `setlocale()` for their apps with any arguments they need at any time they want. Calling it in `fl_open_display()` or a related function could lead to surprising results (undoing the previous call, for example).

The best thing we could do is probably  document the necessity to do it or to provide a call (as you suggested) that application programmers can do it deliberately at any time so they know the arguments it was called with. Something like that, anyway.

WRT our demo programs: I suggest to add this call to both programs we found (or maybe more, if necessary). This could be considered an additional fix of this PR to make it work.

It's a good idea to discuss it in fltk.coredev with all implications.

erco77:
Saw your recent comment, sounds good.

I looked through the fltk.general and fltk.coredev, found no opinions. Sounds like apps set this at the top of main, but no one seemed sure about anything. I ran into this crap a few times (forgot!), including on the recent calandar widget, which ALSO got hung up in all the gory details of internationalization/language stuff, lol. (fltk.general, Subject: "RFC: adding a calendar widget to FLTK?", my post July 25 2019).

As a side note, the "signal11" bluetooth lib I mentioned above tries to not call setlocale() unless it hasn't been set before in their init function using this double-call technique:

```
       // Set the locale if not already
       if ( ! setlocale(LC_CTYPE, NULL) )  setlocale(LC_CTYPE, "");
```

Albrecht-S:
WRT Windows error message formatting: `FormatMessageA()` is the wrong call - it would return an _"ANSI string"_ (single byte, current codepage).

What we need is `FormatMessageW()` which returns a _"wide character string"_ and then convert the string to UTF-8.

I'm working on a fix and I will test it on a native Windows box with Visual Studio when done.

Albrecht-S:
I pushed the commit (88fe2f3701) to branch issue_99. I verified that the conversion really gives us correct UTF-8 characters. Some difficulties arose because VS2010 didn't like initialization inside the loop header in C code (I commented the commit). I also fixed some C++ style `//` comments.

For more info see the commit message. Error messages with non-ASCII characters tested on Win10 (German) with VS2010 successfully.

Albrecht-S:
Side note: I suggest to rename this branch to issue-99 (with dash/minus/hyphen `'-'`) instead of underline. This enables automatic CI builds because our configuration matches only some branch names - other names don't trigger Travis-CI builds. See [.travis.yml](https://github.com/fltk/fltk/blob/master/.travis.yml).

If you do this you could also create a new PR with the new branch `issue-99` and we could refer to this one and delete the old `issue_99` branch. Just to try how it works -- this PR is connected to my original PR on branch [str-1086_file-chooser](https://github.com/Albrecht-S/fltk/tree/str-1086_file-chooser) which is obsolete and I have no idea how to **change** this, other than creating a new PR.

erco77:
> What we need is FormatMessageW() which returns a "wide character string" and then convert the string to UTF-8.

Great -- thanks for doing that.

If you've confirmed that works on actual utf8 messages, I'll update the same code in fluid, which is where that came from.
(see the "NOTE:" at the top of the get_ms_errmsg() function) and update the branch.

_Digression: Perhaps to avoid the duplicate use of the same get_ms_errmsg() code in FLTK (fluid vs. here, and probably
other places if we look), it might be good if this get_ms_errmsg() function, or something like it, were public so fluid can use it.
Not in this issue, but as a separate RFE. So a vote fltk.coredev to add this somehow to the 1.4 FLTK API as a platform specific function, (I assume defined in FL/platform.H?) Perhaps a better, more flexible name fl_win32_format_message(errcode)
so that msgs for different error codes can be returned, just like strerror(errno)_

> Side note: I suggest to rename this branch to issue-99

OK, I'll see about doing that, and will follow up here when I do.
I guess I didn't know the Travis stuff would build branches too, but that's cool.

I guess because this 'issue' started as a pull request, it's kinda in limbo now since it's being managed as a branch,
and from what I can tell, a PR can't be converted to an 'issue' (but an issue can be turned into a PR).

erco77:
>>     Side note: I suggest to rename this branch to issue-99
> OK, I'll see about doing that, and will follow up here when I do.

OK, done. I think. Scary stuff, very unintuitive commands as usual, nothing's simple in git.
Here's what I did: with the issue_99 branch checked out:
```
git branch -m issue_99 issue-99
git push origin :issue_99 issue-99
git push origin -u issue-99
```
Don't ask me what that means, I have no idea. Just generally following directions that were here:
https://stackoverflow.com/questions/30590083/how-do-i-rename-both-a-git-local-and-remote-branch-name

Seems to have worked, 'git branch -a' shows "issue-99" now, no sign of the underbar version.
And on github, the list of branches only shows "issue-99" and no sign of the underbar version.

Albrecht-S:
[sorry, this reply was meant to be sent about 3 hours ago, just noticed that I didn't click the button]

The only thing we don't have now is that the PR gets automatically updated with the progress of its asscociated branch because this is still `Albrecht-S:str-1086_file-chooser` and I don't think we can change this.

So if we wanted to have a clean environment (i.e. a PR with an associated branch so we can see the "checks" alive), then we'd likely need a NEW PR with the correct issue-99 branch. We can leave this PR stale as it is (or for more OT discussions) and continue (if anything) in the new PR with the associated branch.

- If you like to do it, feel free to go ahead (create the new PR).
- If you want me to do it, please tell me.
- If you want it to stay as-is: nothing to do.

Albrecht-S:
OK, regarding the Git instructions: actually you can't "rename" a remote branch. You push the new branch and delete the old one. All the instructions in this SO article are more or less the same with similar commands.

Your commands again:
```
git branch -m issue_99 issue-99    -- renames the local branch
```

Yep, `-m` == move or rename (local) branch.

```
git push origin :issue_99 issue-99  -- somehow this deletes the old branch and renames it on remote %^/
```

A push command with two arguments (branches) in one. The `:branchname` notation is short for `local-branch:remote-branch` with an **empty** local branch, hence the remote branch is deleted. Same as `push -d` or `push --delete` in newer Git versions.
The second arg `issue-99` means that you push the local branch `issue-99` to the remote origin with the same name -- short form for `issue-99:issue-99`.

```
git push origin -u issue-99    -- resets upstream branch with new name, or something. syntax clear as mud
```

`-u` is the short form for `--set-upstream` and means that you make the local branch "track" the remote branch on origin. It also pushes it to the remote which means that you effectively pushed the branch twice, but the 2nd push is a no-op since everything is already up-to-date. "Tracking" means that `git pull` w/o arguments "knows" which remote to pull from.

I agree that the syntax is often not clear. Particularly the branch deletion in the old form instead of `-d` or `--delete`.

erco77:
So I ported your mods to fluid, and tested both fluid and file_chooser with mingw in the Japanese language:

issue-99-japanese-language-mingw-filechooser

issue-99-japanese-language-mingw-fluid

In the fluid test, I purposefully misspelled 'notepaddd' so it would fail.

So I think we're good.
Didn't test with VS 2017 yet, but when I do I'd call this problem solved.

I pushed the modifications to the remote branch. Only thing I changed in your code was removing a stray ^M
that must've come from your win32 editor (commit cfaa39cf85). My fluid mods came after that (commit f0e328f338)

> So if we wanted to have a clean environment (i.e. a PR with an associated branch so we can see the "checks" alive), then we'd likely need a NEW PR with the correct issue-99 branch. We can leave this PR stale as it is (or for more OT discussions) and continue (if anything) in the new PR with the associated branch.
> 
>     If you like to do it, feel free to go ahead (create the new PR).
>     If you want me to do it, please tell me.
>     If you want it to stay as-is: nothing to do.

Hmm, not that I'm any expert on this, but would it not make sense to simply apply my branch to your fork, and accept the PR once it passes the Travis test), so that the PR is solved in the normal way? Then perhaps the entire branch can be deleted off the fltk remote as if it never happened (not sure if that can be done or not), and simply merge the whole shebang with the PR?

You would surely be better at all this than I would, as there's probably a way to preserve the commit history on my branch into your PR.

If that doesn't make sense, I can just merge my branch with master using those merge commands you mentioned, and we can close this PR citing whatever the commit for that merge is, and close this PR, the hope being Travis doesn't barf on the merge result, lol.

Albrecht-S:
@erco77 Please take a look at your private mails. I sent you a few mails during the last hour.

Albrecht-S:
Note: I commented on your commit (f0e328f3380), you missed some changes (see comments). I wonder how this could work in your tests... (with `FormatMessageA()` instead of `FormatMessageW()` ).

Albrecht-S:
I did as suggested by you: pushed branch `issue-99` to my branch `Albrecht-S:str-1086_file-chooser`. [Travis-CI is running](https://github.com/fltk/fltk/runs/867168566)...

Here's what I did (note `issue-99:str-1086_file-chooser`):
```
(on branch issue-99)
$ git push -f albrecht issue-99:str-1086_file-chooser
Total 0 (delta 0), reused 0 (delta 0)
To github.com:Albrecht-S/fltk.git
 + ad6566fad...f0e328f33 issue-99 -> str-1086_file-chooser (forced update)
```

However, we still need to fix fluid (see my previous comment).

erco77:
Hmm, I had an intermediate copy that must have had some in progress work get into the repo. I'll check.
I know I did all that stuff you mentioned.

erco77:
OK, committed the proper fluid file (15a56387724), cloned it back to my windows8 machine, and rebuilt from scratch just to make sure it built, and it does, so should be right this time.

Albrecht-S:
Please follow up on the new PR #103

erco77:
Yep, saw #103 and replied.
I take it you'll close this PR #99 in some kind of state that indicates it's been moved to PR #103.

Albrecht-S:
The original branch `Albrecht-S:str-1086_file-chooser` of this PR has been obsoleted by the new branch `issue-99a` which is a rebased version of our working branch `issue-99`.

Branch `issue-99a` has been merged in PR #103 and I will now delete the obsolete (working) branches mentioned above to clean up.

I believe that this PR will automatically be closed by GitHub when the underlying branch (`Albrecht-S:str-1086_file-chooser`) is deleted. We'll see ... (going ahead deleting the two branches) ...

Albrecht-S:
FTR: I didn't close this PR explicitly. It was closed automatically (by GitHub) when I deleted the underlying branch in my fork (as I wrote above). I also deleted branch `issue-99` in `fltk/fltk`.

PR #110, opened 2020-07-18
save fl_file_chooser directory on cancel (STR3225)

engelsman:
by default, the fl_file_chooser() starts in the current directory
if no filename is given, but the directory is reset to the "system
directory" if the dialog is cancelled without choosing a file.
This patch saves and restores the directory if no file is chosen.

See also https://www.fltk.org/str.php?L3225

### DIFF
--- a/src/fl_file_dir.cxx
+++ b/src/fl_file_dir.cxx
@@ -93,11 +93,16 @@ fl_file_chooser(const char *message,    // I - Message in titlebar
   static char   retname[FL_PATH_MAX];           // Returned filename
 
   if (!fc) {
+    // first time, so create file chooser, remember pointer,
+    // and if no filename given, set it to current directory
+
     if (!fname || !*fname) fname = ".";
 
     fc = new Fl_File_Chooser(fname, pat, Fl_File_Chooser::CREATE, message);
     fc->callback(callback, 0);
   } else {
+    // file chooser exists, so check old/new directory, pattern, filename
+
     fc->type(Fl_File_Chooser::CREATE);
     // see, if we use the same pattern between calls
     char same_pattern = 0;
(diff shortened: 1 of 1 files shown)


src/fl_file_dir.cxx
```diff
@@ -138,8 +138,15 @@ fl_file_chooser(const char *message,    // I - Message in titlebar
         *retname = 0;
       const char *n = fl_filename_name(retname);
       if (n) *((char*)n) = 0;
-      fc->value("");
-      fc->directory(retname);
+      if (*retname) {
+        fc->value("");
+        fc->directory(retname);
+      } else {
```
Albrecht-S, line 144: I suggest a comment like "// user cancelled" here

engelsman:
I have now added comments to the fl_file_chooser [re]initialization code, and realize that I don't understand why the `!fname` and `!*fname` cases are treated differently:

- If `!fname` then if the new and old filter patterns are the same, the previous selected filename is reused, otherwise just its directory
- if `!*fname`  then the patterns are not checked, an empty filename is set, and the previous directory is reused.

Is there any historical reason why these two options couldn't be merged into one?

Albrecht-S:
Thanks, comments look good.

I have no idea why these two cases are handled differently, and I'm not going to dive into it. Sorry, I have nothing to say to your question.

If in doubt, I suggest to leave it as is.

Albrecht-S:
Note the pending checks seem to be cause by a hickup in the communication between GitHub and TravisCI, so nothing to worry about. The build completed successfully.

@engelsman @ManoloFLTK @erco77  Do you see any issues with the proposed patch?

I repeat my question though: are we good to go?
I'd like to merge and close.

engelsman:
Well it addresses the pixmap_browser problem reported in the STR, so it's already an improvement, but I wasn't sure whether the other branch might need to be considered as well, hence the question.

Albrecht-S:
If it fixes the issue mentioned in the STR we can commit/merge and close the STR. So far, so good.

Is there anything wrong with that "other branch" or is it only "cosmetics"?
Note I wrote before:
> If in doubt, I suggest to leave it as is.

I can't decide because I'm not willing to dig deeper (lack of time, other issues and PR's). So if you think it's done, let's go. That's my 2 ct. Fix the STR and close this PR.

ManoloFLTK:
OK for me.

Albrecht-S:
I'm going to merge this now...

@engelsman Since [STR 3225](https://www.fltk.org/str.php?L3225) is assigned to you, would you please close it or add appropriate comment(s) if necessary?

PR #111, opened 2020-07-18
Simplify, clarify and fix bugs in test/demo.cxx

Albrecht-S:
This PR clarifies the special cases in `test/demo.cxx` regarding different build systems and platform specific code.

**The goal is to de-duplicate code and to reduce platform specific code to a minimum.**

I also added a big comment describing the different directory layouts of the different build systems for future development. See [test/demo.cxx](https://github.com/Albrecht-S/fltk/blob/b482fcd0e058106a7837f60ca84b9be859a3532d/test/demo.cxx#L20).

Before this commit some of the `test/demo` programs could not be started in some builds (e.g. `Visual Studio`) because the paths differ between the classic (`autoconf/make`) and the new `CMake` generated build systems.

This should all be fixed now. I could test successfully under Windows and Linux but I'd like to see confirmations from others (see also the commit comments).

Summary of important changes:
- For `CMake` builds I removed the '`/examples`' folder so all executable (fluid and test/demo) programs are now directly in the `/bin` folder of the build root.
- I also added a new folder `/data` in the build root (same level as `/bin`). All test support (data) files including `demo.menu` are now in this folder to keep them apart from the executables.
- [Edit] There's no longer an executable path `../fluid` in `demo.menu` - this is handled in the demo program depending on the build system (**bug fix**). [/Edit]

For all other changes please refer to the source file `test/demo.cxx`.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -15,8 +15,8 @@
 #
 #######################################################################
 
-set (EXECUTABLE_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/../bin/examples)
-set (TESTFILE_PATH ${EXECUTABLE_OUTPUT_PATH})
+set (EXECUTABLE_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR}/../bin)
+set (TESTFILE_PATH ${CMAKE_CURRENT_BINARY_DIR}/../data)
 
 include (../CMake/FLTK-Functions.cmake)
 include (../CMake/fl_create_example.cmake)
@@ -204,13 +204,11 @@ endif (NOT ANDROID)
 
 # We need some support files for the demo programs:
 
-# Note: this is incomplete as of 11 Feb 2015
-# Todo: currently all files are copied, but some of them need configuration:
-# - demo.menu: fluid can't be started (wrong path)
+# Note: this is incomplete as of July 2020
+# Todo: currently all files are copied, but some of them may need configuration:
 # - demo.menu: help_dialog (help_dialog.html) can't find its images (not copied)
-# - maybe more ...
 
-# create data directory for test and demo programs
+# create data directory for test and demo programs if it doesn't exist
 file (MAKE_DIRECTORY ${TESTFILE_PATH})
 
 # copy the test files
@@ -221,12 +219,15 @@ file (COPY
   DESTINATION ${TESTFILE_PATH}
 )
 
+# the main test program 'demo' needs additional hints and configurations
+
+target_compile_definitions (demo PRIVATE GENERATED_BY_CMAKE)
+target_compile_definitions (demo PRIVATE CMAKE_SOURCE_PATH="${CMAKE_CURRENT_SOURCE_DIR}")
+
 # Apple macOS creates bundles instead of executables and needs a little bit
 # more help for demos to run correctly
 
 if (APPLE AND (NOT OPTION_APPLE_X11) AND (NOT OPTION_APPLE_SDL))
-
-  target_compile_definitions (demo PUBLIC USING_XCODE)
 
   # make the menu structure part of the app
   target_sources (demo PRIVATE demo.menu)
--- a/test/demo.cxx
+++ b/test/demo.cxx
@@ -14,9 +14,58 @@
 //     https://www.fltk.org/bugs.php
 //
 
+/*
+  General information on directory structure and file handling.
+
+  The "classic" autotools/make system creates executables in their source
+  folders, i.e. fluid/fluid, test/demo and test/xyz, resp.. The menu file is
+  in folder test/, as is the main demo(.exe) program. In the following text
+  and directory lists all test and demo executables are represented by "demo"
+  and the fluid executable by "fluid", no matter what OS (under Windows: *.exe).
+
+  The CMake build system generates all executables in the build tree and copies
+  the supporting test data files to the build tree as well. This structure is
+  different and needs to be handled separately in this program.
+
+  Additionally, different OS platforms create different types of files, for
+  instance "app bundles" on macOS. All this needs to be considered.
+
+  The overall structure, relative to the FLTK source dir (fltk) and the build
+  tree (build):
+
+  (1) Autotools / Make:
+
+    fltk/fluid              fluid (../fluid/fluid)
+    fltk/test               demo, demo.menu, working directory, data files
+    fltk/documentation/src  images for help_dialog(.html)
+
+  (2) CMake + make (e.g. Unix)
+
+    build/bin               fluid, demo
+    build/data              demo.menu, working directory, data files
+
+  (3) CMake + Visual Studio (TYPE == build type: Debug, Release, ...)
+
+    build/bin/TYPE          fluid, demo
+    build/data              demo.menu, working directory, data files
+
+  (4) CMake + macOS + Xcode
+
+    *FIXME*                 special code to handle bundles: do we need this?
+
+  The built executable 'demo' can also be executed with the menu filename
+  as commandline argument. In this case all the support (data) files are
+  expected to be in the same directory as the menu file or relative paths
+  as needed by the test programs, for instance help_dialog which needs
+  help_dialog.html and related image files.
+*/
+
+#define DEBUG_VARS (1)  // 1 = output variables to stderr, 0 = no
+
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include <errno.h>
 
 #if defined __APPLE__
 #include <ApplicationServices/ApplicationServices.h>
@@ -29,21 +78,7 @@
 #include <FL/Fl_Choice.H>
 #include <FL/filename.H>
 #include <FL/platform.H>
-
-/* Define a macro to decide whether a trailing 'd' needs to be removed
-   from the executable file name. Previous versions of Visual Studio
-   added a 'd' to the executable file name ('demod.exe') in Debug
-   configurations that needed to be removed.
-   This is no longer true with CMake-generated IDE's since FLTK 1.4.
-   Just in case we add it again: leave macro DEBUG_EXE_WITH_D defined
-   and leave the code using this macro as-is.
-*/
-
-// #if defined(_MSC_VER) && defined(_DEBUG) // Visual Studio in Debug mode
-// # define DEBUG_EXE_WITH_D 1
-// #else
-# define DEBUG_EXE_WITH_D 0
-// #endif
+#include <FL/fl_ask.H> // fl_alert()
 
 /* The form description */
 
@@ -56,6 +91,40 @@ void doscheme(Fl_Choice *c, void *) {
 
 Fl_Double_Window *form;
 Fl_Button *but[9];
+
+// Allocate space to edit commands and arguments from demo.menu.
+// We "trust demo.menu" that strings don't overflow
+
+char cmdbuf[256];           // commandline w/o arguments
+char params[256];           // commandline arguments
+
+// Global path variables for all platforms and build systems
+// to avoid duplication and dynamic allocation
+
+char src_path   [FL_PATH_MAX];          // directory of this souce file
+char app_path   [FL_PATH_MAX];          // directory of all demo binaries
+char fluid_path [FL_PATH_MAX];          // binary directory of fluid
+char data_path  [FL_PATH_MAX];          // working directory of all demos
+char command    [2 * FL_PATH_MAX + 40]; // command to be executed
+
+// platform specific suffix for executable files
+
+#ifdef _WIN32
+const char *suffix = ".exe";
+#elif defined __APPLE__
+const char *suffix = ".app";
+#else
+const char *suffix = "";
+#endif
+
+// debug output function (to stderr or ...)
+
+void debug_var(const char *varname, const char *value) {
+#if (DEBUG_VARS)
+  fprintf(stderr, "%-10s = '%s'\n", varname, value);
+  fflush(stderr); // Windows needs this
+#endif // DEBUG_VARS
+}
 
 void create_the_forms() {
   Fl_Widget *obj;
@@ -214,190 +283,145 @@ void dobut(Fl_Widget *, long arg) {
   int men = find_menu(stack[stsize-1]);
   int n = menus[men].numb;
   int bn = but2numb( (int) arg, n-1);
+
+  // menu ?
+
   if (menus[men].icommand[bn][0] == '@') {
     push_menu(menus[men].icommand[bn]);
+    return;
+  }
+
+  // not a menu: run test/demo/fluid executable
+  // find and separate "command" and "params"
+
+  // skip leading spaces in command
+  char *start_command = menus[men].icommand[bn];
+  while (*start_command == ' ') ++start_command;
+
+  strcpy(cmdbuf, start_command); // here still full command w/params
+
+  // find the space between the command and parameters if one exists
+  char *start_params = strchr(cmdbuf, ' ');
+  if (start_params) {
+    *start_params = '\0';         // terminate command
+    start_params++;               // skip space
+    strcpy(params, start_params); // copy parameters
   } else {
+    params[0] = '\0';             // empty string
+  }
+
+  // select application path: either app_path or fluid_path
+
+  const char *path = app_path;
+  if (!strncmp(cmdbuf, "fluid", 5))
+    path = fluid_path;
+
+  // format commandline with optional parameters
+
+#if defined(__APPLE__) // macOS
+
+  if (params[0]) {
+    // we assume that we have only one argument (which is a filename)
+    sprintf(command, "open '%s/%s%s' --args '%s'", path, cmdbuf, suffix, params);
+  } else {
+    sprintf(command, "open '%s/%s%s'", path, cmdbuf, suffix);
+  }
+
+#else // other platforms
+
+  if (params[0])
+    sprintf(command, "%s/%s%s %s", path, cmdbuf, suffix, params);
+  else
+    sprintf(command, "%s/%s%s", path, cmdbuf, suffix);
+
+#endif
+
+  // finally, execute program (the system specific part)
 
 #ifdef _WIN32
 
-    STARTUPINFO         suInfo;         // Process startup information
-    PROCESS_INFORMATION prInfo;         // Process information
+  STARTUPINFO         suInfo;         // Process startup information
+  PROCESS_INFORMATION prInfo;         // Process information
 
-# if DEBUG_EXE_WITH_D
-    const char *exe = "d.exe";          // exe name with trailing 'd'
-# else
-    const char *exe = ".exe";           // exe name w/o trailing 'd'
-# endif
+  memset(&suInfo, 0, sizeof(suInfo));
+  suInfo.cb = sizeof(suInfo);
 
-    memset(&suInfo, 0, sizeof(suInfo));
-    suInfo.cb = sizeof(suInfo);
+  debug_var("Command", command);
 
-    int icommand_length = strlen(menus[men].icommand[bn]);
-
-    char* copy_of_icommand = new char[icommand_length+1];
-    strcpy(copy_of_icommand,menus[men].icommand[bn]);
-
-    // On Windows the .exe suffix needs to be appended to the command
-    // whilst leaving any additional parameters unchanged - this
-    // is required to handle the correct conversion of cases such as :
-    // `../fluid/fluid valuators.fl' to '../fluid/fluid.exe valuators.fl'.
-
-    // skip leading spaces.
-    char* start_command = copy_of_icommand;
-    while (*start_command == ' ') ++start_command;
-
-    // find the space between the command and parameters if one exists.
-    char* start_parameters = strchr(start_command,' ');
-
-    char* command = new char[icommand_length+6]; // 6 for extra 'd.exe\0'
-
-    if (start_parameters==NULL) { // no parameters required.
-      sprintf(command, "%s%s", start_command, exe);
-    } else { // parameters required.
-      // break the start_command at the intermediate space between
-      // start_command and start_parameters.
-      *start_parameters = 0;
-      // move start_paremeters to skip over the intermediate space.
-      ++start_parameters;
-
-      sprintf(command, "%s%s %s", start_command, exe, start_parameters);
-    }
-
-    CreateProcess(NULL, command, NULL, NULL, FALSE,
-                  NORMAL_PRIORITY_CLASS, NULL, NULL, &suInfo, &prInfo);
-
-    delete[] command;
-    delete[] copy_of_icommand;
+  BOOL stat = CreateProcess(NULL, command, NULL, NULL, FALSE,
+                            NORMAL_PRIORITY_CLASS, NULL,
+                            NULL, &suInfo, &prInfo);
+  if (!stat) {
+    DWORD err = GetLastError();
+    fl_alert("Error starting process, error #%d\n'%s'", err, command);
+  }
 
 #elif defined __APPLE__
-    /*
-     Starting with version 1.4.0, FLTK uses CMake as the only supported build
-     system. On macOS, the app developer is expected to run CMake in a
-     directory named './build/Xcode' or './build/Makefiles' to generate the
-     build environment.
 
-     When building FLTK in the next step, teh macOS app bundles are then
-     stored in either:
-     './build/Xcode/bin/examples/hello.app/' for Makefiles
-     './build/Xcode/bin/examples/Debug/hello.app/' for XCode Debug
-     or
-     './build/Xcode/bin/examples/Release/hello.app/' as a symbolic
-     into the Archive system of macOS
+  debug_var("Command", command);
+  system(command);
 
-     'Demo' needs to find and run all of these app bundles, some requiring
-     an additional file name and path for resource files.
 
-     This is my attempt to find the bundles and resources so that Demo.app
-     and all its dependencies will run without any further configuration.
-     They will stop running however if any of the bundles or rresources
-     are moved.
-     */
-    {
-      char src_path[PATH_MAX];
-      char app_path[PATH_MAX];
-      char app_name[PATH_MAX];
-      char command[2*PATH_MAX+2];
-      char *cmd = strdup(menus[men].icommand[bn]);
-      char *args = strchr(cmd, ' ');
+#else // other platforms (Unix, Linux)
 
-      /*
-       Get the path to the source code at compile time. This is where the other
-       resources are located.
-       */
-      strcpy(src_path, __FILE__);
-      char *src_path_end = (char*)fl_filename_name(src_path);
-      if (src_path_end) *src_path_end = 0;
+  strcat(command, " &"); // run in background
+  debug_var("Command", command);
 
-      /*
-       All example app bundles are in the same directory as 'Demo', so set the
-       current dir to the location of Demo.app .
-
-       Starting with macOS 10.12, the actual location of the app has a randomized
-       path to fix a vulnerability. This still works in Debug mode which is
-       */
-      {
-        app_path[0] = 0;
-        CFBundleRef app = CFBundleGetMainBundle();
-        CFURLRef url = CFBundleCopyBundleURL(app);
-        CFStringRef cc_app_path = CFURLCopyFileSystemPath(url, kCFURLPOSIXPathStyle);
-        CFRelease(url);
-        CFStringGetCString(cc_app_path, app_path, 2048, kCFStringEncodingUTF8);
-        CFRelease(cc_app_path);
-        if (app_path[0]) {
-          char *app_path_end = (char*)fl_filename_name(app_path);
-          if (app_path_end) *app_path_end = 0;
-          fl_chdir(app_path);
-        }
-      }
-
-      // extract the executable name from the command in the menu file
-      strcpy(app_name, cmd);
-      // remove any additioanl command line arguments
-      if (args) app_name[args-cmd] = 0;
-      // make the file name into a bundle name
-      strcat(app_name, ".app");
-
-      if (args) {
-        if (strcmp(app_name, "../fluid/fluid.app")==0) {
-          // CMake -G 'Unix Makefiles' ... : ./bin/fluid.app
-          // CMake -G 'Xcode' ... : ./bin/Debug/fluid.app or ./bin/Release/fluid.app
-          // so removing the '/example' path segment from the app_path should
-          // always work.
-          char *examples = strstr(app_path, "/examples");
-          if (examples) {
-            memmove(examples, examples+9, strlen(examples+9)+1);
-          }
-          sprintf(command, "open '%sfluid.app' --args '%s%s'", app_path, src_path, args+1);
-        } else {
-          // we assume that we have only one argument which is a filename, so we add a path
-          sprintf(command, "open '%s%s' --args '%s%s'", app_path, app_name, src_path, args+1);
-        }
-      } else {
-        sprintf(command, "open '%s%s'", app_path, app_name);
-      }
-      system(command);
-      free(cmd);
-    }
-
-#else // Non Windows systems.
-
-    int icommand_length = strlen(menus[men].icommand[bn]);
-    char* command = new char[icommand_length+5]; // 5 for extra './' and ' &\0'
-
-    sprintf(command, "./%s &", menus[men].icommand[bn]);
-    if (system(command)==-1) { /* ignore */ }
-
-    delete[] command;
+  if (system(command) == -1) {
+    fl_alert("Could not start program, errno = %d\n'%s'", errno, command);
+  }
 
 #endif // _WIN32
-  }
+
 }
 
 void doback(Fl_Widget *, void *) {pop_menu();}
 
 void doexit(Fl_Widget *, void *) {exit(0);}
 
-/* Load the menu file. Returns whether successful. */
-int load_the_menu(char* fname) {
+/*
+  Load the menu file. Returns whether successful.
+
+  New strategy: the menu file *should* be usable as is!
+
+  Old strategy was:
+
+  We have different situations:
+
+  (1) Visual Studio: see main(): nothing to do
+  (2) Standard (autotools/make): nothing to do
+  (3) CMake: see main(): nothing to do
+
+  (4) macOS: ???    *FIXME*
+*/
+int load_the_menu(char *menu) {
   FILE *fin = 0;
   char line[256], mname[64],iname[64],cname[64];
   int i, j;
-  fin = fl_fopen(fname,"r");
-#if defined ( __APPLE__ )
+
+  fin = fl_fopen(menu, "r");
+
+//      // *FIXME* Albrecht: Do we need this? I don't think so!
+#if (0) // *FIXME* disabled for testing
+
+#if defined (__APPLE__)
   if (fin == NULL) {
-    // mac os bundle menu detection:
-    char* pos = strrchr(fname,'/');
+    // macOS bundle menu detection:
+    char *pos = strrchr(menu, '/');
     if (!pos) return 0;
     *pos = '\0';
-    pos = strrchr(fname,'/');
+    pos = strrchr(menu, '/');
     if (!pos) return 0;
-    strcpy(pos,"/Resources/demo.menu");
-    fin  = fl_fopen(fname,"r");
+    strcpy(pos, "/Resources/demo.menu");
+    fin = fl_fopen(menu, "r");
   }
-#endif
-  if (fin == NULL) {
+#endif // __APPLE __
+
+#endif // *FIXME* disabled for testing
+
+  if (fin == NULL)
     return 0;
-  }
+
   for (;;) {
     if (fgets(line,256,fin) == NULL) break;
     // remove all carriage returns that Cygwin may have inserted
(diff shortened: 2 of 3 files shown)


Albrecht-S:
I'd appreciate feedback from all devs, not only, but particularly for `macOS` builds with `Xcode` **and** `make`. Do all the tests work?

**Note**: the only test that is known to not work completely right now is `help_dialog` which can be started and displays the html file but does not find the included images -- because they are not copied to the build root. This is a known issue, please ignore.

If you find other bugs and issues, please feel free to report here and/or commit updates and push to [my fork](https://github.com/Albrecht-S/fltk/tree/feature-demo) in branch `feature-demo`.

As usual I **allowed** FLTK devs with **write access** to the repo to push to my fork!

Albrecht-S:
@ManoloFLTK Please test and verify the macOS specific "bundle detection" code and special handling. I removed some of the macOS code because most of it is now done in the main program but there's still code I don't understand and can't test.

Since the new CMake way is to not access (write to) the source folder my intention is to copy all necessary data files to the build folder - now in the `/data` directory. This should be the **working directory** for all test programs and the macOS bundles (resources) should be used from the CMake build root and not from the source folder to avoid to write to the source folder (e.g. test/editor, test/browser, maybe more).

In a future iteration of the code we might be able to remove the `source_path` variable entirely.

My strategy is to construct all necessary folder names in the main program, but the current source code uses a macOS specific code part to construct for instance the `app_path` variable and overwrites the already constructed string (because I didn't know if this special handling is necessary). If yes, then it might make sense to pull this code up into the main program where the `app_path` string is constructed because this string is used in other `path` variables later.

I'm not sure if my intention is clear. If you have questions, please feel free to ask here. If you have other suggestions for code improval, feel free to say.

Thanks to all in advance.

imaclmaca:
Makefile test on OSX.

Deleted my build dir, then used cmake to regenerate.
Tested examples, seem to be OK, though I had to use "open blah.app" to run them from the shell, as the wrapper scripts are not present (this, I think, is normal though!)

Observations: 
- All seems to work OK, AFAICT.
- The help_dialog *does* show the images (I note that my "regular" fltk-1.3 autoconf build does not.)
- I did not see the "/data" dir that Albrecht mentioned...

My Xcode is a bit messed up, so I have not tried it (this Mac is old and has dropped off the end of Apple's maintenance window some time ago...)

Albrecht-S:
@imaclmaca Thanks very much for testing

> I had to use "open blah.app" to run them from the shell, as the wrapper scripts are not present

OK, I have no idea what these wrapper scripts are. What does "run **them** (from the shell)" mean exactly? Did you run each single test app separately or did you start `demo.app` and click on the buttons to start the other apps? I'd be interested in the latter, i.e. if running the apps from **within** `demo.app` works, i.e. if `demo.app` itself does what it is intended to do.

I'm afraid I haven't been clear enough, sorry for that. Or did you do exactly this? If yes, please confirm; this didn't seem clear to me. Otherwise, can you test this as well?

In general I'm interested if the few apps that need test files are finding these files, which would indicate that the demo app sets the correct working directory in all cases. In your case a sequence like this ought to work:
```
$ cd fltk-root
$ make
$ open test/demo.app
```

The "interesting" apps are `test/editor`, `test/browser`, `test/help_dialog`, `test/color_chooser`, and `fluid/fluid valuator.fl` -- all executed from within demo.app as shown above because they need additional files.

I'm particularly interested in the CMake build, but your test is helpful as well. Thanks.

> The help_dialog does show the images

When you build with `make` **in-source** and use the `test/*` apps, this ought to work (but I can't tell if it ever worked on macOS). What I meant to say was that the `help_dialog` test app does not show the images when run from a `CMake` (**out of source**) build (because the images are not (yet) copied to the build folder).

> I did not see the "/data" dir that Albrecht mentioned

This `/data` dir is only generated by CMake in the build folder. When using `make` in the source dir we still use the existing dirs (which made this so complicated).

Thanks for testing anyway, I appreciate this.

imaclmaca:
> > I had to use "open blah.app" to run them from the shell, as the wrapper scripts are not present
> 
> OK, I have no idea what these wrapper scripts are. What does "run **them** (from the shell)" mean exactly? Did you run each single test app separately or did you start `demo.app` and click on the buttons to start the other apps? I'd be interested in the latter, i.e. if running the apps from **within** `demo.app` works, i.e. if `demo.app` itself does what it is intended to do.
> 

With the autoconf builds under OSX the "wrapper scripts" are generated by the build so that the OSX bundles can be run "directly" from the shell, e.g. consider the "inactive" test...

immpcX:test ian$ ls -d inactive*
inactive     inactive.app inactive.cxx inactive.fl  inactive.h   inactive.o
immpcX:test ian$ 
immpcX:test ian$ cat inactive
#!/bin/sh
dir="`dirname \"$0\"`"
exec "$dir/inactive.app/Contents/MacOS/inactive" "$@"
immpcX:test ian$ 

So the script just finds and invokes the binary embedded in the bundle; it is an alternative to saying "open bundle.app" in effect.

> I'm afraid I haven't been clear enough, sorry for that. Or did you do exactly this? If yes, please confirm; this didn't seem clear to me. Otherwise, can you test this as well?
>

I did a few things...

- I launched demo.app from the shell with "open demo.app", then from within the demo ran all the accessible tests.

- from the shell I ran the additional tests that demo.app does not include.

- from the shell I re-ran some of the apps that demo does include, but just as a check; in particular help_dialog and editor. When invoked from within the demo.app, both displayed their content (including the fltk images in help_dialog) correctly. When invoked from the shell editor is blank, and help_dialog displays the text but not the images.

> In general I'm interested if the few apps that need test files are finding these files, which would indicate that the demo app sets the correct working directory in all cases. In your case a sequence like this ought to work:
> 
> ```
> $ cd fltk-root
> $ make
> $ open test/demo.app
> ```
> 

Ah, perhaps we are at cross-purposes: I did not do an autoconf build, I did a cmake build, just using cmake to generate the actual Makefiles for the build (I don't really use the Xcode IDE method, ever.)


> The "interesting" apps are `test/editor`, `test/browser`, `test/help_dialog`, `test/color_chooser`, and `fluid/fluid valuator.fl` -- all executed from within demo.app as shown above because they need additional files.
> 

editor : OK from demo, fails from shell
browser : OK from demo, OK from shell
help_dialog : OK from demo, partial from shell
cooler_chooser : OK from demo, OK from shell
fluid : OK from demo, fails from shell
valuators : OK from demo, OK from shell

> I'm particularly interested in the CMake build, but your test is helpful as well. Thanks.
> 

Note that this was a cmake build, in essence.

> > The help_dialog does show the images
> 
> When you build with `make` **in-source** and use the `test/*` apps, this ought to work (but I can't tell if it ever worked on macOS). What I meant to say was that the `help_dialog` test app does not show the images when run from a `CMake` (**out of source**) build (because the images are not (yet) copied to the build folder).
> 

I just tried an autoconf build of 1.4 - test/help_dialog does not show the images. Nor does "help" from 1.3. Both show the text OK however.


> > I did not see the "/data" dir that Albrecht mentioned
> 
> This `/data` dir is only generated by CMake in the build folder. When using `make` in the source dir we still use the existing dirs (which made this so complicated).
> 

This *was* a cmake generated build, and I do not seem to have a data directory in my build folder.

Albrecht-S:
OK, I understand: you used CMake to generate "Unix Makefiles" (?). If yes (or if no, of course) I have two questions:

1. Did you really pull from my fork `Albrecht-S/fltk` and branch `feature-demo` ([Albrecht-S:feature-demo](https://github.com/Albrecht-S/fltk/tree/feature-demo)) ?

This branch should have these instructions in test/CMakeLists.txt:
```
set (TESTFILE_PATH ${CMAKE_CURRENT_BINARY_DIR}/../data)
...
file (MAKE_DIRECTORY ${TESTFILE_PATH})
```
If this is not the case, then you didn't pull the right branch.

`${CMAKE_CURRENT_BINARY_DIR}` is the `test/` directory in the **build tree**, hence the `/data` directory should be created as `$build_root/data`. If it's an in-source build (not recommended!) then it should be at the same level as the `/test` directory.

2. How exactly did you invoke CMake? Is this an in-source build or out-of-source-tree, e.g. in a `/path/to/fltk/build` directory?

Thanks for your help, again.

imaclmaca:
> OK, I understand: you used CMake to generate "Unix Makefiles" (?). If yes (or if no, of course) I have two questions:
> 

Yes, cmake to generate "Unix Makefiles"


> 1. Did you really pull from my fork `Albrecht-S/fltk` and branch `feature-demo` ([Albrecht-S:feature-demo](https://github.com/Albrecht-S/fltk/tree/feature-demo)) ?
> 

OK - until you asked, I was sure I had. Now the seed of doubt is sown... I need to go check...

It may not be until later, or tomorrow, now though (sorry.)


> 1. How exactly did you invoke CMake? Is this an in-source build or out-of-source-tree, e.g. in a `/path/to/fltk/build` directory?
> 

My "build" dirs are typically just subdirectories of the fltk-root. 
On OSX I *usually* only have one build sub-for. 
(On other platforms I usually have several, e.g. for pango / 32-bit / 64-bit / etc.)

Albrecht-S:
> OK - until you asked, I was sure I had. Now the seed of doubt is sown... I need to go check...

Here's another test and a question: the new demo program ought to output some paths (test statements) to stderr. I hope you see stderr on macOS. (?) And all the `test/demo` executables must be directly in `bin/ `side by side with `fluid(.app)`. From my tests underLinux (I'm using `/build/build-type` subdirs):
```
(fltk root  = /data/git/fltk/feature-demo/)
(build root = /data/git/fltk/feature-demo/build/Debug/)
/data/git/fltk/feature-demo/build/Debug$ bin/demo 

src_path   = '/git/fltk/feature-demo/test'
app_path   = '/data/git/fltk/feature-demo/build/Debug/bin'
fluid_path = '/data/git/fltk/feature-demo/build/Debug/bin'
data_path  = '/data/git/fltk/feature-demo/build/Debug/data'
Menu file  = '/data/git/fltk/feature-demo/build/Debug/data/demo.menu'

Command: '/data/git/fltk/feature-demo/build/Debug/bin/fluid valuators.fl &'
Command: '/data/git/fltk/feature-demo/build/Debug/bin/blocks &'
Command: '/data/git/fltk/feature-demo/build/Debug/bin/fractals &'
Command: '/data/git/fltk/feature-demo/build/Debug/bin/unittests &'
Command: '/data/git/fltk/feature-demo/build/Debug/bin/offscreen &'
. . .
```
MinGW cross-build under Linux for comparison:
```
/data/git/fltk/feature-demo/build/mingw$ wine bin/demo.exe 

src_path   = '/git/fltk/feature-demo/test'
app_path   = 'Z:/data/git/fltk/feature-demo/build/mingw/bin'
fluid_path = 'Z:/data/git/fltk/feature-demo/build/mingw/bin'
data_path  = 'Z:/data/git/fltk/feature-demo/build/mingw/data'
Menu file  = 'Z:/data/git/fltk/feature-demo/build/mingw/data/demo.menu'

Command: 'Z:/data/git/fltk/feature-demo/build/mingw/bin/unittests.exe'
Command: 'Z:/data/git/fltk/feature-demo/build/mingw/bin/offscreen.exe'
Command: 'Z:/data/git/fltk/feature-demo/build/mingw/bin/blocks.exe'
```

You should see all paths and the "Command" output multiple times in its macOS form.

I just committed in a minor fix to output the "command" on all platforms correctly (commit 67f54689631cff). Please pull again from the branch and you should see this commit:
"Fix: Better "Command:" test output on all platforms".

Please post the output here and report if it works. Thanks.

Albrecht-S:
... and for completeness the output in the console window of my VS-2019 (Debug) build:
```
src_path   = 'C:/Users/albrecht/git/fltk/test'
app_path   = 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug'
fluid_path = 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug'
data_path  = 'C:/Users/albrecht/git/fltk/build/vs2019/data'
Menu file  = 'C:/Users/albrecht/git/fltk/build/vs2019/data/demo.menu'

Command: 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug/unittests.exe'
Command: 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug/offscreen.exe'
Command: 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug/fluid.exe valuators.fl'
Command: 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug/checkers.exe'
Command: 'C:/Users/albrecht/git/fltk/build/vs2019/bin/Debug/browser.exe browser.cxx'
```

Just in case someone wants to test and compare. All executables work but (as said before) no images in `help_dialog` which is (currently) as expected.

Note: I did not yet test with MinGW / Mingw-w64 / MSYS2 under Windows. If someone could help this would be appreciated as well.

erco77:
OK, I did:
```
git checkout feature-demo
mkdir build-osx-10.10.5
cd build-osx-10.10.5
cmake -DOPTION_USE_SYSTEM_LIBJPEG=off -DOPTION_USE_SYSTEM_ZLIB=off -DOPTION_USE_SYSTEM_LIBPNG=off ..
cd bin
open demo.app
```
All demos worked except as noted below:

- Had to use `open demo.app` to run it because there's no `demo` executable. (There usually is)
- Fl_Browser demo throws an alert "Can't load <top-level-fltk-dir>/testbrowser.cxx"
  _This is probably because when a program is opened with 'open' the cwd is unknown, or some such._
- "Fluid (UI design tool)" opened but failed to load the "testvaluators.fl" file because it's looking in the top level directory for it (e.g. one up from the build directory)
- The "X Color Browser" demo opens OK, but the path in the top widget is too large and overruns the edges of the black box and even the window edges.
- The checkers demo looks too awesome ;-)
- Help dialog can't open the html file (again, looking in top level dir for it)

I think that tests all of them.

Albrecht-S:
@erco77 Thanks, Greg, for your test and report. I would've been surprised if it had worked OOTB. There's too much code I don't understand (and can't test) involved, particularly some "bundle detection" code and the code to set the current working directory (which is IMHO wrong, but I didn't want to touch it).

> "Can't load [top-level-fltk-dir]/testbrowser.cxx"

This should obviously read "[top-level-fltk-dir]/test/browser.cxx" (missing '/'), but IMHO it should not at all touch the fltk source dir (this is old code). My intention was and is that it should set the program's working dir to `build-root/data` where all files can be found. Again, there's old code I didn't touch because I couldn't test, and I presume that this old code did indeed try to load the support files from the fltk root dir.

> "Fluid (UI design tool)" opened but failed to load the "testvaluators.fl" file ...

Again, missing '/' ?

Do you see the test output (path and command) on stderr as posted by me above? If yes, can you post them here? TIA.

Albrecht-S:
> So basically anything that tries to open a file has trouble. Probably due to the cwd being set oddly somehow. Not sure why it's defaulting to the top level fltk dir all the time, huh.

That's exactly the error. The cwd should be <cmake-build-root>/data since my modifications in this branch, but the macOS code needs some fixes. If you want to take a look, see `FIXME` comments in the `__APPLE__` source part(s). Can you post the test output, please? TIA

erco77:
`Can you post the test output, please?`
There is no test output due to the use of `open demo.app`, which like win32 /subsystem:windows has no output (for the same reasons).
Would need to use Fl_Simple_Terminal (see my comment above) or something similar to show the output.

Albrecht-S:
> You may want to print the value of getcwd() as well, since that inherits down to the children and is possibly relevant.

I'm aware that there's something going awry but I can't test on macOS. I need help.

> Unfortunately on OSX when you use 'open some.app', there's no tty output ...

Can you redirect `stderr`? Probably not, right?

Sorry, I'm lost here. Someone else needs to fix the macOS code. I hope that you or @ManoloFLTK can help. If you have questions, please feel free to ask here. Thanks.

Albrecht-S:
> Would need to use Fl_Simple_Terminal (see my comment above) or something similar to show the output.

Maybe you could open a file and write the debug statements to the file instead of stderr.

> I have a visitor coming soon, so I may not be able to respond till later/tomorrow.

No problem, it's late here (bedtime). I'll be offline in a few minutes.

erco77:
> I'm aware that there's something going awry but I can't test on macOS. I need help.

If it were me, I'd refit the demo program to have either a debug mode or a little subtle "+Terminal" button at the bottom (or right click menu option) that enables an Fl_Simple_Terminal to appear below the bottom of the demo window, like our other demos that do tty output, and print to that. This way error output from demo can be seen.

If I have time I can do that, as it seems to be needed. Better than hacking away at the code to redirect to a file.

Albrecht-S:
Some notes on the code and the issues I'm aware of if someone is able to help with macOS. Please read the long comment at the beginning of `test/demo.cxx` first to see how the files and folders should be in the new CMake build system.

I'll show some lines of code and add my comments here:

> [Line 369] All example app bundles are in the same directory as 'Demo', so set the current dir to the location of Demo.app .

The first part is true: all apps including fluid are now in the same directory. The second part is wrong though: the current dir **should** be set to the `/data` dir of the build root where all helper files can be found. This is done successfully in the main program for all other platforms, but I don't know if (or how) this works under macOS because the "randomized path" mentioned by this (obviously incomplete) comment:

> [Line 372] Starting with macOS 10.12, the actual location of the app has a randomized path to fix a vulnerability. This still works in Debug mode which is [NOTE: comment ends here!]

If this (the **randomized path**) is true, then my attempt to derive all other paths from the executable file path (in `argv[0]`) in the `main()` program (Lines 536++) **may** fail under macOS. This is where we need a debugger or the test output or someone who "knows".

All the code in lines 386 - 396 is [now] questionable macOS specific code, and the `fl_chdir()` in line 396 is presumably wrong (it should be `data_path`). There's already code in main() to set the cwd to `data_path`:
```
  if (fl_chdir(data_path) == -1) { /* ignore */ }
```

The main question is: is `data_path` at this point (line 622 in main) correct under macOS? Does the code in main() work under macOS as well? If yes, then **all the questionable code in lines 385 - 398 (see below) could probably be deleted**.

If not I'd try to pull the macOS specific _"magic code"_ up to `main()` and adjust the path constructions in main, beginning with `app_path` in line 536. The rest of the paths might then be fine. Or not, I don't know. :-(

Here the (old) code (currently lines 385 - 398) for reference (what exactly is it good for?):
```
    {
      app_path[0] = 0;
      CFBundleRef app = CFBundleGetMainBundle();
      CFURLRef url = CFBundleCopyBundleURL(app);
      CFStringRef cc_app_path = CFURLCopyFileSystemPath(url, kCFURLPOSIXPathStyle);
      CFRelease(url);
      CFStringGetCString(cc_app_path, app_path, 2048, kCFStringEncodingUTF8);
      CFRelease(cc_app_path);
      if (app_path[0]) {
        char *app_path_end = (char*)fl_filename_name(app_path);
        if (app_path_end) *app_path_end = 0;
        fl_chdir(app_path);
      }
    }
```

Again, `fl_chdir(app_path);` is wrong here and should either be `fl_chdir(app_path);` or omitted because it is already set in `main()`.

Albrecht-S:
Okay, I took a stab at fixing the macOS code as I suggested above:
> If not I'd try to pull the macOS specific "magic code" up to main() and adjust the path constructions in main, beginning with app_path in line 536.

I believe I also found the "`missing '/'`" issue and fixed it as well (hopefully). Please give the new commit (4cb1249f8eea0a) a try. Thanks.

Albrecht-S:
There's another commit: 8ccbc213a539e. Please pull and try this one.

erco77:
Just pulled and rebuilt, but about to go to sleep.
Seemed to solve the Fl_Browser problem with cmake + OSX 10.10.15 + open demo.app,
so you probably fixed it. Will give it a more full check tomorrow.

I do have a modified demo with the tty debug stuff, but didn't want to check it in b/c
just noticed your new demo code, so will hold till tomorrow.

Albrecht-S:
> I do have a modified demo with the tty debug stuff, but didn't want to check it in b/c just noticed your new demo code, so will hold till tomorrow.

Sorry if I broke your tty debug stuff, but I felt that another update was necessary. I took the opportunity to consolidate debug statements: there's now exactly one function outputting the interesting vars which you could hopefully hook in to do the tty output instead.

I also rebased all my stuff on the latest `master` which includes the resize demos. Select
- Images for Manual ...
- Resize Examples ...

I'm optimistic that this new update works on macOS but there can be some bugs, of course (untested code).

Travis-CI is still running the [macOS build](https://travis-ci.com/github/fltk/fltk/jobs/362881012).

erco77:
Hmm, something went sideways on the mac build since last I checked.
I updated to your current commit, 'Fix macOS build/Sorry for the noise, copy error.'
Even re-cloned and checked out the branch to be sure.

The Fl_Browser demo now complains with an error saying can't load "browser.cxx", no path prefix at all now.

This is with the cmake build, and using 'open' to run the demo.app.

I can't see error output tho, so I'm going to commit my mods for adding Fl_Simple_Tty to the demo. I've had to merge it a few times now, so it'd be good to just get this in there.

Stand by, and I'll push that to your fork.

erco77:
OK, so with the Fl_Simple_Tty mods (that adds a right-click menu to show the debug terminal),
here's the results on the mac using cmake + open demo.app:
```
src_path   = '/net/tmp/fltk-albrecht-demos/test'
app_path   = '/seriss/net/tmp/fltk-albrecht-demos/build-osx-10.10.5/bin'
fluid_path = '/seriss/net/tmp/fltk-albrecht-demos/build-osx-10.10.5/bin'
data_path  = '/seriss/net/tmp/fltk-albrecht-demos/build-osx-10.10.5/data'
Menu file  = '/seriss/net/tmp/fltk-albrecht-demos/build-osx-10.10.5/data/demo.menu'
Cwd        = '/'

Command    = 'open '/seriss/net/tmp/fltk-albrecht-demos/build-osx-10.10.5/bin/browser.app' --args 'browser.cxx''
```

erco77:
So I'm done tweaking for now.
I'll await to see if you want to try fixing the mac path issue, or if you want me to look into it.
Seems to be a simple regression somehow, figured you'd be better to know what changed.

imaclmaca:
> > OK - until you asked, I was sure I had. Now the seed of doubt is sown... I need to go check...
> 
> Here's another test and a question: the new demo program ought to output some paths (test statements) to `stderr`. I hope you see `stderr` on macOS. (?) And all the `test/demo` executables must be directly in `bin/` or `bin/Debug` or whatever subdir side by side with `fluid(.app)`. From my tests under Linux (I'm using `/build/buildtype` subdirs):
> 

OK - my original checkout was messed up (I have no clue what I did...) So deleted it and made a new one, this is...

ian$ git branch -a
* feature-demo
  master
  remotes/origin/Fl_Image_release
  remotes/origin/HEAD -> origin/master
  remotes/origin/branch-1.0
  remotes/origin/branch-1.1
  remotes/origin/branch-1.3
  remotes/origin/feature-demo
  remotes/origin/feature-text-display-bgcolor
  remotes/origin/feature-travis-branches
  remotes/origin/fix-prefs
  remotes/origin/master
  remotes/origin/refactor_Fl_Windows_i
  remotes/origin/str-3449_window-skip-taskbar

Which I think is the correct branch now!


Anyway; made a build folder, then cmake -G "Unix Makefiles .. and build.

The data dir was created. The executables are in the bin folder (the examples sub folder we used to have is now gone.)

So that much is correct.

(Note: if you "open" a file on MacOS from the terminal, you do not get stdio. One advantage of our old wrapper scripts was that you could still get stdio in tests...!)

In testing, apps run OK if double-clicked in the finder, or if "open"ed. 

Help_dialog loads its text, but not the images. Editor, etc., do NOT load their content.

Demo.app does seem to load its menu, whether clicked or "open"ed.

In neither case do help_dialog et al seem to get the correct content. Invoking help_dialog from demo.app is, if anything, worse than just invoking it directly, so there must be some settings difference...

imaclmaca:
FWIW, I have (somewhere) code I wrote that has platform specific hooks to allow an executable to determine its "home" directory.
I used this in the past because neither OSX nor *some* (not all) X11 WM's were setting cwd to anything useful when my app started up.
Windows always set it to the "home" directory of the executable, AFAICT, so I added code so that the other platforms did too.
I then used that to figure out where to load the config files from...

We could use that here?

imaclmaca:
FTR, here's what Greg's terminal shows when I "open demo.app" then try to run a few things...

src_path   = '/Users/ian/src/Git/tests/fltk/test'
app_path   = '/Users/ian/src/Git/tests/fltk/build/bin'
fluid_path = '/Users/ian/src/Git/tests/fltk/build/bin'
data_path  = '/Users/ian/src/Git/tests/fltk/build/data'
Menu file  = '/Users/ian/src/Git/tests/fltk/build/data/demo.menu'
Cwd        = '/'

Command    = 'open '/Users/ian/src/Git/tests/fltk/build/bin/browser.app' --args 'browser.cxx''
Command    = 'open '/Users/ian/src/Git/tests/fltk/build/bin/editor.app' --args 'editor.cxx''
Command    = 'open '/Users/ian/src/Git/tests/fltk/build/bin/help_dialog.app' --args 'help_dialog.html''

erco77:
Great Ian, glad you confirmed with the new tty.

Looking at the code, it appears the intention is to use fl_chdir() in demo's main() to set it to the `data_path`, with the thinking paths for new child processes can be opened relative to that directory, and the child would inherit that cwd.

But obviously this doesn't work with `open(1)`.
What appears to be the (undocumented?) behavior, is a process opened with open(1) will have the cwd be (/), regardless of the cwd of the parent, and regardless of how 'open' is called, e.g.
```
open /some/long/path/foo.app
cd /some/long/path && open foo.app
```
For me, both techniques to open the demo.app show Cwd = '/', which means it'll be the same for the children as well.
So it appears fl_chdir() won't help us here on OSX.

Unless I'm missing something, looks to me like we always have to pass absolute paths to the child for pathnames to open, at least on OSX. And if the apps want to open files relative to a particular directory, those child apps will have to fl_chdir() themselves, so that their file `open(2)`/`fopen(3)` calls can work with relative paths.

Albrecht-S:
@erco See my newest commit with fixes.

I prepared a statement to restore the 'data_path' in the command arg. Please test and uncomment that one if necessary.

No time to read everything now (will do after I sent this), thanks to all.

The (restored) data_path should help, IMHO.

Albrecht-S:
> Looking at the code, it appears the intention is to use fl_chdir() in demo's main() to set it to the data_path, with the thinking paths for new child processes can be opened relative to that directory, and the child would inherit that cwd so that files (like browser.cxx) can be opened with a relative path.

That's **_absolutely correct_** [and so it is under Windows and Unix] and in fact I removed the `data_path` from the earlier code because I assumed that it wouldn't be necessary if the `cwd` was correct. In the old code it was set to `app_path` and the argument was **with** `data_path` (or another construct I don't exactly remember).

> But obviously this doesn't work with open(1).

Yeah, macOS is another beast.

Q: Why does the created process not inherit the `cwd` ?
A: Because it's executed with `open`. Doh!

So, after this I'm pretty sure that you need to edit demo.cxx and restore the prepared statement *with* `data_path` and everything should be fine.

Can you please try this (with the latest commit)? TIA.

erco77:
@Albrecht-S Yes, I pulled your latest, built it, and while Cwd shows the data_path, it still had the same errors, but then commented out the lines you mentioned and uncommented the data_path version of sprintf(), and now the browser app works.

I've committed the modified code, perhaps Ian can confirm.

Albrecht-S:
PS: does the output now show the expected `cwd`? Just for curiosity - as we know now it wouldn't help, but we're now hopefully using `data_path/filename` correctly.

(OK, I see you answered my question already).

Thank you and Ian very much for testing!

**I NEED A MAC !    ;-)  ;-)**

erco77:
Regarding getting a Mac, I'd recommend getting a mac mini. I have 5 or 6 of em.They're cheap and can run all operating systems, are quiet, and you can use your own keyboard/monitor. Just be sure to get an intel mac before they disappear.

Albrecht-S:
I should have mentioned: **I like the tty output, great work!**

However I think we should have a checkbox or something visible rather than the right click menu button. You know, people forget that it exists. ;-)

What about a checkbox `[ x ] debug` with a usual check (or light) button? Instead of 'debug' we could also write 'tty' but maybe 'debug' is more obvious?

Another thought: instead of extending the window, what about a separate, resizable, debug (tty) window?

Albrecht-S:
> Regarding getting a Mac, I'd recommend getting a mac mini.hat 

What about an old mac mini from ~2010 (or maybe earlier) which I have lying around but not used for 8 years or more? Any chance to resurrect it? I wanted to try it anyway, but before you ask: I can't tell which OS it was, but I'm pretty sure it was "a cat".

Albrecht-S:
@imaclmaca Some comments regarding your comments. I've been offline for a while.

> OK - my original checkout was messed up (I have no clue what I did...)

That's very likely my fault. I rebased the feature-demo branch and that hit you probably. When you fetch (or `git pull`) and you read something like "force pushed", then that's what happened. I had to do that after I rebased my branch.

The easiest way to recover is to `git reset --hard` your local branch (take care, `git reset --hard` would delete all modified files in your working copy!). So in this case, something like:
```
git status
<look for advice how to abort possible merges started by git pull and do this>
git fetch
git checkout feature-demo
git reset --hard REMOTE/feature-demo
```
Replace REMOTE with your configured remote repo and feature-demo with another branch if applicable.

Albrecht-S:
> (Note: if you "open" a file on MacOS from the terminal, you do not get stdio. One advantage of our old wrapper scripts was that you could still get stdio in tests...!)

- Where did these wrapper scripts come from?
- `fltk-config --post` in Makefile, maybe?
- What exactly would they do? (Example)
- Should we (try to) restore this functionality with CMake?

If yes, let me know how to do it (with normal system means) and I can try to add it to the CMake build.

Albrecht-S:
OK, I think I found it. Copied from `fltk-config`, slightly modified:
```
#!/bin/sh
# Make a shell script that runs the bundled executable
post="$1"
echo "#!/bin/sh" >"$post"
echo 'dir="`dirname \"$0\"`"' >>"$post"
echo 'exec "$dir/'"$post.app/Contents/MacOS/$post"'" "$@"' >>"$post"
chmod +x "$post"
```
Since the bundle would already have been created by CMake we'd only have to run this script from CMake to create the additional shell script. The shell script for bundle `utf8.app` would look like this:
```
#!/bin/sh
dir="`dirname \"$0\"`"
exec "$dir/bin/utf8.app/Contents/MacOS/bin/utf8" "$@"
```
Is this correct?

Albrecht-S:
What about a shell script that can run an arbitrary bundle like the old shell scripts did? Would this work and give us `stdout` and `stderr`?
```
#!/bin/sh
#
# Shell script that runs a bundled executable
#
# Usage:
#
#   run-bundle /path/to/bundle.app
#
dir="`dirname $1`"
app="`basename $1 .app`"
#
shift
exec "$dir/$app.app/Contents/MacOS/$app" "$@"
```
Works well under Linux with a fake bundle created with `fltk-config --post` after editing it (fltk-config) so it works under Linux.

erco77:
> `exec "$dir/bin/utf8.app/Contents/MacOS/bin/utf8" "$@"`

The /bin/ directory in MacOS seems foreign to me.

As I've always known it, the "correct" directory structure for the executable that Apple's Finder (and 'open') wants is:

`exec "$dir/bin/utf8.app/Contents/MacOS/utf8" "$@"`

..where the executable is in the MacOS directory and has the same name as the .app.
I would caution against changing that structure.

Albrecht-S:
> I would caution against changing that structure.

Well, that was not intentional. It was just a quick hack for demo purposes. Would appreciate if you could confirm that it works (if fixed with the correct structure). TIA.

erco77:
This works:
```
#!/bin/sh
dir="`dirname \"$0\"`"
exec "$dir/utf8.app/Contents/MacOS/utf8" "$@"
```
Had to remove both references to /bin/

erco77:
>     Where did these wrapper scripts come from?
>     fltk-config --post in Makefile, maybe?
>     What exactly would they do? (Example)
>     Should we (try to) restore this functionality with CMake?

Normally `fltk-config --post foo` is the last step in a Makefile which handles taking the single executable `foo` and turning it into a .app bundle by creating the .app directory structure, moves the executable into the .app structure, and then creates the wrapper script in its place.

IIRC it's safe to run 'fltk-config --post' on other platforms, as it's a NOP, but on OSX it should do all the above.

Offhand I'd suggest the cmake builds invoke `fltk-config --post xxx` to manage the whole .app directory structure creation and wrapper management, so that if we find we have to tweak the steps in fltk-config (we had to change some of the resource files once before for a change apple made), the cmake builds will inherit such tweaks, and we're using the same build steps users would use.

erco77:
> What about an old mac mini from ~2010 (or maybe earlier)

If you know the exact model, you can find out what the maximum OSX version is you can upgrade to, and decide based on that. If nothing else, you should be able to install linux and Windows on it too. I tend to put windows on the internal drive, and OSX on external USB/Thunderbolt drives so I can easily boot different OS's. Windows won't install on USB devices, but linux and OSX don't care.

> [new tty mods]  I think we should have a checkbox or something visible rather than the right click menu button.

Agreed, though wasn't sure what to use, as I didn't want to ugly up the nice clean interface that the demo program has.
What about a regular File menu? It could have just these menu items:

```
File/
  Preferences/
    [ ] Debug         <-- toggle
  Preferences/Scheme/
      [x]  normal     <-- radio buttons..
      [ ] plastic
      [ ] gtk+
      [ ] gleam

  Quit                <-- esc as shortcut

Help/
  About demo..
```

> Another thought: instead of extending the window, what about a separate, resizable, debug (tty) window?

Was trying to be consistent with all the other apps where it extends off the bottom (less window clutter), but I'm open. Would be easier implementation wise I suppose..

Albrecht-S:
> This works:
> ...
> Had to remove both references to /bin/

Thanks for verifying. My mistake was that I executed `./fltk-config --post bin/utf8` in the CMake **build folder** - which created the fake bundle and the wrong script with the additional `/bin`. Unfortunately I posted that wrong version. Obviously `fltk-config --post` must be executed exactly in the directory where the executable/bundle is.

> IIRC it's safe to run 'fltk-config --post' on other platforms, as it's a NOP ...

Yes, it queries `uname` and works only on `Darwin`. I edited it to allow `Linux` as well, then it worked fine.

BTW, if you create such a bundle under Linux you _can execute the created script_ and it works. It's useless but there's nothing special with it.

> Offhand I'd suggest the cmake builds invoke fltk-config --post xxx ... so that if we find we have to tweak the steps in fltk-config (...), the cmake builds will inherit such tweaks, ...

No, I don't think this is the way to go. `fltk-config` can't do anything _"special"_ with the bundles, whereas the `CMake` build can create bundles and add special resources (for instance icons, but also other files) which an app may need. You see can how it is done in macro `CREATE_EXAMPLE` and at the end of `test/CMakeLists.txt`. If we ran `fltk-config --post` after CMake created the bundle we'd destroy it. In the future it would likely be the opposite: we can see how CMake generates the bundles and tweak `fltk-config`.

Albrecht-S:
Regarding a menu or check button: Generally I like the menu idea (at the top as usual) but I'd prefer the check button for several reasons:
- a menu would need a full menu height (menu bar) and shift everything else down [1]
- we have Fl_Choice widgets for the scheme selection elsewhere, we should keep it for consistency
- in a future FLTK update [2] we'll be able to populate the Fl_Choice widget from all existing [3] schemes "automagically"

OTOH, the `Help/About` menu could be used to tell devs and users something about the program. I don't know if this is worth it though.

Notes:
1. I'm aware that we could get the lost space back if we also removed the `Exit` button
2. I've been working on an update of the FLTK schemes and the part to create the Fl_Choice widget works already. It's great to have this so we don't need to edit several source files to update the Fl_Choice widgets and/or menus. Unfortunately this ("`Fl_Scheme`") update is a lot of work and I'm currently busy with other things as you know.
3. My plan is to enable user programs to add schemes at runtime. I'm confident that this will work. Somehow.

Albrecht-S:
@erco @imaclmaca @ManoloFLTK 

That all said, I'd appreciate if Ian (and Manolo if available) could confirm that the current status is good to go. We should postpone the menu question as this can be done in a separate step.

In this case I'd squash and merge like this:
- squash unnecessary small updates and fixes (we don't need all the test steps)
- keep essential modifications (e.g. Greg's addition of the `tty` stuff) as separate commits
- finally merge with `master` to keep the (simplified) history

Let's try to "vote" by using smileys on this comment (click at the top smiley icon or below): 
- +1 (:+1:) Good to go
- -1 (:-1:) Not yet (add another comment)
- 0 (abstain): add another comment

Your votes here (below) please:

imaclmaca:
Is there a "grumpy old git" emoticon for use by people who don't like smileys?

Anyway, I'm fine with the new mechansim, I think. (Though I have a build running just now, and have not been able to bring myself to click on the icon yet...)

For the record, attached is the code I use to deteremine the actual "home directory" of the executable at runtime - it's old, but AFAIK it still works...

[get_app_path.zip](https://github.com/fltk/fltk/files/4952621/get_app_path.zip)

imaclmaca:
And, off-topic, my "newest" Mac is circa mid-2010, and still goes well. The newest OS it runs is 10.13, but that's fine as far as I can tell!
(The disk was swapped for a firewire SSD a while back, but otherwise it's pretty much "as stock" and still pretty fast.)

Albrecht-S:
FYI:
1. I added another commit (880d67e) which improves comments and removes dead code.
2. I squashed commits as announced above and the result can be viewed in the new branch [Albrecht-S/feature-demo-v2](https://github.com/Albrecht-S/fltk/tree/feature-demo-v2) just in case someone wants to see the (likely) final result.

See the last three commits in that branch.

Albrecht-S:
@imaclmaca 
> Is there a "grumpy old git" emoticon for use by people who don't like smileys?

Hey man, it's just one click, and there is a **visible result** of our votes. ;-)
Better than counting and summarizing, isn't it?
I'd like to standardize this so we can vote here (on GitHub) "easily" in the future. You'll see who voted when you hover the mouse over the icon and so you can tell how many official (dev) votes we have.

> my "newest" Mac is circa mid-2010, and still goes well. The newest OS it runs is 10.13, but that's fine as far as I can tell!

Thanks, that can help me, I'll try to get mine up.

Albrecht-S:
@erco77 I'd like to merge the current status with changes made in branch [Albrecht-S/feature-demo-v2](https://github.com/Albrecht-S/fltk/tree/feature-demo-v2) which is technically the same as branch feature-demo, but reduced to three commits.

This is what I'd like to merge if you agree (question is about the commits squashed by me).

erco77:
I see you kept the tty mod without squashing, which is good.
All else is fine w/me.

Albrecht-S:
@erco77 
> I see you kept the tty mod without squashing, which is good.

:+1: 

> I take it the OSX wrapper scripts will eventually be handled as part this issue, right?

No, not directly. I'll merge this PR which will set its status to "merged" which is the same as "closed with success". The title says it all "... in test/demo.cxx". That's what I wanted to achieve.

I heard of the "missing bundle shell scripts" here for the first time - as you know I'm not a Mac user. I'll look into it though, independently of this PR. Feel free to open an issue so we don't forget. This PR and its discussion is alread way too long.

Thanks for your and @imaclmaca 's support. Going to merge the branch now...

erco77:
> I heard of the "missing bundle shell scripts" here for the first time - as you know I'm not a Mac user.
> I'll look into it though, independently of this PR. Feel free to open an issue so we don't forget. 

Done: issue #115

PR #114, opened 2020-07-20
Fix issue with Fl_Tiled_Image: does not correctly draw to widget sub-region

fire-eggs:
Discovered while trying to use Fl_Tiled_Image into a sub-region of my window that it does not draw correctly.

Fl_Tiled_Image/draw() is written as if the W / H parameters are "rect right" / "rect bottom" instead of width/height.

To demonstrate, I've included a modified version of unittest_images.cxx. If the unittests project is built with TILE_IMAGE defined, the "drawing images" unit test will use a pixmap and  Fl_Tiled_Image to draw the "checker board" background behind the images.

Without the fix, you will not see the checker board. With the fix, the checker board is shown.
The other test program, tiled_image, is unaffected.

### DIFF
--- a/src/Fl_Tiled_Image.cxx
+++ b/src/Fl_Tiled_Image.cxx
@@ -182,12 +182,12 @@ Fl_Tiled_Image::draw(int X,     // I - Starting X position
   if (cx > 0) iw -= cx;         // crop image
   if (cy > 0) ih -= cy;
 
-  for (int yy = Y; yy < H; yy += ih) {
+  for (int yy = Y; yy < Y+H; yy += ih) {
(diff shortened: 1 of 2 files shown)


Albrecht-S:
Thanks, I'll look into it when time permits.

Albrecht-S:
@fire-eggs : I modified your patch slightly (whitespace and comments) and committed it in cdda89455c6767b5f43628b1c9e45b2dc8f0d971 w/o the modification of `test/unittest_images.cxx`. Thanks for the test code nevertheless.

Closing this PR now because it's fixed.

PR #118, opened 2020-07-25
add fl_filename_match comments as per STR3186

engelsman:
add Notes: section to fl_filename_match documentation
explaining case-sensitivity [or not] and use of \x

see also https://www.fltk.org/str.php?L3186

### DIFF
--- a/src/filename_match.cxx
+++ b/src/filename_match.cxx
@@ -1,7 +1,7 @@
 //
 // Pattern matching routines for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2010 by Bill Spitzak and others.
+// Copyright 1998-2020 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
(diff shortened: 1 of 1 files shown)


engelsman:
I started this as another "low-hanging fruit" in the Developer Roadmap / STR system

The current change is the simplest possible, with just a few additional comments in the documentation. Note that some of the things that are matched are a bit counter-intuitive. For what it's worth, I shall attach a crude test program

I had though the next level, if it were needed, would be to modify the current fl_filename_match(s, p) to call a new fl_filename_match(s, p, int case_sensitivity=-1) where case_sensitivity of -1 (or -ve) would give the current mixed behaviour, 0 would be truly case-insensitive, and 1 (or +ve) would be truly case-sensitive. In reality I thought that this could be just a few lines of additional code.

However, I just realized that the current fl_filename_match(s, p) only compares on a char/byte basis and not on UCS codepoints or UTF-8 sequences, which might be needed if this were used with a UTF-8 aware fl_open() for example. Now I'm not sure how to proceed.

engelsman:
[test_fl_filename_match.zip](https://github.com/fltk/fltk/files/4975234/test_fl_filename_match.zip)

Albrecht-S:
@engelsman Thanks for taking this STR. Just a note at
> modify the current fl_filename_match(s, p) to call a new fl_filename_match(s, p, int case_sensitivity=-1) where case_sensitivity of -1 (or -ve) would give the current mixed behaviour, 0 would be truly case-insensitive, and 1 (or +ve) would be truly case-sensitive.

Looks useful, but could probably be done simpler by just adding `int case_sensitivity=-1` as an optional argument. No extra call needed IMHO.

Regarding case sensitivity: good points, I'd guess that UTF-8 awareness lacks in many other cases as well, and I don't know how difficult it would be to write case sensitive and case insensitive comparisons WRT international (including eastern) languages. Is there something like a "case" in Japanese and Chinese, for instance? I don't know.

I know this (at least the part regarding case sensitivity) is not very helpful but all I can contribute for now.

Albrecht-S:
Thanks to @engelsman for the PR; I decided to merge it now because it's only doc changes. Adding true case sensitivity would require a new issue or PR.

I'm also going to close STR 3186.

PR #133, opened 2020-09-05
Make building fluid optional (ON by default)

MoAlyousef:
Hello
Sometimes it doesn't make much sense to build fluid by default, especially when only the library is needed, such as with package managers etc.
I've modified the cmake files to make building fluid optional, and set it to be ON by default as to not break older builds. The build can be turned off using the cmake definition "-DOPTION_BUILD_FLUID=OFF " when invoking cmake from the command line.

### DIFF
--- a/CMake/export.cmake
+++ b/CMake/export.cmake
@@ -33,9 +33,11 @@ if (CMAKE_CROSSCOMPILING)
     PROPERTIES IMPORTED_LOCATION ${FLUID_PATH}
   )
 else ()
-  add_subdirectory(fluid)
-  set (FLTK_FLUID_EXECUTABLE fluid)
-  set (FLUID fluid) # export
+  if(OPTION_BUILD_FLUID)
+    add_subdirectory(fluid)
+    set (FLTK_FLUID_EXECUTABLE fluid)
+    set (FLUID fluid) # export
+  endif(OPTION_BUILD_FLUID)
 endif (CMAKE_CROSSCOMPILING)
 
 add_subdirectory(src)
--- a/CMake/export.cmake
+++ b/CMake/export.cmake
@@ -33,11 +33,11 @@ if (CMAKE_CROSSCOMPILING)
     PROPERTIES IMPORTED_LOCATION ${FLUID_PATH}
   )
 else ()
-  if(OPTION_BUILD_FLUID)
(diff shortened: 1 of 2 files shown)


Albrecht-S:
Thanks for your PR. This is possible, but there are two issues:
- (minor) the CMake option should be called `FLTK_BUILD_FLUID`, not `OPTION_BUILD_FLUID` to be consistent with other "`BUILD`" options
- there are dependencies between options: we can't build the test and demo programs if we don't build `fluid` - unless we're cross-compiling which uses the build host's `fluid`. This should be checked and the build should fail early in such a case.

Another option would be to disable `FLTK_BUILD_TEST` (and maybe also `FLTK_BUILD_EXAMPLES`) automatically with a warning if `FLTK_BUILD_FLUID` is `OFF`. Or something like this, it's IMHO not clear what would be the best option.

MoAlyousef:
I have renamed the option to FLTK_BUILD_FLUID.
As for the other options, I see FLTK_BUILD_EXAMPLES is Off by default, while FLTK_BUILD_TEST is On by default. Would it be reasonable to make both Off by default?
Another option is to use the CMakeDependentOption macro:
https://cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html

Xunie:
> it's IMHO not clear what would be the best option.

If the user specifies an incorrect build with non-compatible options? That's an error and should **not** be treated as warning. (Fail-fast and all, but that's just my two-cents.)

Xunie:
I don't think this should be closed. This is a valid enhancement, right?

MoAlyousef:
I should've discussed this prior to opening the pull request. I actually need my current fork for some other changes.

Albrecht-S:
Don't worry, I added this to my personal agenda anyway - i.e. I'll check if it can be added and how. But I'm still not convinced that this is a useful addition because building fluid alone is not a big deal and it's required to build some of the test programs. I'd like to reduce the number of build options rather than adding new ones.

> I actually need my current fork for some other changes.

@MoAlyousef I suggest to **always use a branch** (never use '_master_') if you want to create a pull request or do/test other things. This way you can have multiple branches (and PR's) active at the same time w/o conflicts and you can easily update your fork by pulling the _upstream/master_ branch (and rebase your own branches on current _master_).

PR #151, opened 2020-11-01
read_win_rectangle_unscaled: Fix a memory leak and …

fire-eggs:
…awkward code from the drivers conversion.

In FLTK 1.3, the `p` array could be passed in, thus the usage of `oldp` construct. In FLTK 1.4, the code _always_ allocates the array, allowing the suggested changes.

The memory leak occurs in the very unusual situation that there is "nothing to copy" and is fixed by allocating after that check.

[Discovered via cppcheck 2.2]

### DIFF
--- a/src/drivers/WinAPI/Fl_WinAPI_Screen_Driver.cxx
+++ b/src/drivers/WinAPI/Fl_WinAPI_Screen_Driver.cxx
@@ -504,13 +504,7 @@ Fl_WinAPI_Screen_Driver::read_win_rectangle(
 Fl_RGB_Image *Fl_WinAPI_Screen_Driver::read_win_rectangle_unscaled(int X, int Y, int w, int h, Fl_Window *win)
 {
   int   d = 3;                  // Depth of image
-  int alpha = 0; uchar *p = NULL;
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Good catch, thanks! I think the modified code is correct.

However, there's even more "awkward code" that could be eliminated, particularly variable ` int alpha = 0;` could be removed entirely (replaced with literal `0` where needed) since it's constant throughout the code. Then it's obvious that the condition at line 581 can never be true and the entire `if (...)` statement can be removed:
```
    if (alpha)
        *tg++ = alpha;  // alpha
```

@fire-eggs Kevin, would you like to update your PR accordingly? Otherwise we could do this afterwards as well.

That all said, I'd like to have this reviewed by @ManoloFLTK because he knows best what's this all about. Maybe there are similar places (other platforms) with the same issue.

Albrecht-S:
FTR: variable `int d = 3;` is also constant and could likely be eliminated as well...

ManoloFLTK:
Thanks for spotting this potential leak.
That should now be fixed with commit 98a4e49.

PR #165, opened 2020-11-24
add new resize example to test and dox

engelsman:
add example(s) showing Albrecht's innovative overlap resizable
technique to the "How does resizing work?" documenation and tests

See "resizable question" original discussion thread under:
https://www.fltk.org/newsgroups.php?gfltk.general+v:39635


engelsman:
New image files were produced using Preview on MacOS,
reducing from 144 to 72 dpi, then halving width/height,
but file sizes could probably be reduced further

erco77:
FYI the technique I use is to use GIMP:

1. Load the PNG
2. Choose Image -> Mode ->Indexed, and choose a 'maximum number of colors' value that is the lowest 2's exponent. For pure b+w with some gray for antialiasing, I suggest a value of 8. For just a few colors with no gradients, 16 or 24. Otherwise I don't bother using indexed colors, as the spacing savings becomes negligible.
3. Save the PNG, and it should be really small.

engelsman:
Followed erco's instructions -- isn't the gimp complicated? :-) -- to go from:
```
-rw-r--r--@ 1 duncan  staff  16984 Nov 23 20:44 Desktop/resize-example5a.png
-rw-r--r--@ 1 duncan  staff  23267 Nov 23 20:49 Desktop/resize-example5b.png
-rw-r--r--@ 1 duncan  staff  16921 Nov 23 20:53 Desktop/resize-example5c.png
```
to
```
-rw-r--r--  1 duncan  staff  2645 Nov 24 08:13 resize-example5a.png
-rw-r--r--@ 1 duncan  staff  3241 Nov 24 08:14 resize-example5b.png
-rw-r--r--@ 1 duncan  staff  2653 Nov 24 08:15 resize-example5c.png
```
I can squash the commits later to zap the larger images

Albrecht-S:
@engelsman Hi Duncan, I apologize for the long delay, but now it's done. Thank you very much for your contribution.

I squashed your commits (hence the force-push) and added another commit with minor fixes. Removing the "Figure n.n: " text was necessary since we removed these from all HTML docs.

engelsman:
@Albrecht-S : It's funny, but I was just thinking about this, because (a) I had sort of promised to consolidate the example code but never had time, and (b) after the discussions about Fl_Flow I was musing that [if I had the time to contribute] maybe it would be an idea to kick off a chapter before the resizing called "Widget coordinates and layout" or something like that.

The first part would discuss the pros and cons of window-relative positioning, and the second part would provide some discussion about raw Fl_Groups, then Fl_Pack and Fl_Tile, and of course a section of Fl_Flow, with maybe some brief notes about Fl_Scroll, Fl_Tab and Fl_Wizard. I thought the other widgets derived from Fl_Group are basically just fixed containers for related components, so probably don't need to be mentioned. 

But to be honest, I can't promise anything because I simply don't have the time, and because I don't have any inspiration about what sort of examples to create, unless the the chapter can re-use the images from the class docs and just talk around them. So don't hold your breath and wait for me.

Albrecht-S:
@engelsman: Thanks, Duncan, it's all fine as it is. If you have some good ideas, feel free to open another PR or discuss your ideas in fltk.coredev. I have some ideas for changes of the docs as well, hence I'd want to avoid duplicate work or even merge conflicts.

FWIW: I'm currently working on improvements of the "Print/Copy front window" button for Windows and Linux (on macOS it's in the top menu) which can - once I'm ready - be used in all FLTK programs. This should enable us to create screenshots of test programs easily. The latest enhancement is that the extra window with the buttons appears only if an environment variable is defined but we can also define a global shortcut to `show()` the screenshot (Print/Copy) window or something else.

In my tests I'm working with examples/clipboard.cxx to capture the screenshots. I'm thinking about extending it (or to write another program) to save the screenshots as PNG images but I'm also thinking of more ootions. This looks very promising and is IMHO something FLTK users would like too.

A nice side effect is zooming screenshots. You can use ctrl/+ to zoom the `clipboard` program and as soon as you copy an image to the clipboard it appears scaled in the `clipboard` program.

engelsman:
I've added a section to the user manual describing window-relative coordinates and the basic layout
manager widgets, which will leave space to add sections on Fl_Grid and Fl_Flow if necessary.
I've also added a test/wizard.cxx example in order to have a screenshot for the manual.

P.S. Still trying to get my head round the implications of 2FA, TOTP, KeePassXC, etc. hence this patch.
I'll try to submit as a proper pull request this weekend, so that the changes can be discussed.

[0001-add-coordinates-and-layout-section-to-user-manual.patch.txt](https://github.com/fltk/fltk/files/7607899/0001-add-coordinates-and-layout-section-to-user-manual.patch.txt)

PR #169, opened 2020-12-07
[FLTK 1.1] Fix min not defined compiler error.

ComputerNerd:
Instead of a macro use std::min from algorithm. It's in C++98 so most likely legacy compilers won't be negatively impacted by this change.

### DIFF
--- a/src/fl_draw.cxx
+++ b/src/fl_draw.cxx
@@ -32,7 +32,7 @@
 // Expands all unprintable characters to ^X or \nnn notation
 // Aligns them against the inside of the box.
 
-#define min(a,b) ((a)<(b)?(a):(b))
+#include <algorithm>
 #include <FL/fl_draw.H>
 #include <FL/Fl_Image.H>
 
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Hi @ComputerNerd, thank you for your patch(es). Note that FLTK 1.1 is no longer maintained and won't get another release, but nevertheless I'm looking into your patches which **may** be applied to Git so users can pull from Git if they like.

However, this one is not acceptable because we don't use anything of `std::` in FLTK (at least for now and particularly not in 1.1). Please see our [CMP](https://www.fltk.org/cmp.php) for more info on that.

I'm curious: why do you want to replace the macro? The title says "Fix min not defined compiler error" - how can that be if there's a macro `min`? I've seen that you're building FLTK 1.1 with MinGW for Win95, but which compiler version issues this error?

ComputerNerd:
I was also confused when I first got the error. I saw the min macro at the top of the file and in the fl_draw.cxx I didn't see anything that would cause it. It turns out that the Debian maintainers still seem to support FLTK 1.1 and have been writing patches for it. They moved the min macro down lower to resolve the error.

Here you can see their change and explanation:
https://salsa.debian.org/ucko/fltk1.1/-/blob/master/src/fl_draw.cxx
Quote from the page
```src/fl_draw.cxx: Move the min macro's definition down, to ensure no headers have the opportunity to undefine it.```

What I'll do is close this pull request and submit another one using the fix used by the Debian maintainers.

Also thank you for linking to the CMP document and for reviewing my patches. I will make sure any future changes comply with this.

For the compiler I'm using GCC 10.2.0. It seems like this issue started happening with GCC 6 according to the Debian maintainers.

PR #172, opened 2020-12-08
Fix min not defined compiler error (without using std::min) by using the same method the Debian maintainers did by moving it down

ComputerNerd:
This accomplishes the same goal as my previous pull request except it doesn't use std:: which is not acceptable per the FLTK CMP as discussed here: https://github.com/fltk/fltk/pull/169

### DIFF
--- a/src/fl_draw.cxx
+++ b/src/fl_draw.cxx
@@ -39,6 +39,7 @@
 #include <ctype.h>
 #include <math.h>
 
+#undef min
 #define min(a,b) ((a)<(b)?(a):(b))
 #define MAXBUF 1024
 
--- a/src/fl_draw.cxx
+++ b/src/fl_draw.cxx
@@ -32,7 +32,6 @@
 // Expands all unprintable characters to ^X or \nnn notation
 // Aligns them against the inside of the box.
 
-#define min(a,b) ((a)<(b)?(a):(b))
 #include <FL/fl_draw.H>
 #include <FL/Fl_Image.H>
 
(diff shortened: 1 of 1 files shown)


erco77:
Should this maybe be prefaced by an '#undef min' or surrounded by an '#ifndef min' check to avoid "redefined" if already defined? I know I've seen OS headers define min() and max().. IRIX? GCC?

ComputerNerd:
Yes, that makes sense to me. I went with the #undef approach so that way we are using our own macro instead of using a macro from a system header. Although it's very likely that any min macro from a system header would function the same I don't see any reason to rely on that when we can be 100% sure we are using our own macro.

Windows.h also does that. I typically compile my Windows programs with -DNOMINMAX as a part of the CFLAGS.

Albrecht-S (approved):
This looks like a good move (moving the #define down + adding the #undef).

Albrecht-S:
Thanks for your updated PR. This seems to be acceptable and useful.

Note to myself (or @erco77 if you like): check FLTK 1.4 for similar mods.

erco77:
Just noting this PR is for the no longer supported branch fltk 1.1.x.
Albrecht, should we just merge and have this just remain in git as a commit beyond the final 1.1.x release?

Albrecht-S:
@erco77 Yes, that was my plan. There are some more (2-3) other PR's from @ComputerNerd which I wanted to revisit, evaluate and possibly merge all of these together but I got distracted by other work in the meantime. If you feel inclined, go ahead and merge this one, but I to "rebase and merge" rather than doing a real "merge". Well, now I saw that this PR consists of two commits which should likely be squashed ...

erco77:
OK, I'll leave it to you, sounds good to squash this one.

PR #174, opened 2020-12-14
[FLTK 1.4] Skip the call to MonitorFromRect when it is not needed.

ComputerNerd:
This pull request makes two changes. The first is to only call the MonitorFromRect function when it's needed.

The second is to use GetProcAddress instead of linking against it directly so that way older versions of Windows can still run FLTK 1.4. This function was the only thing preventing it from running on Windows 95 and Windows NT 4.0.

### DIFF
--- a/src/Fl_win32.cxx
+++ b/src/Fl_win32.cxx
@@ -565,13 +565,22 @@ void Fl_WinAPI_Screen_Driver::open_display_platform() {
 
 void Fl_WinAPI_Screen_Driver::desktop_scale_factor() {
   typedef HRESULT(WINAPI * GetDpiForMonitor_type)(HMONITOR, int, UINT *, UINT *);
+  typedef HMONITOR(WINAPI * MonitorFromRect_type)(LPCRECT, DWORD);
   GetDpiForMonitor_type fl_GetDpiForMonitor = NULL;
-  if (is_dpi_aware)
+  MonitorFromRect_type fl_MonitorFromRect = NULL;
+  if (is_dpi_aware) {
     fl_GetDpiForMonitor = (GetDpiForMonitor_type)GetProcAddress(LoadLibrary("Shcore.DLL"), "GetDpiForMonitor");
+    if (fl_GetDpiForMonitor != NULL)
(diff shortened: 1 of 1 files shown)


src/Fl_win32.cxx
```diff
@@ -565,13 +565,22 @@ void Fl_WinAPI_Screen_Driver::open_display_platform() {
...
     fl_GetDpiForMonitor = (GetDpiForMonitor_type)GetProcAddress(LoadLibrary("Shcore.DLL"), "GetDpiForMonitor");
+    if (fl_GetDpiForMonitor != NULL)
+      fl_MonitorFromRect = (MonitorFromRect_type)GetProcAddress(LoadLibrary("User32.DLL"), "MonitorFromRect");
+  }
   for (int ns = 0; ns < screen_count(); ns++) {
-    HMONITOR hm = MonitorFromRect(&screens[ns], MONITOR_DEFAULTTONEAREST);
+    HRESULT r = !S_OK;
     UINT dpiX, dpiY;
-    HRESULT r =  fl_GetDpiForMonitor ? fl_GetDpiForMonitor(hm, 0, &dpiX, &dpiY) : !S_OK;
+    // The result of MonitorFromRect is only needed when is_dpi_aware is true and when GetProcAddress successed getting the address for GetDpiForMonitor.
```
Albrecht-S, line 579: s/successed/succeeded/ ?

Albrecht-S (changes requested):
Code looks good except for some inconsistencies in handling whether "something" is NULL or not. I'd prefer using, for instance `if (func)` rather than `if (func != NULL)` for a more consistent appearance.

I didn't test anything though and I'd like @ManoloFLTK to review this.

Albrecht-S:
@ManoloFLTK I'm wondering if `desktop_scale_factor()` can be called more than once or only at initialization time.

If it could be called more than once, shouldn't the function pointer(s) be static so we don't need to call `GetProcAddress()` more than once per function name?

ManoloFLTK:
Thanks for this PR. It's good to recover compatibility with Windows 95.
The suggested changes have been committed to the git repository.

Function Fl_WinAPI_Screen_Driver::desktop_scale_factor() runs only once,
when the display is opened.

Issue #190, opened 2021-02-15
Android: Threading issues


MoAlyousef:
On Android, FLTK's locking mechanisms don't work correctly (deadlock crashes), neither for enabling the global lock (via Fl::lock) or thru the Fl::awake(void *) messages (pipes). The Fl_lock.cxx file has the locking functions in an Android ifdef and remain to be implemented. I was able to get things working by subclassing the Fl_Android_System_Driver from the Fl_Posix_System_Driver instead of the Fl_System_Driver. I had to copy some implementation of the other posix systems (X11 namely for locking) since Android also uses pthreads. 
```rust
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4e8d8c759..c7aa9d93c 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -300,6 +300,7 @@ elseif (ANDROID)
 
   set (DRIVER_FILES
     drivers/Android/Fl_Android_Application.cxx
+    drivers/Posix/Fl_Posix_System_Driver.cxx
     drivers/Android/Fl_Android_System_Driver.cxx
     drivers/Android/Fl_Android_Screen_Driver.cxx
     drivers/Android/Fl_Android_Screen_Keyboard.cxx
@@ -311,6 +312,7 @@ elseif (ANDROID)
   )
   set (DRIVER_HEADER_FILES
     drivers/Android/Fl_Android_Application.H
+    drivers/Posix/Fl_Posix_System_Driver.H
     drivers/Android/Fl_Android_System_Driver.H
     drivers/Android/Fl_Android_Screen_Driver.H
     drivers/Android/Fl_Android_Window_Driver.H
diff --git a/src/drivers/Android/Fl_Android_System_Driver.H b/src/drivers/Android/Fl_Android_System_Driver.H
index 5e6dafb1e..a985d9324 100644
--- a/src/drivers/Android/Fl_Android_System_Driver.H
+++ b/src/drivers/Android/Fl_Android_System_Driver.H
@@ -25,6 +25,7 @@
 #include <string.h>
 #include <FL/fl_string.h>
 #include "../../Fl_System_Driver.H"
+#include "../Posix/Fl_Posix_System_Driver.H"
 #include <stdarg.h>
 
 /*
@@ -38,7 +39,7 @@
  - multithreading
  */
 
-class Fl_Android_System_Driver : public Fl_System_Driver
+class Fl_Android_System_Driver : public Fl_Posix_System_Driver
[...]

PR #194, opened 2021-02-22
Android threads

MoAlyousef:
Addresses issue #190.


MatthiasWM:
The Android Driver has been removed as it was unfinished work without progress. A new implementation is planned that will be based on SDL2 and will cover iOS, Android, and Linux frame buffer use. And yes, the proposed use of the Posix backend is the correct way to extend this port.

MatthiasWM:
Won't fix in the current FLTK version.

PR #199, opened 2021-03-10
Fix stack corruption while loading gif image

fire-eggs:
I have a GIF image which repeatedly causes a "Stack corruption around OutCode" error within Visual Studio.  

After examining several GIF decode implementations, including one also derived from gif2ras (see [https://www.unf.edu/public/cap6400/ychua/xv-2.21/xvgif.c] ) I've determined that `OutCode` needs to be at least 4096, not the current 1024.

After changing the array size to 4097, the stack corruption no longer happens for the GIF image in question.

The GIF image which causes the error is attached in this ZIP [yes, it's SFW].

[animu.ru-clannad-(2560x1920)-wallpaper-001.gif.zip](https://github.com/fltk/fltk/files/6117909/animu.ru-clannad-.2560x1920.-wallpaper-001.gif.zip)


fire-eggs:
fouled up w/ codeql-analysis. trying again.

PR #200, opened 2021-03-10
Fix stack corruption when loading GIF

fire-eggs:
I have a GIF image which repeatedly causes a "Stack corruption around OutCode" error in Fl_GIF_Image.cxx while running in Visual Studio.

After examining several GIF decode implementations, including one also derived from gif2ras
[see(https://www.unf.edu/public/cap6400/ychua/xv-2.21/xvgif.c)] I've determined that OutCode needs to be at least 4096, not the current 1024.

After changing the array size to 4097, the stack corruption no longer happens for the GIF image in question.

The GIF image which causes the error is attached in this ZIP [yes, it's SFW].

[animu.ru-clannad-(2560x1920)-wallpaper-001.gif.zip](https://github.com/fltk/fltk/files/6118016/animu.ru-clannad-.2560x1920.-wallpaper-001.gif.zip)

### DIFF
--- a/src/Fl_GIF_Image.cxx
+++ b/src/Fl_GIF_Image.cxx
@@ -337,7 +337,7 @@ void Fl_GIF_Image::load_gif_(Fl_Image_Reader &rdr)
 
     if (CurCode == EOFCode) break;
 
-    uchar OutCode[1025]; // temporary array for reversing codes
+    uchar OutCode[4097]; // temporary array for reversing codes
     uchar *tp = OutCode;
     int i;
     if (CurCode < FreeCode) i = CurCode;


ManoloFLTK:
Many thanks for this find. 

I have seen that the highest index in the OutCode array necessary
when reading your .gif file is 1178, which is indeed higher than the 1024 assigned
to that array in the present code. So we definitely need to change that value.

My question is: do you have a basis for setting the size of OutCode to 4096?

fire-eggs:
According to the GIF spec:
```
4.  The output codes are of variable length, starting  at  <code size>+1 bits  per 
code, up to 12 bits per code.  This defines a maximum code value of 4095 (hex FFF).
```
The various GIF decoders I've examined either use a fixed size table, where the table is 4096 bytes, or dynamically re-alloc the table when necessary.

The FLTK code is "based on gif2ras". Looking at [another "gif2ras" based decoder](https://www.unf.edu/public/cap6400/ychua/xv-2.21/xvgif.c) the code table is set to 4097 (see line 100).

Albrecht-S:
@fire-eggs Thanks for the research, this looks plausible.
@ManoloFLTK Would you like to merge this PR? If yes I suggest to select "Rebase and merge" in the "Merge button" above - this will preserve the original author and commit.

PR #203, opened 2021-03-19
Multithread fix for PNG loading

fire-eggs:
On Windows I have been successfully using FLTK and OpenMP to load images in parallel.  To get it to work, I had to make this modification to the PNG load code: a STATIC variable resulted in frequent thread collision issues.

By moving to a class member I solved the thread collision issues while preserving the requirement that it not be a stack variable (as per the comment in the original code).


Albrecht-S:
@fire-eggs Kevin, thanks for this proposal. However - as it is often the case - this is more complicated than it seems at a first glance. Certainly the code as it is now is not thread-safe, that's absolutely correct. Moving `fp` to a class member can solve _this particular issue_, but there are drawbacks that let me hesitate to merge this PR:
1. The proposed fix adds a pointer to the `Fl_PNG_Image` object (8 bytes on a 64-bit system) that is useless after the image has been loaded from the file and even always (useless) if the image is loaded from memory. This is a minor issue, but 8 bytes are 8 bytes, and FLTK is designed to be _Fast and **Light_**.
2. This does not solve the issue that `setjmp` used in the png library is known to be not thread-safe on some platforms [1], which means that this fix would still not be thread-safe.

Maybe we can find a better solution? (To be continued...)

[1] From png/INSTALL with line numbers for reference:
```
428  Libpng uses setjmp()/longjmp() for error handling.  Unfortunately setjmp()
429  is known to be not thread-safe on some platforms and we don't know of
430  any platform where it is guaranteed to be thread-safe.  Therefore, if
431  your application is going to be using multiple threads, you should
432  configure libpng with PNG_NO_SETJMP ...
```

fire-eggs:
My thanks for the feedback! I'd not thought too hard about the larger picture. I'll noodle on this and additional ideas will be appreciated!

Albrecht-S:
1. Further analysis shows that setjmp/longjmp on Linux is (at least under certain circumstances) thread-safe as an addition to the info from libpng above. See man setjmp(3), man attributes(7).
2. To find the issue I used `g++ -W` or (to suppress other warnings) `g++ -Wclobbered`. You can still see it if you remove the `static` attribute from `fp` (see also (3) below).
3. The Linux manpage says that local variables can be optimized out and are thus **undefined** under certain circumstances (see man setjmp). Hence you need to compile with optimization (aka Release) to trigger the warning.
4. The manpage also says that it *should* not be an issue if you make the variable `volatile` (instead of static). I tried that (which needs some casts) but the warning persisted. Maybe a compiler issue? False warning?
5. I could not (yet) replicate the warning with clang (tried it on macOS only). clang doesn't understand `-wclobbered`, I need to dig deeper into it.

Another idea would be to give fp as a function argument to the problematic function which should preclude the "clobbered" / undefined issue. Although this is generally possible it would require to include <stdio.h> in the public header file which I would try to avoid.

Will continue testing later...

fire-eggs:
>Another idea would be to give fp as a function argument to the problematic function which should preclude the "clobbered" / undefined issue. Although this is generally possible it would require to include <stdio.h> in the public header file which I would try to avoid.

I really like this one. To avoid including stdio, I'd declare it as a void* in the header, Fl_PNG_Image.cxx "knowing" to cast it appropriately. Is that too "hacky"?

Albrecht-S:
> To avoid including stdio, I'd declare it as a void* in the header, the functions in Fl_PNG_Image.cxx "knowing" to cast it appropriately. Is that too "hacky"?

That's something I definitely thought of. Is it too hacky? In some cases we're doing this because we need platform specific data types but we can't "know" (or don't want to know) the platform/type in the header. This here is different: do we want to do something to silence a compiler warning for thread-safety which is not even guaranteed to achieve this goal (at least on some platforms)? What about all the other local variables in this function? We'd also need to document it clearly (as you can see the attempt in the current comment) to make clear **why** this is done. Hard to say if it's "too hacky".

Anyway, I would accept it if it's clearly documented (I could do that myself) and if we can't come up with a better solution.

That said, FLTK and all its functions and structures is/are not thread-safe by design unless you call `Fl::lock()` and `Fl::unlock()` - which admittedly doesn't solve your problem loading images in a multi-threaded application, of course. I can see that...

fire-eggs:
I thought this would be an "easy" thing to provide a fix for - seems more complicated than it's worth for a single person's issue. 

>That said, FLTK and all its functions and structures is/are not thread-safe by design unless >you call Fl::lock() and Fl::unlock() - which admittedly doesn't solve your problem loading >images in a multi-threaded application, of course. I can see that...

Actually I needed the reminder - I'll have to revisit those functions and see if using them takes care of some of stupid hacks I came up with ...

Albrecht-S:
BTW, OT, FTR: I noticed that your patch uses `FILE *fp` in the header file w/o explicitly including `stdio.h` but compiles w/o error message. The reason is that `stdio.h` gets included (unintentionally) in some other deeply nested include files. One of my goals in FLTK 1.4 development is to get rid of such unintended includes in public header files for (hopefully) obvious reasons: we should not include unnecessary system header files in public headers to avoid increased compile times of user programs and other unintended side effects. It's IMHO also bad style for a library as such and can increase library compile times as well.

I also noticed that Fl_JPEG_Image has the same issue (-Wclobbered) because it also uses FILE *fp (as a local variable) and setjmp/longjmp like the PNG lib does. This needs to be addressed as well.

FYI: I'm now working on removing at least some of those unintended includes (which would make your patch fail, but that's not a problem for now).

Albrecht-S:
> I thought this would be an "easy" thing to provide a fix for ...

In my experience it's often the details that are not easy to fix. Your patch would have fixed a single issue (and yes, I really mean "fixed"), but when you look at the details and, as you wrote, the big picture, things can be different...

Albrecht-S:
@fire-eggs I believe this issue is fixed with commit 7f969b54964381f79558b924184f9c2c9afb4a57.

Kevin, can you confirm that this fixes your multithread issue? If yes, please feel free to close this PR. Thanks.

FTR: the commit mentioned above uses a different approach by allocating the file pointer `fp` with operator `new` which also removes the static variable. The same fix has been applied to Fl_JPEG_Image - which didn't have the static variable but also issued the `[-Wclobbered]` warning (gcc "release" build with optimization).

fire-eggs:
Basic testing shows no obvious problems - seems to be working fine! My thanks for the attention!

Albrecht-S:
Welcome, and thanks for the feedback.

FTR: please ignore the message above reading "Albrecht-S added a commit that referenced this pull request ... Make Fl_Input::handle_key() protected": wrong issue number, should have been issue 211.

PR #240, opened 2021-06-23
failure to call pclose

fire-eggs:
pipe is opened with `popen()` but `pclose()` is not called.

Source:
Coverity Scan: Windows build (VC++ 2019)
Reported as "Resource Leak"


Albrecht-S:
Thanks for the report and the fix. Merged.

@fire-eggs Note: I slightly modified the commit message.

PR #241, opened 2021-06-23
Fix potential memory leak

fire-eggs:
Allocated object not freed on return from successful `hide_common` call.

Source: 
Coverity Scan : Windows build (VC++ 2019)

### DIFF
--- a/src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
+++ b/src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
@@ -475,7 +475,10 @@ void Fl_WinAPI_Window_Driver::hide() {
     }
   }
 
-  if (hide_common()) return;
+  if (hide_common()) {
+    delete[] doit; // note: `count` and `doit` may be NULL (see PR #241)
+    return;
+  }
 
   // make sure any custom icons get freed
 //  icons(NULL, 0); // free_icons() is called by the Fl_Window destructor


src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
```diff
@@ -475,8 +475,13 @@ void Fl_WinAPI_Window_Driver::hide() {
     }
   }
 
-  if (hide_common()) return;
-
+  if (hide_common()) {
+    if (count) {
```
Albrecht-S, line 479: Using `if (count)` is **_potentially_** not correct here because `count` is reset to `0` in line 470 and not necessarily incremented in line 473 (this would depend on another condition).
The correct condition here would be `if (doit)` or just use `delete[] doit;` unconditionally because the argument of `delete[]` is allowed to be a NULL pointer.

Albrecht-S: The same (count may be 0) applies to line 522 (in this modified source file) which should be fixed independently (I will do this in another commit).

Note to myself: the statement `delete[] doit;` in line 522 should be moved below the closing `'}'`, i.e. outside the `if (count) {...}` condition.

Albrecht-S (changes requested):

Albrecht-S:
Thanks for the report. This is really a bug, but the fix is potentially incorrect. See my review comment.

src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
Albrecht-S: Fixed in commit https://github.com/fltk/fltk/commit/d3e8648402a5c1094843e17f8fe72fe628af6e69 (force-pushed).

Albrecht-S (approved):

Albrecht-S:
@fire-eggs I merged your branch and added my own "fix" in commit 75211977b2e. I wrote "fix" in quotes because it's only a formal, hypothetical fix that should never happen in practice -- as long as the two loops setting `count` are identical. After reading the code again I admit that your original fix would have been correct in practice but I merged my changes nevertheless. Hope you don't mind.

fire-eggs:
Just happy to call attention to an issue!

Albrecht-S:
Glad you found and **reported** it. Good job, thanks!

PR #242, opened 2021-06-23
Fix use of variable out-of-scope

fire-eggs:
fixpath goes out-of-scope but is subsequently used as `filename`.

Note: the issue was reported using a Linux build, but would manifest on Windows [backslashes in path].

Source:
Coverity Scan: Linux build (gcc 9.3.0)


Albrecht-S:
Merged. Thanks.

PR #248, opened 2021-07-07
macOS: Fl::add_timeout() must always create a new timer

Albrecht-S:
The old version would reschedule an existing timer if a matching timer
existed already which was unique to macOS.

The new behavior is consistent on all platforms and allows to create
multiple timer entries with the same callback and userdata.

### DIFF
--- a/src/drivers/Cocoa/Fl_Cocoa_Screen_Driver.cxx
+++ b/src/drivers/Cocoa/Fl_Cocoa_Screen_Driver.cxx
@@ -465,18 +465,7 @@ static void do_timer(CFRunLoopTimerRef timer, void* data)
 
 void Fl_Cocoa_Screen_Driver::add_timeout(double time, Fl_Timeout_Handler cb, void* data)
 {
-  // check, if this timer slot exists already
-  for (int i = 0; i < mac_timer_used; ++i) {
-    MacTimeout& t = mac_timers[i];
-    // if so, simply change the fire interval
-    if (t.callback == cb  &&  t.data == data) {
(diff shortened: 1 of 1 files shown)


Albrecht-S:
File [timer2.cxx](https://github.com/fltk/fltk/files/6777809/timer2.cxx.txt) demonstrates the issue. Output on Windows and Linux starts after **one second delay**:
```
$ ./timer2
TICK 1
TICK 2
TICK 3
TICK 4
```
The old version on macOS would output only "TICK1" after **four seconds delay**. The proposed fix makes this consistent on all platforms.

@ManoloFLTK I decided not to commit this directly because I wanted to ask you for your opinion. Please review the patch. IMHO the removed part is just unnecessary and leads to inconsistent behavior but I hope I didn't miss anything.

Thanks in advance for reviewing. I'll pull it into the master branch myself if you don't see any problems.

Albrecht-S:
PS: commit 87475c20d6cc81912e58f21369b101271d4732f6 fixes an unrelated issue, see commit message.

Albrecht-S:
BTW, here's a [link to a tutorial](http://arkiletian.com/cs2/timeouts/timers/) to a pyFLTK tutorial where the author Robert Arkiletian makes use of multiple timeouts with the same ID. This wouldn't work with the old macOS code.

ManoloFLTK:
OK with this change that allows to create multiple timeouts with the same callback and userdata.

Albrecht-S:
@ManoloFLTK Thanks for confirmation. Merged.

Issue #129, opened 2020-08-24
fltk-config needs improvements when created by CMake


Albrecht-S:
Please add all your observations regarding fltk-config (created by CMake) to this issue so we have a record and fltk-config can be fixed.

Albrecht-S:
@imaclmaca wrote in https://github.com/fltk/fltk/issues/115#issuecomment-676698661:

> I find that it [ `fltk-config` ] doesn’t return anything for the "fltk-config --optim” case, it is just blank. ...
> I do not know if this is just that the fltk-config is not set, or if the build really is running without any optimisations enabled!
The autoconf version says:
```
ian$ ./fltk-config --optim
-Os -Wno-deprecated-declarations -Wall -Wunused -Wno-format-y2k -fno-exceptions -fno-strict-aliasing
```

I believe it's both:
1. we don't set all the **additional** compiler options we set with `configure/make`
2. even if users set their own option with CMake (`OPTION_OPTIM`) this is not yet reflected in fltk-config

Note however that CMake uses some default debug and optimization switches to actually **build** `FLTK` **IF** you set an appropriate `CMAKE_BUILD_TYPE` for standard (single config) build types like Makefiles and also for the selected build type when building in multi config builds (VS, Xcode).

Multi config builds (VS, Xcode) have another "special problem": since there are multiple configurations, `fltk-config` can't be uniquely configured during the CMake configuration phase because the compiler switches differ WRT the chosen build type. Hence we might need to create `fltk-config` during the build process (per build configuration) and/or at installation time. I need to check options how to do this...

ManoloFLTK:
About the situation on the macOS platform:

The script BUILD/fltk-config created by cmake fails in my hands for 2 reasons :
1) the compilation option "-isysroot xxx" which determines what SDK version is used is missing;
2) the -lfontconfig present in LDLIBS must be removed.

My belief is that these 3 lines of BUILD/fltk-config
# compiler names
CC="/Applications/Xcode11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc"
CXX="/Applications/Xcode11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++  

should be rewritten as follows (and for my setup):
# compiler names
CC="/Applications/Xcode11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc" \
[...]

PR #310, opened 2021-12-10
fltk-config: Don't use "-luuid" in shared linking flags on Windows.

mmuetzel:
The uuid library just exports various GUIDs. There is no actual shared library with that name on Windows.
The necessary symbols are already (statically) linked into the shared FLTK library. No need to link to it again when linking to the shared FLTK library. It can actually be harmful to link with "-luuid" when (libtool?) projects try to link a shared library with the shared FLTK library.

Remove "-luuid" from the LDFLAGS for "fltk-config --ldflags". But keep it for "fltk-config --ldstaticflags".

### DIFF
--- a/fltk-config.in
+++ b/fltk-config.in
@@ -233,8 +233,18 @@ else
 fi
 
 # Calculate needed libraries
+if test "$OSTYPE" = "msys" -o "$OSTYPE" = "cygwin"; then
+    # Remove "-luuid" from shared linker flags.
+    for lib in $LDLIBS; do
+        if test "$lib" != "-luuid"; then
+            LDSHAREDLIBS="$LDSHAREDLIBS $lib"
+        fi
+    done
+else
+    LDSHAREDLIBS="$LDLIBS"
+fi
 LDSTATIC="$libdir/libfltk.a $LDLIBS"
-LDLIBS="-lfltk$SHAREDSUFFIX $LDLIBS"
+LDLIBS="-lfltk$SHAREDSUFFIX $LDSHAREDLIBS"
 
 if test x$use_forms = xyes; then
     LDLIBS="-lfltk_forms$SHAREDSUFFIX $LDLIBS"


mmuetzel:
Forgot to reference the downstream issue: https://github.com/msys2/MINGW-packages/issues/10293

Albrecht-S:
Thanks for this PR. I see that this might work if the facts that '-luuid' should not be used when linking with shared libs is correct. OK, so far, so good.

However, removing the lib in fltk-config seems to be too late. I'd rather fix it in the build system so it is not exported to fltk-config in the first place.

I'll investigate this and follow-up later.

I have one question though: what is your use case, i.e. how is fltk-config generated in this environment? Is it with configure/make or CMake?

FYI: Unfortunately we still have both configure/make and CMake (sigh), and both *may* generate slightly different fltk-config files (which is another fact that needs investigation).

I'll labe this PR with `CMake` but I really need to know which of the two build systems triggered this PR. Thanks.

Albrecht-S:
See also #129

mmuetzel:
This was with the `fltk-config` in MSYS2. It is built with cmake (ninja):
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-fltk/PKGBUILD

I might be misunderstanding. But this only works because the shared fltk library already exports the necessary uuid symbols. So it should continue to be linked to `-luuid`.
The issue is only for projects that link a shared library to the shared fltk library. And maybe also only if they are using libtool for linking that shared library.

We came across this originally when trying to build Octave on Windows. It fails to link its shared libraries if `-luuid` is used. That flag was inherited from `fltk-config`. Octave uses libtool in case that matters.

Albrecht-S:
@mmuetzel Thanks for the additional explanations. I see my questions answered.

The only reason I asked for details was because I wanted to generate `fltk-config` properly in the first place rather than "removing" `-luuid` later. Since we have the two build systems to maintain it was important to know by which of the two you created your fltk-config.

I'll see what I can do and let you know. If this doesn't work out as I expect then your patch might be the solution. Thanks.

mmuetzel:
I guess the same issue occurs when building with `make`. See also: https://savannah.gnu.org/bugs/?39066
That is about a similar issue in MXE Octave which uses autotools to build fltk.

Albrecht-S:
Thanks again, I would have double checked this anyway, but it's good to know in advance.

PR #313, opened 2021-12-11
Fluid: fixed subtype menu to clear the menu when deactivated.

jdpalmer:
If the subtype menu is disabled because there are no subtypes, the deactivated menu will still display the subtype of the last widget that had a subtype value.  Clearing the menu when there are no subtypes fixes this minor inconsistency.

### DIFF
--- a/fluid/Fl_Widget_Type.cxx
+++ b/fluid/Fl_Widget_Type.cxx
@@ -1728,7 +1728,11 @@ Fl_Menu_Item *Fl_Widget_Type::subtypes() {return 0;}
 void subtype_cb(Fl_Choice* i, void* v) {
   if (v == LOAD) {
     Fl_Menu_Item* m = current_widget->subtypes();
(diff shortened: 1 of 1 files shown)


MatthiasWM:
Thank you very much for your contribution. Indeed, the menu is not updated correctly. I took your very helpful code and changed it into displaying "Normal" as a menu label, instead of an empty Choice (which would be unexpected for the user IMHO). The Choice is still deactivated as before. I played with the idea of hiding it, but that came even more unexpected.

MatthiasWM:
To clarify, I did not reject your patch, I just rewrote it slightly. Hope that's ok ;-)

jdpalmer:
Perfect.  I had thought about doing the same thing you did but couldn't decide on a "default" subtype name.

MatthiasWM:
I just stole it from the other menus... .

PR #315, opened 2021-12-12
Add support for making FLUID optional during compile/install (CMake)

AlexMax:
This required the concept of a blank `FLTK_FLUID_EXECUTABLE` with the patching of various places in the build that took its existence for granted.  

One thing that I was not sure about was handling FLUID in cross-compilation situations.  I don't really know what the right play is there, since cross-compilation is not my forte.


Albrecht-S:
Thank you for this PR. I've seen such requests already and I will take a look into this. There are good parts which can be used but there are also some parts I would implement differently. I'll let you know the details later when I can find the time.

Can you explain why you would not want to build fluid? Building fluid doesn't take much time and is a prerequisite for building some of the test programs (which I also noticed that your patch takes care of).

Albrecht-S:
@AlexMax Just in case this PR was caused by your report on Gitter/Matrix that you could not build fluid, here's the proposed solution by Matthias (@MatthiasWM):

Citation:

<MatthiasWM (Matthias Melcher)> It's not Fluid that is failing to build. Your setup does not build fltk_images correctly because it does not find the (correct) jpeg library. Just build FLTK with its internal libs:
```
CMake ... -D OPTION_USE_SYSTEM_LIBJPEG=Off \
-D OPTION_USE_SYSTEM_ZLIB=Off \
-D OPTION_USE_SYSTEM_LIBPNG=Off
```

AlexMax:
> Can you explain why you would not want to build fluid? Building fluid doesn't take much time and is a prerequisite for building some of the test programs (which I also noticed that your patch takes care of).

Your subsequent guess was sort of correct.  Building FLUID was resulting in libjpeg problems that _only_ occurred in Github Actions builds, and given the fact that my use of the library doesn't even use libjpeg or FLTK in the first place, I figured that simply being able to disable the FLTK build was more expedient than trying to debug a CI build, especially when I discovered that I was not the first one to request a FLUID toggle.

For what it's worth, `-DOPTION_USE_SYSTEM_LIBJPEG=OFF` wasn't a fix for my issue either.  A [build log](https://paste.ee/p/oMdie) shows the files being built, but missing jpeg symbols when the time came to link into FLTK.

MatthiasWM:
Thanks for the great PR. I did a few changes and started a new PR #539 which helps me with cross compiling as well.

PR #316, opened 2021-12-12
Fluid: "Widget Class" subclasses Group but incorrectly shows Window subtypes.

jdpalmer:
In Fluid, if you select New>Code>Widget Class, the resulting entry shows subtypes for Fl_Window but the generated code subclasses Fl_Group (i.e., the selected subtype is ignored).


MatthiasWM:
The subtype menu has no meaning in Widget Classes and should be graded out. To get the Window superclass API, the user has to enter a "Class" that contains the word "Window".

jdpalmer:
Right.  This patch changes subtypes() to return 0 for Widget Class thereby disabling the subtype menu.

PR #320, opened 2021-12-13
STR 3442: copy/paste now inserts widget as expected.

MatthiasWM:
Fluid would add pasted widgets as the last child of the
current group, even if they were cut at the beginning
of the group. This patch adds a 'startegy' to adding
widgets to allow merging closer to the 'current' widget.

Also added d'n'd for the widget bin and some other
UI improvements.

STR 3442: some preparation, variable renaming

STR 3442: unifying functions to add widgets.

Adding widgets interactively is merged into one function,
making undo/redo work as expected and removing
unexpected differences between adding by menu or bin.

STR 3442: adding drag'n'drop for windows from bin.

Just drag the window symbol out of the tree and drop it anywhere
on the desktop.

Visual representation of dragged Window.

STR 3442: insert individual nodes after 'current'

STR 3442: adding new widget at expected positions.

The widget bin can finally do drag'n'drop.
Widgets appear in the tree close to the last selection.
Duplicating now ads right below.
Cut and paste inserts after last selection.

Fluid: replaced #define with strict enum type.

### DIFF
--- a/fluid/CMakeLists.txt
+++ b/fluid/CMakeLists.txt
@@ -24,6 +24,7 @@ set (CPPFILES
   Fl_Widget_Type.cxx
   Fl_Window_Type.cxx
   Fluid_Image.cxx
+  Shortcut_Button.cxx
   about_panel.cxx
   align_widget.cxx
   alignment_panel.cxx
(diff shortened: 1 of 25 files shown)


MatthiasWM:
Don't merge this. There is still a bug.

Albrecht-S:
Reopening, assuming it was closed accidentally: "There is still a bug".

MatthiasWM:
This was opened and closed again by myself because it did not catch all the changes that I wanted to make.

PR #321, opened 2021-12-13
Str3442: Fluid workflow and UI improvements

MatthiasWM:
Copy/Paste after current widget.
Duplicate in place.
Drag'n'drop from the widget bin.


MatthiasWM:
Living on the edge...

PR #343, opened 2021-12-19
STR 3289: Fluid i18n, gettext, catguts improvements

MatthiasWM:
In this patch I have been trying to pull all the internationalisation woes together and solve them hopefully to everyones liking. I implemented what the [GNU foundation](https://www.gnu.org/software/gettext/) suggests for handling static text.

The original implementation in FLTK 1.1 has worked for some devs well. I am not sure if the introduction of textdomains into GNU gettext started to break things, but I proceeded in [2010](https://www.fltk.org/str.php?L2246) to disimprove the situation by moving the unsupported static initialisation into a dynamic translation later in the code. While this by itself was not wrong, it did remove the marker that was needed to extract the translatable text from the source code.

See [STR 3289](https://www.fltk.org/str.php?L3289), [STR 2861](https://www.fltk.org/str.php?L2861), and [STR 2678](https://www.fltk.org/str.php?L2678) for my list of shame.

This fix reintroduces the urgently needed markers while still moving initialisation to the GNU recommended point in time.

I added two project settings that allow users to define the static marker function, just like the original `gettext`. A second setting expects a preprocessor macro. If that macro is not defined, the entire i18n becomes pretty much transparent at compile time, and the code can still be compiled on all platforms, even if the build system finds no gettext library. Check `test/preferences.fl` as an example.

As a last little extra, I added full support for menu items that have an image *and* text at the same time from within Fluid (just set the label text and the image field at the same time), including i18n of course.


MatthiasWM:
And thanks to autocorrect, it should have been POSIX's `catgets` in the headline, and it's not related to feline intestines.

Albrecht-S:
@MatthiasWM You know you can edit the title at least in the GitHub PR? Done.

Unfortunately we can't change the commit message. :-(

PR #378, opened 2022-01-25
Background color attribute for Fl_Text_Display.

MatthiasWM:
The implementation stays source code compatible with older versions of `Style_Table_Entry`. The background color field was added which can stay the default value 0. The previously existing `attr` field now supports two flags, `ATTR_BGCOLOR` and `ATTR_BGCOLOR_EXT`. The first version changes the background of text with this attribute. The _ext version expands the background color to the end of the line if the line end or breaks.

I also fixed the clipping at the end of tile changes and at the end of the line. This was very visible with italicised characters that may bleed into the next attribute.

### DIFF
--- a/FL/Fl_Text_Display.H
+++ b/FL/Fl_Text_Display.H
@@ -156,12 +156,11 @@ public:
     ATTR_BGCOLOR        = 0x0001, ///< use the background color
     ATTR_BGCOLOR_EXT_   = 0x0002, ///< internal use
     ATTR_BGCOLOR_EXT    = 0x0003, ///< extend background color to the end of the line
-    ATTR_UNDERLINE      = 0x0004, ///< a single underline, underline types ar mutually exclusive
+    ATTR_UNDERLINE      = 0x0004, ///< a single underline, underline types are mutually exclusive
     ATTR_GRAMMAR        = 0x0008, ///< grammar suggestion (blue dotted underline)
     ATTR_SPELLING       = 0x000C, ///< spelling suggestion (red dotted underline)
     ATTR_STRIKE_THROUGH = 0x0010, ///< line through the middle of the text
     ATTR_LINES_MASK     = 0x001C, ///< all underline and strike through types combined
-    // how about underscore, strikethrough, undersquigle(color)?
   };
 
   Fl_Text_Display(int X, int Y, int W, int H, const char *l = 0);
--- a/FL/Fl_Text_Display.H
+++ b/FL/Fl_Text_Display.H
@@ -146,6 +146,17 @@ public:
     Fl_Font     font;   ///< text font
     Fl_Fontsize size;   ///< text font size
     unsigned    attr;   ///< currently unused (this may be changed in the future)
+    Fl_Color    bgcolor;///< text background color if ATTR_BGCOLOR or ATTR_BGCOLOR_EXT is set
+  };
+
+  /**
+   attribute flags
+   */
+  enum {
+    ATTR_BGCOLOR      = 0x0001, ///< use the background color
+    ATTR_BGCOLOR_EXT_ = 0x0002, ///< internal use
+    ATTR_BGCOLOR_EXT  = 0x0003  ///< extend background color to the end of the line
+    // how about underscore, strikethrough, undersquigle(color)?
   };
 
   Fl_Text_Display(int X, int Y, int W, int H, const char *l = 0);
--- a/FL/Fl_Text_Display.H
+++ b/FL/Fl_Text_Display.H
@@ -153,9 +153,13 @@ public:
    attribute flags
    */
   enum {
-    ATTR_BGCOLOR      = 0x0001, ///< use the background color
-    ATTR_BGCOLOR_EXT_ = 0x0002, ///< internal use
-    ATTR_BGCOLOR_EXT  = 0x0003  ///< extend background color to the end of the line
+    ATTR_BGCOLOR        = 0x0001, ///< use the background color
+    ATTR_BGCOLOR_EXT_   = 0x0002, ///< internal use
+    ATTR_BGCOLOR_EXT    = 0x0003, ///< extend background color to the end of the line
+    ATTR_UNDERLINE      = 0x0004, ///< a single underline, underline types ar mutually exclusive
+    ATTR_GRAMMAR        = 0x0008, ///< grammar suggestion (blue dotted underline)
+    ATTR_SPELLING       = 0x000C, ///< spelling suggestion (red dotted underline)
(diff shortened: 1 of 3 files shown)


Albrecht-S:
> I also fixed the clipping at the end of tile changes and at the end of the line. This was very visible with italicised characters that may bleed into the next attribute.

What are "tile changes"? A typo or ...?

Does this also include end of line issues without background changes? This is a long standing issue which you can see if you open a source file in test/editor with contents like:
```
//
// more comments
```
The first line with the double slash cuts off a tiny bit at the top of the second slash. If you found how to fix this, does it also apply to this case?

image


**Regarding review:** I'll take a closer look tomorrow. I took a quick look but it's way too much code to review details now (too late, I feel tired and unconcentrated).

MatthiasWM:
Tile changes -> Style changes, sorry. Yes, it does fix those clipping issues too. It's done by rendering all the background in a line first, and then all the text elements. It still has some artefacts when only a partial line is updated, but they are much less visible now.

Some of the code changes are only indents because I loop the rendering routine twice now. I also changed 'editor' to showcase both background modes for a C++ source file in line and block comments.

Albrecht-S:
@MatthiasWM Working on the review now...
> Yes, it does fix those clipping issues too.

Unfortunately not. :-(
image

Clipping is still present in the first line of the screenshot. A trailing `'.'` (second line) or space (third line) prevents the clipping artifact.

To be continued...

src/Fl_Text_Display.cxx
```diff
@@ -2293,19 +2301,27 @@ void Fl_Text_Display::draw_string(int style,
...
 
     if (style & PRIMARY_MASK) {
       if (Fl::focus() == (Fl_Widget*)this) {
-        if (Fl::screen_driver()->has_marked_text() && Fl::compose_state)
-          background = color();// Mac OS: underline marked text
-        else
+        if (Fl::screen_driver()->has_marked_text() && Fl::compose_state) {
+          background = bgbasecolor; // Mac OS: underline marked text
+        } else {
           background = selection_color();
```
Albrecht-S, line 2311: I like the additional curly braces! This makes the nested if/else constructs better readable.

Albrecht-S (approved):
I must admit that I don't understand the code of Fl_Text_Display completely, particularly not the drawing code with all these styles etc..
However, I can't see anything wrong in the diffs, hence :+1:

Albrecht-S:
My review pertains to the proposed changes in this PR.

The clipping issue is a different aspect that needs not be resolved with this PR. But it would be nice to have. ;-)

MatthiasWM:
> @MatthiasWM Working on the review now...
> > Yes, it does fix those clipping issues too.
> Unfortunately not. :-( image
> 
> Clipping is still present in the first line of the screenshot. A trailing `'.'` (second line) or space (third line) prevents the clipping artifact.

It works for me here (macOS):

Screenshot 2022-01-26 at 22 00 40

Albrecht-S:
@MatthiasWM wrote:
> It works for me here (macOS)

Yes, I can confirm this too. I double-checked my Linux build (git pull + complete rebuild) and the issue is still visible. Another Windows/MinGW build (cross-compiled under Linux, executed by `wine`) shows the same clipping effect but less visible.

My guess is that we're using fl_measure() to set the clipping area for the text and that the results thus depend on the platform and/or font (attributes).

If my guess is correct and we're setting a clipping area for text drawing, wouldn't it now (with the latest changes in mind) be possible to draw text without setting the clipping at all? Maybe unless it's drawing at the widget borders.

Albrecht-S:
I think we'd better open a new issue for the clipping problem but for now I'd like to mention [STR 3364](https://www.fltk.org/str.php?L3364): _"**Fl_Text_Display: wrong text width calculation**"_ which includes a somewhat elaborate patch that adds a whole bunch of Windows specific text calculations and a "trick" to use fl_text_extents() only in some border cases. Interesting.

I'm not able to investigate this right now but if **this** issue is closed with a solution of the clipping issue then the related `STR 3364` could probably closed as well. Otherwise a new issue should contain the reference to `STR 3364`.

FL/Fl_Text_Display.H
```diff
@@ -153,9 +153,13 @@ public:
    attribute flags
    */
   enum {
-    ATTR_BGCOLOR      = 0x0001, ///< use the background color
-    ATTR_BGCOLOR_EXT_ = 0x0002, ///< internal use
-    ATTR_BGCOLOR_EXT  = 0x0003  ///< extend background color to the end of the line
+    ATTR_BGCOLOR        = 0x0001, ///< use the background color
+    ATTR_BGCOLOR_EXT_   = 0x0002, ///< internal use
+    ATTR_BGCOLOR_EXT    = 0x0003, ///< extend background color to the end of the line
+    ATTR_UNDERLINE      = 0x0004, ///< a single underline, underline types ar mutually exclusive
```
Albrecht-S, line 159: s/ar/are

Albrecht-S:
commit c3492baa7ffe04314c99c996c9025020661ecaa5 is **MUCH** better now (Linux + Windows).

:+1: :1st_place_medal:

Thanks!

Though nitpickers would say that the background color should be a little wider (2-3 px), but that's really a minor issue. Text is much more important than background color.

src/Fl_Text_Display.cxx
```diff
@@ -2351,7 +2351,8 @@ void Fl_Text_Display::draw_string(int style,
     // Make sure antialiased ÄÖÜ do not leak on line above:
     // on X11+Xft the antialiased part of characters such as ÄÖÜ leak on the bottom pixel of the line above
     static int can_leak = Fl::screen_driver()->text_display_can_leak();
-    if (can_leak) fl_push_clip(X, Y, toX - X, mMaxsize);
+    // Clip top an bottom only. Add margin to avoid clipping horizontally
+    if (can_leak) fl_push_clip(x(), Y, w(), mMaxsize);
```
Albrecht-S, line 2355: s/an/and

Sorry.

FL/Fl_Text_Display.H
MatthiasWM: Sigh, typos.

MatthiasWM:
> Though nitpickers would say that the background color should be a little wider (2-3 px)

Yes, I saw that, but then we get a bad background at the start of non-italics. It gets more complicated when an _M_ and M next to each other actually leak into each other. I think I'll leave it at this.

So this is a mix of underline styles as an example. The thickness of the lines is unfortunately only an integer, so lines may appear a bit thick or thin, depending on scaling.

Screenshot 2022-01-27 at 01 18 20

PR #379, opened 2022-01-25
Consolidate timeout handling across platforms

Albrecht-S:
Add Fl_Timeout class

Debug output function Fl_Timeout::debug() may be removed later.

Output debug information in some test programs

Move platform independent code of Fl::wait() to main part

- basic timeout handling
- Fl::run_checks()
- Fl::idle()
- waiting time calculation (next timeout)
- remove platform dependent "screen driver" stuff

Test programs test/timer_test.cxx and test/timer_race.cxx may be
removed or modified later for "better" timeout tests.


Albrecht-S:
This is an attempt to unify timeout handling across platforms. All timeout related code is now platform independent.

Please review and test if you can find any issues with "real" programs that depend on `Fl::add_timeout()` and similar.

MatthiasWM:
It works perfectly fine for me on macOS 12.0.1 Monterey on the M1 processor.

ManoloFLTK:
I believe the creation of the new Fl_Timeout class reveals a suboptimal aspect of classes Fl_System_Driver and Fl_Screen_Driver, namely that a few member functions of Fl_Screen_Driver should rather be members of Fl_System_Driver.

Fl_Screen_Driver should contain what varies with the screen-related aspects of the platform.
Fl_System_Driver should contain what varies with the platform OS.
The (future) Wayland platform introduces a platform that changes completely screen-aspects of a preexisting platform, X11, while being largely, but not entirely, identical at the OS level to the X11 platform.
This justifies the new class Fl_Unix_System_Driver that both X11 and Wayland platforms use
for overriding most members of Fl_System_Driver.

The present Fl_Screen_Driver::wait(double) member function should be moved to Fl_System_Driver.
This allows Fl::wait(double) to be written
```
double Fl::wait(double t) {
 // platform independant part
……
// platform dependant part
return Fl::system_driver()->wait(t); //<--- the only proposed change is here
}
```
and to trigger the shared Fl_Unix_System_Driver::wait(double) override on both X11 and Wayland.

If we don't do that, we have to create 2 overrides of Fl_Screen_Driver::wait(double) for X11 and Wayland, to add a new member, wait(double) to Fl_Unix_System_Driver class, and to have each call ((Fl_Unix_System_Driver*)Fl::system_driver())->wait(t).  Independently from introduction of the Wayland platform, I believe making the process wait for events of all kinds is more system- than screen-related.

The same reasons suggest to replace Fl_Screen_Driver::ready() by Fl_System_Driver::ready(). 
Also, 2 protected members of Fl_X11_Screen_Driver, poll_or_select() and poll_or_select_with_delay(), would be better members of Fl_Unix_System_Driver. Finally, 2 static functions present in Fl_x.cxx, create_bmp() and own_bmp_to_RGB(), should be better members of Fl_Unix_System_Driver.

@Albrecht-S Wat do you think? The option to change Fl_System_Driver and Fl_Screen_Driver or not needs to be agreed upon before committing, in any order, the new Fl_Timeout class and Wayland platform.

Albrecht-S:
@ManoloFLTK Very good points.

I also noticed that there are some misplacements, particularly the (in my branch) already removed `Fl::add_timeout()` etc.). I didn't want to touch too much code though because it's largely unrelated to the Fl_Timeout changes, but I agree. Your analysis sounds sensible, I'll look into the details and follow-up.

ManoloFLTK:
I have used the new code under macOS and noticed that all features I could think of work well:
- 2 timeouts set to the same timepoint, as in test/clock or test/unittests, apparently trigger window refreshes "simultaneously";
- test/mandelbrot and its idle callback run correctly
- the changes in src/Fl_cocoa.mm have left the delicate use of class NSAutoreleasePool in an adequate state.

I had written before I find the procedure "awkward". Specifically I refered to the fact the algorithm changes the scheduled time of every other timeouts each time one runs. Thus, timeouts aren't independent from each other. But I see the clear benefit of this approach: everything becomes platform-independent except a call to Fl_System_Driver::gettime() which moreover is used by other FLTK features.

+1 to commit on my side.

Albrecht-S:
@MatthiasWM Thanks for your test report.

@manolo Thank you very much for your review and reporting your test results. The point about `NSAutoreleasePool` was indeed my biggest concern. I'll squash/rebase/push my branch after the final cleanup. You wrote:
> timeouts aren't independent from each other.

In fact all timeouts are independent of each other. All timeouts are maintained in a queue ordered by expiration time and each timeout's remaining time is decremented separately. There's one "simplification" though: the "delta" time used to decrement all queued timeouts is calculated only once for all timeouts, hence all timeouts must be decremented by this delta time together. However, this doesn't affect the independence of the timeouts.

Issue #252, opened 2021-07-13
Merge improvements of NTK (FLTK fork) into FLTK (cairo, speed)


grammoboy2:
"
There is a really nice project based on FLTK here:

https://non.tuxfamily.org/

It uses the NTK toolkit, which is a fork of FLTK:

https://non.tuxfamily.org/wiki/NTK

Due to some community issues it seems offline at the moment, but the code is also here:

https://github.com/falkTX/ntk
"

Albrecht-S:
I'm not familiar with NTK. Please tell us which improvements you'd like us to merge.

W/o looking at and w/o any knowledge of NTK I wonder why FLTK should backport changes from NTK rather than users use NTK directly? Since you seem to know NTK and its features worth backporting or merging, please elaborate. Thanks.

xahodo:
The project seems to have low activity.

Some of the claimed sweetness is transparent and overlapping widgets, but doesn't fltk support overlapping widgets already?

Albrecht-S:
> The project seems to have low activity.

This is one reason why the OP opened this issue (potential lack of maintenance) as he told me in private mail.

**Overlapping widgets** are not officially supported by FLTK although there's nothing that would prevent using it. Not officially supported does only mean that it may work as expected or not, maybe depending on the platform or something else. It's just that we don't guarantee that overlapping widgets work. Of course all Fl_Group derived widgets and their respective children must overlap, i.e. the child widgets must be contained in their parents, but "overlapping" in this context means probably that siblings overlap.

Regarding transparency: I've seen that the NTK maintainer added an alpha channel to Fl_Color (IIRC). This is an extension of the FLTK color model which uses R, G, B + I where I is an index in the color palette. In general you can only use one of (R, G, B) or I and he replaced the interpretation of I with alpha if at least one of R, G, B is non-zero (if I understood this correctly). How drawing with transparency works I can't tell, but I assume that this works with Cairo which is the only graphics option in NTK. I'm not sure if this can ever be an option for FLTK.

Albrecht-S:
Regarding Cairo drawing (development roadmap):
1. We're striving to get Wayland support on Unix/Linux as a replacement of X11 drawing included in the 1.4.0 release. FLTK's Wayland drawing will be based 100% on Cairo.
2. Once Wayland support is implemented we plan to add a Cairo drawing option on Unix/Linux platforms which would then use the existing Cairo graphics driver.

[...]

PR #385, opened 2022-02-01
OpenGL implementation of all `fl_` "Drawing Fast Shapes" graphics calls

MatthiasWM:
I am investing quite some time to make the OpenGL version of fl_line/rect/etc. graphics calls pixel-compatible to their native counterparts. So far, everything non-round works pretty well. 

This allows any FLTK widget and widget hierarchy to be used inside an Fl_Gl_Window. The widgets should look exactly the same, but float on top of whatever background the user is rendering.

UnitTest was redone to make native and GL drawing comparable and verifiable. This will also help me with writing the SDL port which I hope to do before 1.4.0. It would make FLTK work on iOS (iPhone, iPad) and Android. This would also match the proposed addition of making FLTK touch friendly, and is much less ambitious than the Pico/Nano route.

Lastly, I would love to add an alpha channel API to make box types for overlay use in OpenGL widgets. So far, I am thinking about adding `fl_alpha()` complementary to `fl_color()`, and additional calls to set the lookup table. Even though we define color in four bytes, the LSB is reserved of indexes colors and can't be abused for an alpha channel.

Suggestions appreciated.


Albrecht-S:
Two comments:

1. I'm wondering if this is really something we should do such a short time before the 1.4.0 release. Thus I'm also not sure if this is a valid addition to the 1.4.0 milestone. However, if it would be doable seamlessly w/o much risk of disturbing other functions, well, then I wouldn't object.

2. Alpha channel: Currently we have an ambiguity in our Fl_Color representation: there must either be a color index (I) in the low byte or (if the low byte is zero) the high order R,G,B values are used. The ambiguity is when both (R,G,B) and (I) are non-zero. A new interpretation could be either (R,G,B,A) or (0,0,0,I), i.e. the color table index is only used if R=G=B=0, otherwise it's interpreted as the alpha channel. The only problematic value (0,0,0,0) would need to be defined (either black or color index 0). IIRC the NTK fork of FLTK did something like this. See issue #252 for links to NTK.

Albrecht-S:
Addition: I believe that changing the interpretation of color values has potential to break user applications and should definitely not be done before 1.4.0 gets released. But it's an option for future enhancements.

test/unittest_images.cxx
```diff
@@ -295,4 +295,4 @@ Fl_RGB_Image *ImageTest::i_ga = 0;
 Fl_RGB_Image *ImageTest::i_rgb = 0;
 Fl_RGB_Image *ImageTest::i_rgba = 0;
 
-UnitTest images("drawing images", ImageTest::create);
+UnitTest images(7, "drawing images", ImageTest::create);
```
Albrecht-S, line 298: If I understand (guess) this correctly "7" is the index in the list of unittests (left box). I suggest to use named constants (enum) for the index. This enum could be defined in a central place (header file) so index changes would be easier (inserting a new test at an arbitrary place).

MatthiasWM:
1. This feature and its API exist in "master" since at least Jan. 2016, but it was buggy and undocumented. I am fixing things and making them universally usable. I will modify or add a test to make the feature more visible. I think it is a great step towards more modern UI's.

2. Yes, it's a problem for all black values, no matter what the alpha value should be. There is no sane solution to this, and I don't want to change existing code. My solution would be an alpha component in the lookup table, because that's where color indexes get converted to RGB, and after the conversion, the value is no longer ambiguous. An alpha channel, from a GUI standpoint, makes only sense if you want widgets hovering over content (vs. widgets over widgets, content over widgets), and for that limited use, index lookups are good enough.

Example use: https://youtu.be/2CII8zb-rcs?t=890 and https://youtu.be/sq4TVNilfGc?t=66

test/unittest_images.cxx
MatthiasWM: Makes sense. I'll do that.

Albrecht-S:
> My solution would be an alpha component in the lookup table, ...

Do you mean the "color map" with "lookup table"?

Wouldn't this be very restrictive because ...
1. we have only 256 entries in this table
2. you would need to change the color map if you want to change the alpha value for a particular drawing operation or color
3. you couldn't draw arbitrary RGB colors with arbitrary alpha (w/o always changing the color map)?

Am I missing anything or did I misunderstand what you wrote?

I agree with everything else you wrote.

MatthiasWM:
The videos both feature transparent UI elements over a 3D scene. The first one has just some buttons. In the second one, the entire hierarchy floats over the 3d scene. In FLTK, this could be done with an Fl_Tree_View with text that has no transparency, but a box type that does have some alpha value.

My point is, that a UI has usually very few colors to look consistent, whereas the content (3d scene, for example) can use millions of colors with any alpha value. My API idea so far is:

`fl_alpha(uchar a)` sets an alpha value for the following operations, just as `fl_color(...)` does for color.
`fl_color(Fl_Color rgbx, uchar a)` does both in a single command
`fl_color(uchar r, uchar g, uchar b, uchar a)` would make sense

As for the color lookup table, I would add
`fl_set_color(Fl_Color index, Fl_Color c, uchar a)` and
`fl_set_color(Fl_Color index, uchar r, uchar g, uchar b, uchar a)`

No tested yet, but I think that there are no conflicts with existing APIs.

So if you need an Fl_Button with 10% on the background and 90% on the text, you could
```
fl_set_color(32, 128, 128, 128, 25);
fl_set_color(33, 0, 0, 0, 221);
myButton->color(32);
myButton->textcolor(33);
```
I briefly considered using bit 8 (0x00000100) to indicate RGBA vs index vs RGB, especially since the least significant bit in blue makes the least perceptible cool change, but it feels like a bad hack that would haunt us forever.

MatthiasWM:
Actually, your comment about the video proved a point for myself: it seems to be quite natural for users to have text floating over content, maybe the result of closed captioning/subtitles in movies that many viewers see to switch on, even if they have no hearing impairment.

Albrecht-S:
OK, I understand your thoughts about the API. This looks all plausible, but my point stands that there are only a limited number of colors. Some of these are reserved ("gray ramp" etc.). I remember that we (my previous company) redefined the upper part (128-255) or some other part for "our own" color map which we set with constant values. Maybe 64 values would've been enough, but anyway. Color indices in the color map are a valuable resource, we should take care.

That said, I understood also your assumption that many GUI's would only need a few colors for widgets to be consistent. So maybe it's a way we can go.

For direct drawing `fl_color(uchar r, uchar g, uchar b, uchar a)` and similar API's would definitely make sense.

test/unittest_images.cxx
MatthiasWM: Done in my repo

MatthiasWM:
This implementation is not completely finished yet. I am considering merging it into master, because it is "good enough" in rendering FLTK widgets in OpenGL contexts (Fast Rendering functions). I would then open new tickets for Complex Shapes in GL, images in GL, and the transparency/alpha features we discussed above.

Albrecht-S:
I'm wondering (I have no idea how OpenGL works, please take this comment as a question). I'm just skimming through the "Files changed" list. We have:
```
void Fl_Gl_Window::draw_begin() {
  ...
  glOrtho(-0.5, w()-0.5, h()-0.5, -0.5, -1, 1);
  ...
}
```
which looks like a global coordinate transformation. I take it this compensates for different coordinate bases using a translation of `-0.5` and the window size transformation to (1.0, 1.0). Correct?
But then we have lots of similar (explicit) coordinate translations of `±0.5` in the GL drawing code, like in `src/drivers/OpenGL/Fl_OpenGL_Graphics_Driver_rect.cxx`:
```
void Fl_OpenGL_Graphics_Driver::rectf(int x, int y, int w, int h) {
  ...
  glRectf(x-0.5f, y-0.5f, x+w-0.5f, y+h-0.5f);
}
```
and finally in the "application (test) program" `test/unittest_circles.cxx` a translation by `0.5`:
```
void arc(int xi, int yi, int w, int h, double a1, double a2)
{
  if (a2<=a1) return;

  double rx = w/2.0;
  double ry = h/2.0;
  double x = xi + rx + 0.5;
  double y = yi + ry + 0.5;
 ...
}
```
Is this all correct and consistent? If this was a user program, would a user expect to need to do this `+0.5` coordinate translation?

I'm confused. I would tend to believe that neither the application program nor all the internal gl_draw_x() functions (should) need further coordinate transformations once the global coordinate transformation is set up?

This is just a thought, I didn't test anything. But again, I have no idea how this all works together. Please enlighten me. ;-)

Albrecht-S:
Coming back to `Fl_Color` with alpha channel. I thought about it again and here's an option we might consider.

If we say R = red, G = green, B = blue, I = index in color map, and A = alpha, then we have either (R, G, B, A) or (R, G, B, I) with the known ambiguity if `I > 0` **and** `(R | B | G) > 0` which is an invalid combination. I don't know which resolution the current code would use: (R,G,B) or (I), but this doesn't matter. Note: (R | G | B) means bitwise OR.

A possible way out of this dilemma would be a minimal compromise if we defined:
- if `(R | G | B) == 0` then the color is interpreted as (0, 0, 0, I) with index I from the color map (index 0 == black)
- if `( R | G | B ) != 0` then the color is interpreted as (R, G, B, A)

This definition would give previously invalid combinations a valid meaning.

The **only minor drawback** would be that you can't use "real black" with alpha directly as an `Fl_Color` value but the tuple (1, 1, 1, A) would be interpreted as a **_very dark gray_** with alpha channel A. Even (0, 0, 1, A) would be "almost black with alpha A".

The color map would still get a new (optional) alpha component so (0, 0, 0, I) could be a color map entry with alpha channel as you (@MatthiasWM) suggested and in color map entry `I` users could set "real black" with alpha `A` by calling
```
  fl_set_color(I, 0, 0, 0, A); // color map entry I is black with alpha channel A
```

Full backwards compatibility would be given, provided all color values used in existing programs were valid, i.e. either `(0, 0, 0, I)` (index) or `(R, G, B, 0)` (RGB color). IMHO we don't need to care about programs using other (invalid) `Fl_Color` combinations.

MatthiasWM:
About `glOrtho`: this is the officially recommended use to map pixels 1:1. Without the offset, integer coordinates would point at the bottom left of a pixel instead of the center of a pixel. With antialiasing enabled, OpenGL would draw 4 pixels at 25%. The same is true for line drawing. But - and this is the crux about the rasterisation - when drawing a rectangle, the bottom left *should* be between pixels, which results in all those 0.5 offsets for pixel drawing vs. vector drawing.

Since nobody really follows the original suggestion of using the 0.5 offset, we will get unexpected results. One of those is text drawing in FLTK which expects no 0.5 offset. I added that too, but that will be bad for existing software and make text blurry.

tl;dr: I will revert the offsets to 0 against Khronos' recommendation and adjust offsets until I have a perfect match.

MatthiasWM:
Much improved, alpha channel works pretty well (see Cube demo)

MatthiasWM:
This completes the implementation of all calls relevant for drawing FLTK widgets in OpenGL space. It also adds transparency to the colormap and rect RGBA color rendering in `fl_..` calls. Added documentation to the OpenGL section of the docs.

There will be a second part to this eventually that verifies the rendering of "Complex Graphics", and a third one that implements rendering images using `fl_...` calls. At some point, I would like to improve the Code example to demo the capabilities a bit better.

pdhahn:
First I'd like to express how grateful we folks who dabble in OpenGL + FLTK are to hear about this progress. I have previously ported qwtplot3d from Qt to FLTK for internal use, and I hope to take advantage of your changes.

Humble suggestion: perhaps one of these days it might be useful to think about adding a new "extended" color type Fl_Color64 that is a 64-bit long(-long) int vs. the current Fl_Color which is a 32-bit int. The extended type would give plenty of room both for RGBA and an expanded color table index.

MatthiasWM:
Interesting. Care to share what kind of work you do with qwtplot3d? 

This particular ticket implements 2D widgets as an overlay for OpenGL windows. I will write some 3D manipulators for my slicer app eventually, and somehow I feel that a 3d UI could be part of FLTK, but it may require another library to keep overhead for 2D users low.

Fl_Color64 is another good approach for color, especially with new PCs now supporting 12 bit colors. Before we leave the "Fast and Light" part though, I'd like to check if the current implementation is already sufficient.

pdhahn:
Re. qwtplot3d - what I can share openly is that I largely use it to more effectively visualize correlations vs. just using a "flat", colored 2D mapping to do so.

BTW the port was nontrivial, but fortunately I came out of a Qt programming world in seismic processing UI's. Naturally, I used FL_Gl_Window, but that left me wishing that there were an actual FLTK OpenGL-oriented, _Fl_Group_-based class (not _Fl_Window_-based) to derive from, if that is even technically feasible(?). The reason: I'd like to put one or more 3d objects within an Fl_Scroll, but as I learned a while back, that cannot work with window-based classes, such as Fl_Gl_Window; sub-windows in the scroll container don't display right.

Issue #345, opened 2021-12-19
Fluid: improve numeric input fields in Widget Dialog


MatthiasWM:
The fields for x, y, w, and h should understand easy mathematical formulas. 

Let's say we want to move a widget 24 pixels down, and the current y coordinate is already displayed, the field should understand if we just append `+24` to the text, calculating the new y coordinate for us. If we select multiple widgets, entering `y+24` should move them all 24 pixels down. We can add a few more predefined values, like `py` for the parent, `sy` for the previous sibling, and `cy` for the first child.

This is meant for interactive design only and will not change the output to the source or header file.

PR #394, opened 2022-02-13
Better Fluid Widget dimension controls

MatthiasWM:
#345 : x, y, w, and height field in the Widget Dialog are now true numeric field. 

It is possible to enter simple formulas into any of the fields to have values calculated immediately. Let's say, the widget x position is `120`. To move the widget 5 units to the right, simply enter `120+5' and Fluid will calculate the new position.

The field support variables that come in really handy when moving multiple selected widgets, for example, entering `x+5` will move all selected widgets 5 units to the right.

Available variables are `x`, `y`, `w`, and `h` to get the dimensions of the current widget. `px` oct. will return the dimensions of the parent, `sx` etc. of the previous sibling, and `cx` etc. will return the bounding box of the children of each selected Widget. So setting a group to `cx-10`:`cy-10`:`cw+20`:`ch+20` will resize the group to contain all children with a ten pixel space around them.

The variable `i` starts at 0 and increments for every selected widget.

### DIFF
--- a/fluid/Fl_Type.cxx
+++ b/fluid/Fl_Type.cxx
@@ -299,6 +299,30 @@ Fl_Type::~Fl_Type() {
   if (comment_) free((void*)comment_);
 }
 
+// Return the previous sibling in the tree structure or NULL.
+Fl_Type *Fl_Type::prev_sibling() {
+  Fl_Type *n;
+  for (n = prev; n && n->level > level; n = n->prev) ;
+  return n;
+}
+
+// Return the next sibling in the tree structure or NULL.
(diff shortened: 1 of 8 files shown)


wcout:
Couldn't resist to test this 'calculator'. Works quite nicely. Division by 0 ( e.q. something/0 or something/something) is not trapped and crashes fluid.

MatthiasWM:
Thanks for testing. I tried division by zero on my Mac and it did not crash, so I didn't bother. I will fix that of course.

MatthiasWM:
Division by zero issue is fixed.

PR #397, opened 2022-02-20
DOC: Provide a link to the "FLTK standard box types" image

fire-eggs:
The doc for Box Types has the text "see figure 1" which does not exist! [It is figure 5.3 in PDF, no number in HTML].

This change provides a hyperlink which allows the reader to actually navigate to the cited figure.

Confirmed to work as desired in HTML; I don't have LATEX so cannot confirm the PDF.

### DIFF
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -573,6 +573,7 @@ enum Fl_When { // Fl_Widget::when():
     leaving the interior unchanged. The blue color in the image below
     is the area that is not drawn by the frame types.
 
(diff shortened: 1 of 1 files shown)


Albrecht-S:
@fire-eggs Hi Kevin, thanks for the patch. I verified that it works for the PDF docs as well.

However, I took the freedom to add the (partial) figure title "Standard Box Types" to the link rather than just "figure". I hope you don't mind.

fire-eggs:
No concerns on this end!

PR #415, opened 2022-03-20
Add utf8 encoding support in setup.cmake for MSVC compiling

ZJUGKC:
The source files are all encoded by UTF-8 without BOM. This will raise warnings and errors when compiling by the MSVC with some active code pages (ACP). The file `test/ask.cxx` contains such characters with UTF-8 encoding. Adding `/utf-8` here is equivalent to specifying `/source-charset:utf-8 /execution-charset:utf-8`, and all warnings and errors are removed.

### DIFF
--- a/CMake/setup.cmake
+++ b/CMake/setup.cmake
@@ -103,7 +103,7 @@ if (WIN32)
   if (MSVC)
     add_definitions (-DWIN32_LEAN_AND_MEAN)
     add_definitions (-D_CRT_SECURE_NO_WARNINGS)
-    add_definitions ("/utf-8")  # This is equivalent to specifying `/source-charset:utf-8 /execution-charset:utf-8`
+    add_compile_options (/utf-8)  # This is equivalent to specifying `/source-charset:utf-8 /execution-charset:utf-8`
     set (BORDER_WIDTH 2)
   endif (MSVC)
   if (CMAKE_C_COMPILER_ID STREQUAL GNU)
--- a/CMake/setup.cmake
+++ b/CMake/setup.cmake
@@ -103,6 +103,7 @@ if (WIN32)
   if (MSVC)
     add_definitions (-DWIN32_LEAN_AND_MEAN)
     add_definitions (-D_CRT_SECURE_NO_WARNINGS)
+    add_definitions ("/utf-8")  # This is equivalent to specifying `/source-charset:utf-8 /execution-charset:utf-8`
     set (BORDER_WIDTH 2)
   endif (MSVC)
   if (CMAKE_C_COMPILER_ID STREQUAL GNU)
--- a/FL/Fl_Image.H
+++ b/FL/Fl_Image.H
@@ -187,7 +187,7 @@ public:
     \see count(), w(), h(), data_w(), data_h(), d(), ld()
    */
   const char * const *data() const {return data_;}
-  int fail();
+  int fail() const;
   /**
     Releases an Fl_Image - the same as '\p delete \p this'.
 
--- a/src/Fl_Image.cxx
+++ b/src/Fl_Image.cxx
@@ -168,7 +168,7 @@ void Fl_Image::label(Fl_Menu_Item* m) {
  \return ERR_FILE_ACCESS if there was a file access related error (errno should be set)
  \return ERR_FORMAT if image decoding failed.
  */
-int Fl_Image::fail()
+int Fl_Image::fail() const
 {
     // if no image exists, ld_ may contain a simple error code
     if ( (w_<=0) || (h_<=0) || (d_<=0) ) {


Albrecht-S:
Thanks for this PR.

I never saw any character set (UTF-8) related warnings when compiling with MSVC. Can you explain how I would compile "with some active code pages (ACP)"? I would like to test this. TIA.

According to CMake documentation using "add_definitions()" could be used but the correct statement in this case would be [add_compile_options()](https://cmake.org/cmake/help/latest/command/add_compile_options.html#add-compile-options).

Can you please try if this works for you and modify the PR accordingly? Thanks.

ZJUGKC:
I use windows server 2019 with code page 936 (gb2312), simplified Chinese (PRC, Singapore). I use cmake command line to generate the VC project files, and compile it by `MSBuild`. 
1. This process raises some warnings. Some files contains characters which cannot be displayed in current code page, You should save them as UNICODE format files.
2. This process raises errors in line 39 and 42 of file `test/ask.cxx`. These two strings are encoded by UTF-8, and cannot be parsed correctly under code page 936. Compiler says that the unexpected return character is in constant string.

Albrecht-S:
@ZJUGKC Thank you for your further information. I conclude that I would have to change my Windows system settings (Codepage) to see this warning, is this correct?

You wrote:
> Some files contains characters which cannot be displayed in current code page, You should save them as UNICODE format files.

Sorry, this is not going to happen, for several reasons:
- all FLTK source files are encoded in UTF-8 without BOM
- UTF-8 is a valid Unicode encoding (UTF-8 **IS** Unicode)
- what Microsoft calls "UNICODE" is UTF-16, just another Unicode encoding
- we decided to use UTF-8 for all strings in the FLTK library and thus our source files are (and must be) in UTF-8 encoding as well.

That said, we appreciate all help from users. Thank you. Of course we're willing to help and solve issues if/when they arise. If your `CMake` setup proposal helps and doesn't break other builds, then this will accepted and implemented.

I asked you to try a small modification though (did you try, what are your results?) and I have another question: are you aware of a specific Visual Studio version that first implemented the compiler switch "`/utf-8`"? This would be important so we don't break compilation with older versions.

FTR: I found [in Microsoft docs](https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-140) that Visual Studio 2015 supports it but couldn't find anything about older versions.

Albrecht-S:
FWIW, here's another hint (meant as a workaround, not as a solution). I found [this answer on StackOverflow](https://stackoverflow.com/a/65945041/17213189):

"Visual Studio supports EditorConfig files (https://editorconfig.org/)".

It might not help with the compilation issue, but I wanted to mention it anyway. A simple `.editorconfig` file might look like this:
```
[*]
charset = utf-8
```

**_Disclaimer_**: I have no idea how this may help or not, I'm just referring to other resources.

ZJUGKC:
1. `/execution-charset`, `/source-charset`, and `/utf-8` are supported since Visual Studio 2015 update 3.
2. I do not use Visual Studio IDE. I write BAT script to build FLTK in order to execute CI/CD. So revising `.editorconfig` file makes no sense to me.
3. I replace `add_definitions ("/utf8")` with `add_compile_options (/utf8)`, and it also works fine.

Albrecht-S:
@ZJUGKC Great, thanks.
> ... and `/utf-8` are supported since Visual Studio 2015 update 3.

Then I believe in `CMake` this would correspond to `MSVC_VERSION = 1900` according to [Microsoft docs](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros) and [CMake docs](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html). There doesn't seem to be an indicator for "update 3" though. I'll check this further and take care of this.

> ... revising .editorconfig file makes no sense to me

OK, that was only meant as a hint if someone uses it.

Thanks for the update with `add_compile_options (/utf8)` and the confirmation that it works.

I'll do my own tests and if it works I'll merge it. Thanks again.

MatthiasWM:
Instead of changing the build system, wouldn't;t ift be more sensible to change the ask.cxx source code in lines 39 and 42 to not use the Euro or Paragraph but instead use the backslash notation. It's kind of funny that are still avoiding RTTI, templates, and the C++ std library to avoid incompatibility with ancient systems, yet we use UTF-8 in our source code.

Albrecht-S:
@MatthiasWM wrote:
> Instead of changing the build system, wouldn't it be more sensible to change the ask.cxx source code ...

1. This particular code in `test/ask.cxx` was intended to show how you can set an icon with UTF-8 text that is more than on byte but only one character. Sure, we could use `\xnn` escape sequences but this would be harder to read (although the number of bytes would be obvious).
2. I don't think that these are the only places in the sources where we use plain UTF-8 (non-ASCII) text in source files. I don't know why the OP didn't mention other warnings/errors but maybe these other occurrences are in platform specific (non-Windows) code or in (doxygen) comments. Unfortunately I didn't find examples quickly enough.
3. I would prefer to be open for all sorts of UTF-8 encoded strings in the source files in the future.
4. Microsoft is often an exception. ;-)
5. Maybe there are even more reasons...

That said, I would prefer to add the option rather than tweaking source files.

ZJUGKC:
B.T.W, `const` filter should be added to Fl_Image::fail(), because this method does not modify its member data.

Albrecht-S:
@ZJUGKC This PR has been merged although I edited the commits.

Thanks for adding the Fl_Image::fail() 'const' although this would have better been addes as another PR or issue. Anyway, I committed both the Fl_Image changes and the MSVC / CMake stuff.

I also added the check for VS 2015 and later. We can't check for individual updates though, i.e. users with VS 2015 need to update their version to update 3 and VS versions before 2015 don't get this compiler switch.

Note that I can no longer "officially" merge this PR (by means of the GitHub Merge button), hence I'm closing it now. Thanks for your contribution.

PR #420, opened 2022-03-23
Remove vsscanf_l() call from Win32 driver (it appears to be a BSD-ism…

imaclmaca:
… and unsupported by the MS tools) and rename several clocale specific methods from "name" to "vname" since they take a va_list not a variable list of arguments.


Albrecht-S:
Thanks, Ian. Code looks good and builds fine on Linux, macOS, and Windows (Linux/MinGW cross-compiler), hence I merged it.

The fact that `vsscanf()` is not available in pre-2013 Visual Studio is not solved though (as discussed in fltk.coredev) and should be addressed separately.

PR #424, opened 2022-04-04
Add Zenity-based file chooser based on the KDialog one.

HugLifeTiZ:
If available, it is used on Linux regardless of the current desktop, outprioritizing even the Kdialog backend, because it offers free XDG portal integration, which means it picks the correct file chooser on all desktops, and allows for meaningful file selection in sandbox environments like Flatpak.

I tested this by building Obsidian against this branch as a Flatpak, using a manifest that I'm currently working on. I removed the `--filesystem=home` permission I had been using, and asked it to build me a WAD. It did successfully bring up the KDE file chooser, and I was able to successfully build and save a WAD to a location outside of the sandbox! But it laid bare the fact that a lot of stuff in the Kdialog chooser was simply not implemented, and thus, they're not implemented in this Zenity chooser, either.

There are a lot of things the Kdialog chooser doesn't do even though Kdialog can do them, and thus, same for Zenity. For example, you can supply an initial filename, as well as filters. The Kdialog backend simply doesn't do that, and thus neither does the Zenity backend. Obsidian does supply a procedurally generaated filename suggestion to the file chooser. It's entirely absent in this backend.

I have to admit that I am pretty far out of my depth in this C++ code. I'm not even sure how it's separating filenames from multiple selections so that I can correct the handling; Zenity's default filename separator is a pipe character. You can change it on the command line, but I figure it's better to do that in this backend.

The Zenity backend does exist though, now. My branch currently uses it by default, but maybe the default should be not to use it until it implements everything that the Zenity chooser can do. Unfortunately, I don't think I'll be able to do more than this. :(

### DIFF
--- a/src/Fl_Native_File_Chooser_GTK.cxx
+++ b/src/Fl_Native_File_Chooser_GTK.cxx
@@ -943,7 +943,7 @@ Fl_Native_File_Chooser::Fl_Native_File_Chooser(int val) {
         }
         Fl_Zenity_Native_File_Chooser_Driver::have_looked_for_zenity = true;
       }
-      // if we found kdialog, we will use the Fl_Kdialog_Native_File_Chooser_Driver
+      // if we found zenity, we will use the Fl_Zenity_Native_File_Chooser_Driver
       if (Fl_Zenity_Native_File_Chooser_Driver::did_find_zenity) platform_fnfc = new Fl_Zenity_Native_File_Chooser_Driver(val);
     }
 #endif // USE_ZENITY
--- a/src/Fl_Native_File_Chooser_GTK.cxx
+++ b/src/Fl_Native_File_Chooser_GTK.cxx
@@ -17,6 +17,9 @@
 
 #include <config.h>
 #include <FL/Fl_Native_File_Chooser.H>
+#if USE_ZENITY
+#  include "Fl_Native_File_Chooser_Zenity.H"
+#endif
 #if USE_KDIALOG
 #  include "Fl_Native_File_Chooser_Kdialog.H"
 #endif
@@ -920,14 +923,33 @@ void Fl_GTK_Native_File_Chooser_Driver::probe_for_GTK_libs(void) {
 #endif // HAVE_DLSYM && HAVE_DLFCN_H
 
 Fl_Native_File_Chooser::Fl_Native_File_Chooser(int val) {
-  // Use kdialog if available at run-time and if using the KDE desktop,
-  // else, use GTK dialog if available at run-time
+  // Use zenity if available at run-time even if using the KDE desktop,
+  // because its portal integration means the KDE chooser will be used.
+  // Else use kdialog if available at run-time and if using the KDE
+  // desktop, else, use GTK dialog if available at run-time
   // otherwise, use FLTK file chooser.
   platform_fnfc = NULL;
   if (Fl::option(Fl::OPTION_FNFC_USES_GTK)) {
+#if USE_ZENITY
+    if (val != BROWSE_MULTI_DIRECTORY) {
+      if (!Fl_Zenity_Native_File_Chooser_Driver::have_looked_for_zenity) {
+        // First Time here, try to find zenity
+        FILE *pipe = popen("zenity --version 2> /dev/null", "r");
+        if (pipe) {
+          char *p, line[100] = "";
+          p = fgets(line, sizeof(line), pipe);
+          if (p && strlen(line) > 0) Fl_Zenity_Native_File_Chooser_Driver::did_find_zenity = true;
+          pclose(pipe);
+        }
+        Fl_Zenity_Native_File_Chooser_Driver::have_looked_for_zenity = true;
+      }
+      // if we found kdialog, we will use the Fl_Kdialog_Native_File_Chooser_Driver
+      if (Fl_Zenity_Native_File_Chooser_Driver::did_find_zenity) platform_fnfc = new Fl_Zenity_Native_File_Chooser_Driver(val);
+    }
+#endif // USE_ZENITY
 #if USE_KDIALOG
     const char *desktop = getenv("XDG_CURRENT_DESKTOP");
-    if (desktop && strcmp(desktop, "KDE") == 0 && val != BROWSE_MULTI_DIRECTORY) {
+    if (!platform_fnfc && desktop && strcmp(desktop, "KDE") == 0 && val != BROWSE_MULTI_DIRECTORY) {
       if (!Fl_Kdialog_Native_File_Chooser_Driver::have_looked_for_kdialog) {
         // First Time here, try to find kdialog
         FILE *pipe = popen("kdialog -v 2> /dev/null", "r");
--- a/src/Fl_Native_File_Chooser_Zenity.cxx
+++ b/src/Fl_Native_File_Chooser_Zenity.cxx
@@ -108,7 +108,7 @@ int Fl_Zenity_Native_File_Chooser_Driver::show() {
     default:
       option = "--file-selection";
   }
-  char *preset = (char*)"";
+  char *preset = NULL;
   if (_preset_file) {
     preset = new char[strlen(_preset_file) + 15];
     sprintf(preset, "--filename '%s'", _preset_file);
@@ -124,7 +124,8 @@ int Fl_Zenity_Native_File_Chooser_Driver::show() {
   if (_title) {
     sprintf(command+strlen(command), " --title '%s'", _title);
   }
-  sprintf(command+strlen(command), " %s %s ", option, preset);
+  sprintf(command+strlen(command), " %s %s ", option, preset ? preset : "");
+  delete[] preset;
   if (_parsedfilt) sprintf(command+strlen(command), " \"%s\" ", _parsedfilt);
   strcat(command, "2> /dev/null"); // get rid of stderr output
 //puts(command);
--- a/src/Fl_Native_File_Chooser_Zenity.cxx
+++ b/src/Fl_Native_File_Chooser_Zenity.cxx
@@ -108,9 +108,16 @@ int Fl_Zenity_Native_File_Chooser_Driver::show() {
     default:
       option = "--file-selection";
   }
-  const char *preset = ".";
-  if (_preset_file) preset = _preset_file;
-  else if (_directory) preset = _directory;
+  char *preset = (char*)"";
+  if (_preset_file) {
+    preset = new char[strlen(_preset_file) + 15];
+    sprintf(preset, "--filename '%s'", _preset_file);
+  }
+  else if (_directory) {
+    // This doesn't actually seem to do anything, but supply it anyway.
+    preset = new char[strlen(_directory) + 15];
+    sprintf(preset, "--filename '%s'", _directory);
+  }
   char *command = new char[strlen(option) + strlen(preset) + (_title?strlen(_title)+11:0) +
                            (_parsedfilt?strlen(_parsedfilt):0) + 50];
   strcpy(command, "zenity ");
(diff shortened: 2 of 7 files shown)


HugLifeTiZ:
Okay, so I at least have some guidance as for what to do next. It turns out that the kdialog chooser *does* support an initial filename and filters. It looks like the initial filename and filters are arguments to the sole command line argument that controls that mode. I took a look at the command line that gets invoked when Obsidian tries to generate a WAD:

```
zenity  --title 'Select output file' --file-selection --save 2022-04-04-1153_anathema_at_the_wildfire  "WAD files (*.wad)\nAll files (*)" 2> /dev/null
```

What it *should* be is:

```
zenity  --title 'Select output file' --file-selection --save --filename 2022-04-04-1153_anathema_at_the_wildfire --file-filter "WAD files|*.wad" --file-filter "All files|*"
```

So the initial filename is given as an additional command line switch, and filters work the same way. It takes an arbitrary number of filters, and rather than separating them via a newline escape in the string, you use additional invocations of the switch. Rather than using it of the form `name (pattern)`, it's of the form `name|pattern`.

I *might* be able to get it to supply the initial filename correctly; I just have to correctly figure out how to prepend `--filename ` to the option with intimidating C-style string handling. 😅

But I am *absolutely* way out of my depth when it comes to figuring out how to accept the filters correctly. And I still don't know what in the world it's doing to separate multiple files in order to have it split by the pipe character.

HugLifeTiZ:
Okay, so I've got it at least taking a preset filename correctly. Filters working correctly would be nice, but they're not strictly needed for Obsidian. And again; out of my depth on that.

src/Fl_Native_File_Chooser_GTK.cxx
```diff
@@ -920,14 +923,33 @@ void Fl_GTK_Native_File_Chooser_Driver::probe_for_GTK_libs(void) {
...
+        if (pipe) {
+          char *p, line[100] = "";
+          p = fgets(line, sizeof(line), pipe);
+          if (p && strlen(line) > 0) Fl_Zenity_Native_File_Chooser_Driver::did_find_zenity = true;
+          pclose(pipe);
+        }
+        Fl_Zenity_Native_File_Chooser_Driver::have_looked_for_zenity = true;
+      }
+      // if we found kdialog, we will use the Fl_Kdialog_Native_File_Chooser_Driver
+      if (Fl_Zenity_Native_File_Chooser_Driver::did_find_zenity) platform_fnfc = new Fl_Zenity_Native_File_Chooser_Driver(val);
```
Albrecht-S, line 947: The comment above this line mentions kdialog rather than zenity. Likely a copy/paste error?

src/Fl_Native_File_Chooser_Zenity.cxx
```diff
@@ -0,0 +1,284 @@
...
+    case Fl_Native_File_Chooser::BROWSE_MULTI_FILE:
+      option = "--file-selection --multiple";
+      break;
+
+    default:
+      option = "--file-selection";
+  }
+  char *preset = (char*)"";
+  if (_preset_file) {
+    preset = new char[strlen(_preset_file) + 15];
```
Albrecht-S, line 113: This and the cast above looks like causing a memory leak. Perhaps the initialization should be
```
  char *preset = 0; // or NULL
```
and preset should be delete[]'ed somewhere (later).

Albrecht-S:
Thank you very much for this PR. The code looks good (well formatted) and I added two comments to particular issues I found. However, my action can't be called a code review, I just skimmed over the changes and flagged these two issues.

Reviewing the code (i.e. the semantics) and all the issues and potential add-on's you mentioned is way out of my comfort zone. I might give it a try by testing (I have zenity installed) but please don't hold your breath.
```
$ zenity --version
3.32.0
```
 Maybe (and hopefully) someone else will take a look at this.

Albrecht-S:
BTW, what is a "WAD"? I found something searching the net, but it would be nice if you told us what this means.

erco77:
> BTW, what is a "WAD"? 

WAD is an old game data archive format (similar to zip, tar) made famous by the game "DOOM". You could edit DOOM WAD files to, for example, change the game's characters into other things, and save it back so you could customize the game play. The format has endured apparently in modern games..
https://fileinfo.com/extension/wad

src/Fl_Native_File_Chooser_Zenity.cxx
HugLifeTiZ: Will initializing it like that still work for the sprintf at line 127? That's where it ultimately gets used, so I think right after that is a good place to delete it.

HugLifeTiZ:
> BTW, what is a "WAD"? I found something searching the net, but it would be nice if you told us what this means.

It doesn't really have anything to do with this *specifically*; my testbed for these changes is a Flatpak build of [Obsidian](https://github.com/dashodanger/Obsidian), which is a level generator for Classic Doom. Content for Classic Doom is stored in the WAD (Where's All the Data?) format. So in the context of this PR, using Obsidian to generate a WAD means bringing up the file chooser and telling it where to save the WAD and what it should be called.

src/Fl_Native_File_Chooser_GTK.cxx
HugLifeTiZ: Yup, that's exactly what it is. My bad! I pushed a commit to fix it.

src/Fl_Native_File_Chooser_Zenity.cxx
```diff
@@ -0,0 +1,284 @@
...
+    preset = new char[strlen(_directory) + 15];
+    sprintf(preset, "--filename '%s'", _directory);
+  }
+  char *command = new char[strlen(option) + strlen(preset) + (_title?strlen(_title)+11:0) +
+                           (_parsedfilt?strlen(_parsedfilt):0) + 50];
+  strcpy(command, "zenity ");
+  if (_title) {
+    sprintf(command+strlen(command), " --title '%s'", _title);
+  }
+  sprintf(command+strlen(command), " %s %s ", option, preset);
```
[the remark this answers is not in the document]
HugLifeTiZ: I don't know why you're sure I know that, I have repeatedly admitted to being a baby when it comes to C++.

...I do actually know that, though. :P

That sounds good to me; I've updated the code accordingly and pushed another commit. Thank you!

HugLifeTiZ:
@ManoloFLTK, you're the original author of the Kdialog backend, right? Do you think you'd be able to help with changing the rest of this adapted backend for Zenity?

src/Fl_Native_File_Chooser_Zenity.cxx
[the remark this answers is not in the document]
Albrecht-S: Commit looks good, thanks.

ManoloFLTK:
> @ManoloFLTK, you're the original author of the Kdialog backend, right? Do you think you'd be able to help with changing the rest of this adapted backend for Zenity?

Could you, please, summarize what can't be done with kdialog but could by zenity?
What happens with kdialog for a sandboxed application?
Would it make sense to replace kdialog by zenity rather than add zenity and keep kdialog?

HugLifeTiZ:
Those are great questions!

> summarize what can't be done with kdialog but could by zenity?

Kdialog does not offer any sort of portal integration, so you would not get the desktop's native file chooser while invoking it, and it would do nothing to help the app work in a sandbox. Zenity uses the xdg desktop portal though, so any invocation of it, inside or outside a sandbox, will use the desktop's native file chooser. It seems like the same limitations exist, such as that the "multi-directory" mode can't be done meaningfully; you can't pick more than one directory if you combine `--directory --multiple`.

> What happens with kdialog for a sandboxed application?

Kdialog does not exist in any of the Flatpak runtimes, not even the official KDE one. But Zenity *does* exist even in the KDE runtime. The funny thing though is that it does *not* exist in the GNOME runtime, because GNOME expects developers to write "real" code instead. That's okay, though; an app using FLTK would not target the GNOME runtime. So to answer the question: nothing happens, because it doesn't exist.

> Would it make sense to replace kdialog by zenity rather than add zenity and keep kdialog?

It *almost* makes sense, but not quite, and the main reason why is that Zenity is not pre-installed in KDE Neon, and likely not pre-installed in other KDE-first distributions as well. This affects non-sandboxed apps, which is still most of them. So if Kdialog is removed from FLTK, users of non-sandboxed FLTK apps on Plasma would get the GTK dialogs again.

ManoloFLTK:
@HugLifeTiZ @Albrecht-S 
I've extended the proposed zenity-based file chooser in my fltk fork https://github.com/ManoloFLTK/fltk branch `zenity-file-chooser` and am considering whether to commit it to the main branch. I've much reduced its size by deriving the new class (`Fl_Zenity_Native_File_Chooser_Driver`) from an already existing class (`Fl_Kdialog_Native_File_Chooser_Driver`).
My concern is that this chooser takes priority over the GTK chooser, and so FLTK would loose two features with zenity
1) The `hide/show hidden files` button. There's however a workaround, to type ctrl-H which switches between showing and hiding .* files. This supposes the user is aware of that.
2) The `preview` button.

Any comment about this concern ?

Albrecht-S:
My first comments w/o testing, regarding...
- new zenity file chooser in general: I would need to test it (which I'll do later).
- missing features (hidden files and preview): this would be bad. Many users would likely not be aware of `^h`. I had the same issue on macOS. More about this after my tests.
- priority 1: would it make sense to add yet another build option?
- priority 2: would it make sense (and be possible) to add a runtime `Fl::option` ?

ManoloFLTK:
@Albrecht-S 

I have just added a "Show hidden files" button to the macOS native file chooser (2fd1866).

> would it make sense to add yet another build option?

Yes it would, but that adds to already many options. It's already a CMake option in the OP's commit under name  OPTION_USE_ZENITY.

> would it make sense (and be possible) to add a runtime Fl::option ?

That's probably my preference. I don't see obstable to do that.

Albrecht-S:
> > would it make sense (and be possible) to add a runtime Fl::option ?
> That's probably my preference. I don't see obstacle to do that.

Well, it would only make sense if we could build more than one file chooser in the FLTK library and not select one of them by build options. Then users could really choose the one they like, or a system manager can set one in the system FLTK preferences.

Currently we have already one option (`"Native File Chooser uses GTK"`) in the "FLTK Preferences" settings (currently in fluid). This is an `on/off/default` setting. You know this file chooser stuff better, what would you suggest as potential options, considering whatever we have now (GTK, KDE, Zenity, ...?). I believe it's now becoming a little confusing.

ManoloFLTK:
Commit d905162 to the `zenity-file-chooser` branch of my FLTK fork is my proposal for a complete implementation. It adds a dedicated FL:option() :

options

and documents in a detailed way how the 4 possible file chooser styles are controlled:

doc

ManoloFLTK:
> missing features (hidden files and preview): this would be bad. 

Concerning the `preview` issue: it may be not completely bad because the `zenity` dialog shows an icon for each file which does preview image files, though with a very small size.

MatthiasWM:
Which reminds me that I still have a branch sitting here that takes the options out of FLUID into an independent app... . No worries, Manolo, I will port the new option over.

ManoloFLTK:
@HugLifeTiZ Many thanks for your Pull Request. It has been expanded and then pushed to the master branch of the FLTK repo (commits 576271f and 41f3761). I will therefore close this PR.

@Albrecht-S Thanks for your review. The new code is now pushed to master.

HugLifeTiZ:
Thank you all for picking up and expanding upon this work. This will be very helpful for packaging FLTK apps in Flatpak, and improving platform cohesion. I'm looking forward to restarting my work on Obsidian's Flathub packaging to make use of this. :)

PR #432, opened 2022-04-14
Check for child visibility in Fl_Scroll.

DieSkaarj:
Hi there,

I'd like to suggest the inclusion of a check for child visibility when calculating the scroll area in a Fl_Scroll widget. This means the total area scrollable will be according to shown widgets only.

Thanks!
David

### DIFF
--- a/src/Fl_Scroll.cxx
+++ b/src/Fl_Scroll.cxx
@@ -147,7 +147,7 @@ void Fl_Scroll::recalc_scrollbars(ScrollInfo &si) const {
   Fl_Widget*const* a = array();
   for (int i=children(); i--;) {
     Fl_Widget* o = *a++;
-    if ( o==&scrollbar || o==&hscrollbar ) continue;
+    if ( o==&scrollbar || o==&hscrollbar || o->visible==0 ) continue;
     if ( first ) {
         first = 0;
         si.child.l = o->x();
--- a/src/Fl_Scroll.cxx
+++ b/src/Fl_Scroll.cxx
@@ -147,7 +147,7 @@ void Fl_Scroll::recalc_scrollbars(ScrollInfo &si) const {
   Fl_Widget*const* a = array();
   for (int i=children(); i--;) {
     Fl_Widget* o = *a++;
-    if ( o==&scrollbar || o==&hscrollbar || o->visible==0 ) continue;
+    if ( o==&scrollbar || o==&hscrollbar || o->visible()==0 ) continue;
     if ( first ) {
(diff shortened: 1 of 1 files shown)


DieSkaarj:
If the inner bounding box is required to be a consistent size a box can be used in a way similar to how resizing works.

Albrecht-S:
Hi @DieSkaarj, thank you for your proposal, and I apologize for the late reply. You wrote:
> If the inner bounding box is required to be a consistent size a FL_Box can be used in a way similar to how resizing works.

It is well documented that the `resizable()` widget of an `Fl_Group` is allowed to be an **_invisible box_** (https://www.fltk.org/doc-1.4/classFl__Group.html#afd220e45e1ac817bde7d6a25fdf74e37):
_"It is possible to achieve any type of resize behavior by using an invisible [Fl_Box](https://www.fltk.org/doc-1.4/classFl__Box.html) as the resizable ..."_

Although this is not exactly the same as adding an invisible box to an `Fl_Scroll` group this is well known to work and has IIRC been suggested in some discussions in our user support group [fltk.general](https://groups.google.com/g/fltkgeneral).

Your suggestion would break programs that rely on this behavior and can therefore not be accepted. I'm sorry to say that.

DieSkaarj:
Hi Albrecht-S,

Thank you for your response.

Fair enough.

Kind regards,
David

Albrecht-S:
Hi @DieSkaarj, after thinking about your proposal for a little longer my own feeling is that this PR is a legitimate attempt to improve FLTK. It doesn't make much sense to calculate the bounding box including invisible widgets. If there was not the backwards compatibility aspect.

However, I'm now believing that we should consider (not yet: apply) your PR regarding the distinction of "hidden" versus "invisible" widgets. I'm going to open a thread in our fltk.coredev developer group to ask for comments. There's still hope that we can apply your PR. I'll post a link in a follow-up message...

Albrecht-S:
OK, it took a while to post my message to fltk.coredev in thread '[RFC: "invisible" versus "hidden"?](https://groups.google.com/g/fltkcoredev/c/f4evVu37O80/m/IqQU0XFOAAAJ)'. Let's wait and see what other FLTK devs think about it.

Albrecht-S:
@DieSkaarj Good news: we got 3 of 3 positive votes, hence I'm going to merge your PR. Thanks for finding this and for providing the patch.

DieSkaarj:
@Albrecht-S You, and the team, are very welcome. And, thank you for investigating it further.

PR #440, opened 2022-05-05
Fix kdialog native picker not properly handling spaces in filenames

mtolly:
After upgrading fltk I noticed that it uses the recently added `kdialog` native file picker. However it does not correctly return filenames that contain a space anywhere in the path, instead splitting them into multiple results.

This fixes it by using kdialog's `--separate-output` which modifies `--multiple` to separate multiple results with newlines instead of the default spaces.

I verified that this fixes the `native-filechooser.cxx` test app, and also tested modifying it to use `BROWSE_MULTI_FILE`. Both single and multi file select now behave correctly when selecting filenames with spaces.


ManoloFLTK:
Thanks for your contribution.

PR #455, opened 2022-06-25
Fix unicode support in scandir_win32.c by using GetFileAttributesW

ZJUGKC:
There is a GetFileAttributes calling in scandir_win32.c which does not support UTF-8 encoding file name. I modified it using GetFileAttributesW.

### DIFF
--- a/src/scandir_win32.c
+++ b/src/scandir_win32.c
@@ -89,7 +89,6 @@ int fl_scandir(const char *dirname, struct dirent ***namelist,
   strcpy(findIn, dirname);
 
   /* #if defined(__GNUC__) */
-  /* #warning FIXME This probably needs to be MORE UTF8 aware now */
   /* #endif */
   for (d = findIn; *d; d++) if (*d=='/') *d='\\';
   if (len==0) { strcpy(findIn, ".\\*"); }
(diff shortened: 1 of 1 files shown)


src/scandir_win32.c
```diff
@@ -89,28 +89,28 @@ int fl_scandir(const char *dirname, struct dirent ***namelist,
   strcpy(findIn, dirname);
 
   /* #if defined(__GNUC__) */
```
Albrecht-S, line 91: Please remove the empty comment as well. Thanks.

Albrecht-S (changes requested):

Albrecht-S:
Thanks for the PR. The changes look correct, but please see my comment above.

Note that you can push another commit to do what I requested. The final merge will be squashed (have both commits combined). TIA.

src/scandir_win32.c
Albrecht-S: Thanks for removing the comments.

Albrecht-S:
Thanks for updating the PR. I'll look further into it and try to test it locally before I merge (unless someone else takes it).

Albrecht-S:
Merged. Thank you for your contribution.

FTR: I tested it with several Unicode file names like "äöü.txt" (German Umlauts). Some of these files/directories were created in a DOS terminal, others using MinGW. The results were all valid UTF-8 strings.

PR #458, opened 2022-07-02
Change colors of message box in fl_ask.H

ZJUGKC:
Add two functions `fl_message_colors` and `fl_message_button_colors` in fl_ask.H to change colors of message dialogs.

### DIFF
--- a/FL/fl_ask.H
+++ b/FL/fl_ask.H
@@ -78,6 +78,20 @@ inline void fl_message_font(Fl_Font f, Fl_Fontsize s) {
   fl_message_size_ = s;
 }
 
+extern FL_EXPORT Fl_Color fl_message_color_;
+extern FL_EXPORT Fl_Color fl_message_labelcolor_;
+inline void fl_message_colors(Fl_Color b, Fl_Color l) {
+  fl_message_color_ = b;
+  fl_message_labelcolor_ = l;
+}
+
+extern FL_EXPORT Fl_Color fl_message_button_color_;
+extern FL_EXPORT Fl_Color fl_message_button_labelcolor_;
+inline void fl_message_button_colors(Fl_Color b, Fl_Color l) {
+  fl_message_button_color_ = b;
+  fl_message_button_labelcolor_ = l;
+}
+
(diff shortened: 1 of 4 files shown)


Albrecht-S:
@ZJUGKC Thank you for this PR. Although the proposed new functions look good at a first glance, there are some mostly technical details that should be changed.

1. First of all there's a technical (Git, GitHub) aspect. Please update your fork to the latest `master` branch from [fltk/fltk](https://github.com/fltk/fltk) and then open a new, clean **branch** on top of `master` for your proposal. Never add your commits on `master` in your fork which might be outdated soon and make things harder for the maintainers (see your 4 commits on master which include a merge). A new clean branch can always be rebased and should contain only commits of one topic (PR). Thank you.

2. Although your additions are the same style as the existing code they introduce new global variables that can be accessed directly by the user. This old (unfortunately necessary, because backwards compatible) FLTK 1.3 style should no longer be followed. Rather than adding new global variables to the public `fl_message_*` name space new variables should be hidden in the new class `Fl_Message` as `static` class variables with accessor methods like `fl_message_something()` as you proposed. This includes setters but could also need getter methods because the global variables would be hidden and otherwise unaccessible.

3. Many (probably most) of the existing functions that change the behavior of these common dialogs are "one shot functions" that set a particular aspect to be used only for the next call of one of these dialogs and that aspect would be reset to standard once the next dialog was started. Your supposed change would change the behavior of **all** following dialogs unless changed by the user again to another value.

Point 3 is worth a discussion which should be done in our forum (mailing list) [fltk.coredev](https://groups.google.com/g/fltkcoredev) rather than in the context of a PR. But for a start: I wouldn't mind allowing the functions proposed by you to change the global appearance of all dialog functions for the entire runtime of the application - if this is what you intended.

If you still like to propose your changes - maybe modified after discussion in fltk.coredev  - then please open a new PR taking points 1. and 2. above into account. Thank you in advance.

And BTW, thank you for all your constructive reports as issues and PR's. All contributions are appreciated.

Albrecht-S (changes requested):
There are some required changes, please see "Conversation" comments.

PR #463, opened 2022-07-08
enhancement: Change colors and fonts of message dialogs in fl_ask.H

ZJUGKC:
Now I add these `fl_message_*` functions for changing colors and fonts of common dialogs by using static member variables in `Fl_Message`.

### DIFF
--- a/src/Fl_Message.cxx
+++ b/src/Fl_Message.cxx
@@ -70,6 +70,19 @@ const char *Fl_Message::message_icon_label_;
 
 Fl_Box *Fl_Message::message_icon_;
 
+Fl_Color Fl_Message::message_color_default_ = FL_GRAY;
+Fl_Color Fl_Message::message_color_ = FL_GRAY;
+Fl_Color Fl_Message::message_labelcolor_default_ = FL_FOREGROUND_COLOR;
+Fl_Color Fl_Message::message_labelcolor_ = FL_FOREGROUND_COLOR;
+Fl_Color Fl_Message::message_button_color_default_ = FL_GRAY;
+Fl_Color Fl_Message::message_button_color_ = FL_GRAY;
+Fl_Color Fl_Message::message_button_labelcolor_default_ = FL_FOREGROUND_COLOR;
+Fl_Color Fl_Message::message_button_labelcolor_ = FL_FOREGROUND_COLOR;
+Fl_Font Fl_Message::message_button_font_default_ = FL_HELVETICA;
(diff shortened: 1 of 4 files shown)


src/Fl_Message.cxx
```diff
@@ -327,6 +340,25 @@ int Fl_Message::innards(const char *fmt, va_list ap, const char *b0, const char
...
     message_->labelsize(fl_message_size_);
 
+  window_->color(message_color_);
+  message_color(message_color_default_); // reset
+  message_->labelcolor(message_labelcolor_);
+  message_labelcolor(message_labelcolor_default_); // reset
+
+  for (int b = 2; b >= 0; b--) {
+    button_[b]->color(message_button_color_);
+    message_button_color(message_button_color_default_); // reset
```
Albrecht-S, line 350: This seems to reset the button color to the default value inside the loop which implies that only one button color would ever be changed. This and all the following "reset" statements should IMHO be outside (below/after) the loop.
Note: this is not a full review, it's just something I stumbled across.

ZJUGKC:
Yes, colors of buttons will be changed to the same value because the number of buttons in different kind of standard dialogs may be 1, 2, or 3.

You are right, the statement of color resetting for dialog buttons can be removed from loop.

MatthiasWM:
I am not a fan of this change. We are adding quite a lot of code to customise the simplest dialog in the book. If we apply this patch, we will get another PR to customise box types, allow more buttons, have a backdrop image, and animated gifs as graphics. This is never-ending.

I see `fl_message` as a quick and minimal (or fast and light) way to inform the user. By setting the background color globally, the message box will already adapt. For everything else, we have Fluid or a few lines of code to generate dialog boxes of any complexity.

Albrecht-S:
I agree that we don't need yet another method to customize these standard dialogs. As @MatthiasWM wrote, there would be even more options to change.

One more argument for not applying this PR: I would like to create a building block (class) for user dialogs that could be used and customized by programmers to make their own dialogs. This could for instance be class `Fl_Dialog` which would provide the basic dialog features like popping up a dialog window and returning its "value" (whatever this is, text, int, float) to the user program. Such `Fl_Dialog` based dialogs should be easily customizable and extensible by user programs (adding widgets etc.). The future `Fl_Dialog` class would likely be based on `Fl_Grid` to simplify (dynamic) layouting. Developers could also subclass it for special dialogs. That's the plan...

MatthiasWM:
Solution: please create your own dialog boxes with the additional features, or check if the soon to be released `Fl_Dialog` class will fit your needs.

PR #468, opened 2022-07-17
update readme

ngdream:


ngdream:
i seriously think you should make a better readme

ManoloFLTK:
I'm not convinced the .md format would improve several of the existing README files, and I like the raw text format's advantage of universality.
But I'm not against switching to the .md format is that's what a majority supports.

Albrecht-S:
Here are my thoughts about the future of our README files:
1. My plan is to strip the README files down to a bare minimum. There are way too many README files, partially with overlapping topics like CMake which applies to all platforms. This should be "simplified" and README files should be just for the first steps with links to the official documentation (see points 3++ below).
2. I would keep the .txt file type and bare UTF-8 text because it is readable everywhere. Markup for these files is IMHO not necessary.
3. To support point 1. above I would like to restructure the official docs a "tiny bit" and add one or more better structured chapter(s) about building the FLTK library and about building FLTK applications. These new chapters would contain most of the current contents of our README files.
4. To make point 3. work the new "`Building FLTK`" chapter would have to be restricted to the "principles" of building for any platform with `CMake` and another chapter/section/paragraph on building with "old-style" `configure` and `make`. The latter would only be for Unix/Linux platforms and, as far as it applies, for MinGW and MSYS2 which use a POSIX shell as well.
5. The new "`Building FLTK Applications`" chapter would concentrate on `CMake` usage and a short paragraph on using `fltk-config` in `Makefiles` and similar.
6. We should NOT strive to write instructions for all Unix/Linux flavors because these are "moving targets".
7. We can IMHO also not document and support all CMake usage variants. It should be enough to show how a simple (1 source file) or medium (2-5 source files) program can be built. Software developers writing more complex programs will hopefully be able to use `CMake` or their own tools based on the instructions given in these chapters/paragraphs.

This should be all we need (improve the documentation) and then we can simplify and/or remove many of the existing README files.

PS: I know this is more or less OT here but I wanted to write it down at least once here. Sorry for the long comment.

erco77:
On 11/4/22 05:58, Matthias Melcher wrote:

> I agree that the README files need to be consolidated and refreshed.
>
     +1 - definitely.

     I know Albrecht has wanted to work on this but is probably
     overloaded, so we could all pitch in I reckon.

     Albrecht: if you have ideas for how to pair down the number
     files or ideas on where to consolidate, let us know and we can
     perhaps all help work in that direction.


     I've made a few additions for the Windows command line builds,
     as that's the one I always forget and need to refer to an example
     or I'll lose hours trying to remember the incantations.

> Are to other Devs OK with renaming .txt files to .md files,
> so that we can use GitHub markdown while still remaining readable in the console?
>

     -1 for making the README files .md's.

     Suggest we stick with simple text and .txt extensions and follow a
     standard format that includes an index at the beginning with
     section / subsection numbering for large documents.

     I don't think the primary audience reading them will be github users,
     or won't have a proper way to view .md files (no viewer/browser).

     I'd prefer html over md, simply because people could click on the documents
     to view it easily; browsers are pretty standard OS tools across all platforms.
     But HTML is hard to read from the command line (e.g. with 'more'), so -1
     for that too.

    .md viewers/editors are not standard OS tools and are (still?) kinda rare
     and hard to find, so clicking on them won't easily view them. For instance
     clicking them from a web browser won't open it properly/easily.

erco77:
It would be good if there was a simple build instructions document for how to build FLTK in all its possible ways, and have that either intro into ALL techniques (configure, cmake, platforms etc) with simple common examples for each, and perhaps refer to separate documents for more advanced/uncommon details (e.g. debug flags, cross compiling, etc)

This way the common user just wanting to simply build FLTK can do so by viewing the BUILD document.

So maybe:
```
README                       -- top level. Short file: refers to other docs
  |
  |__ BUILD                  -- examples for all, with only common examples.
  |    |                        Refer to other documents "for more info".
  |    |__ README-cmake      -- cmake specific builds
  |    |__ README-configure  -- configure specific builds
  |           |
  |           |__ README-cmake-windows     \
  |           |__ README-cmake-linux        |  hoping we can avoid this with
  |           |__ README-configure-windows  |  separate sections in above files
  |           |__ README-configure-linux   /
  |
  |__ README-cairo           -- cairo specific stuff
  |__ COPYING                -- license
  |__ etc..
```
..all the above ending in .txt so they can be clickable in browsers/folder viewers.
The only dots in filenames would be for the extension. So:
```
     BAD: ABC.def.txt
    GOOD: ABC-def.txt
```

erco77:
Oh, I didn't see Albrecht's comments after Matt's when I replied (was reading thru email).
Looks good to me; there's a lot of ways to consolidate, so I defer to Albrecht's ideas on consolidation.
But I stick to my comments on .txt extensions / no dots in filenames other than for extensions.

Albrecht-S:
> ... there's a lot of ways to consolidate, so I defer to Albrecht's ideas on consolidation ...

@erco77 Greg, thanks for your well structured documentation of your own ideas above. I see that the general structure of your approach is much like mine.

All: Regarding formatting (aka markdown): we have lots of textual documentation in our doxygen docs and a real ancient section on building FLTK with configure/make which I would like to consolidate as written above.

Doxygen docs support links and formatting (e.g. code examples) which makes me believe that it would be the best way to write well-formatted docs w/o using things like markdown which could be read locally in HTML or PDF form or online.

I believe that having the build documentation (for the library and FLTK applications) integrated in the official (and online readable) docs would make many things easier for users.

Maybe we can consolidate almost all README files to only one file that describes where the relevant documentation can be found.

erco77:
Just adding: I think it's important to have build related docs be .txt for several reasons:

1. So docs can be read in command-line only (headless) contexts such as ssh to remote machines
2. Prevents copy/paste problems with command line examples
3. Easy for devs to make fixes/commits in cli code editors

Regarding (2), problems I've seen when copy/pasting commands from non-ascii docs such as PDFs and auto-generated html: unexpected line folds, word breaks, tabs/whitespace conversions, shell-special character translations, e.g. shell quotes like (") turned into Unicode equivalents like (“,”), and similar for backquotes, etc.

Stuff like this causes support issues that are hard to troubleshoot, and pop up due to small changes in complex tool chains that generate the conversions (e.g. LaTeX backends, etc)

Albrecht-S:
@ngdream Sorry for "taking over" your PR for discussion.

PR #489, opened 2022-08-28
support to inline svg files for fluid

etorth:
Code mainly micmics ```.jpg``` file support, only change to use ```write_cstring()``` since ```.svg``` file is text-based.
Works fine in my test.


MatthiasWM (approved):
Compiled and test. It's working nicely. I keep forgetting that we can load svg images ;-) .

PR #481, opened 2022-08-12
Fluid: enhancement for widget_class for extra option: reposition-and-rescale

etorth:
A refined version of #477 

When using widget_class, in containing window or Fl_Group, we put a Fl_Box as placeholder with a ```xywh```.
And for the widget_class current fluid generates ```xywh```, but when the width and height is bigger than Fl_Box's width and height, there is no auto-rescaling. then event like click outside Fl_Box's ```xywh``` get lost.

fluid currently has the reposition choice but which only handles ```xy``` of given ```xwyh```, there is not rescaling to handle ``wh``` part.

This PR changes the ```Relative``` button into a choice with three option: ```None```/```Relative```/```Relative & Rescale```.
When choose ```None``` or ```Relative```, everything is same with previous version, but when choice ```Relative & Rescale```, the generated code will hand the rescaling issue.

take following widget_class as example:

image

code generated with reposition, but no rescaling:

```cxx
QD_CondChecker_hasItem::QD_CondChecker_hasItem(int X, int Y, int W, int H, const char *L)
  : Fl_Group(0, 0, W, H, L) {
{ Fl_Box* o = new Fl_Box(0, 0, 115, 25, "Has");
  o->align(Fl_Align(FL_ALIGN_CLIP|FL_ALIGN_INSIDE));
} // Fl_Box* o
{ m_input = new Fl_Value_Input(115, 0, 45, 25);
  m_input->labeltype(FL_NO_LABEL);
} // Fl_Value_Input* m_input
{ Fl_Box* o = new Fl_Box(160, 0, 50, 25, "Item");
  o->align(Fl_Align(FL_ALIGN_CLIP|FL_ALIGN_INSIDE));
} // Fl_Box* o
{ Fl_Choice* o = new Fl_Choice(210, 0, 145, 25);
  o->down_box(FL_BORDER_BOX);
  o->labeltype(FL_NO_LABEL);
  o->menu(menu_);
} // Fl_Choice* o
position(X, Y);
end();
resizable(this);
}
```

code generated with rescaling:

```cxx
QD_CondChecker_hasItem::QD_CondChecker_hasItem(int X, int Y, int W, int H, const char *L) :
  Fl_Group(0, 0, W, H, L)
{
  { Fl_Box* o = new Fl_Box(round(W*0.000000), round(H*0.000000), round(W*0.323944), round(H*1.000000), "Has");
    o->align(Fl_Align(FL_ALIGN_CLIP|FL_ALIGN_INSIDE));
[...]

PR #497, opened 2022-09-08
Fluid: Reposition and rescale

etorth:
Continuation of #481

### DIFF
--- a/fluid/Fl_Widget_Type.cxx
+++ b/fluid/Fl_Widget_Type.cxx
@@ -40,6 +40,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdint.h>
 
 // Make an Fl_Widget_Type subclass instance.
 // It figures out the automatic size and parent of the new widget,
@@ -754,7 +755,7 @@ void h_cb(Fluid_Coord_Input *i, void *v) {
   }
 }
 
-void wc_relative_cb(Fl_Light_Button *i, void *v) {
+void wc_relative_cb(Fl_Choice *i, void *v) {
   if (v == LOAD) {
     if (!strcmp(current_widget->type_name(), "widget_class")) {
       i->show();
@@ -2179,7 +2180,7 @@ void Fl_Widget_Type::write_static() {
   const char* t = subclassname(this);
   if (!subclass() || (is_class() && !strncmp(t, "Fl_", 3))) {
     Fl_Widget_Class_Type *wc_root = has_wcroot();
-    if(wc_root && wc_root->wc_relative){
+    if(wc_root && wc_root->wc_relative == 2){
       write_declare("#include <math.h>");
     }
     write_declare("#include <FL/Fl.H>");
@@ -2343,7 +2344,7 @@ void Fl_Widget_Type::write_code1() {
       write_c("new %s(%d, %d", t, o->w(), o->h());
   } else {
     Fl_Widget_Class_Type *wc_root = has_wcroot();
-    if(wc_root && wc_root->wc_relative){
+    if(wc_root && wc_root->wc_relative == 2){
       write_c("new %s(round(W*%f), round(H*%f), round(W*%f), round(H*%f)", t,
               1.0f*o->x()/wc_root->o->w(),
               1.0f*o->y()/wc_root->o->h(),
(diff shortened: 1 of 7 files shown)


MatthiasWM:
Sorry for the very late reply. 

OK, so I see what you are doing here, and I believe that we can do a much easier implementation. If you set relative position, the Widget Class is generate at 0, 0 first and later moved, so the Fl_Group::position() method does all the work. We can do the same for resizing. We just let Fl_Group::resize() do all the work for us. So these would be the code snippets for all three options:

Default:
```
UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
  Fl_Group(X, Y, W, H, L)
{
  { new Fl_Button(20, 20, 63, 20, "button");
  } // Fl_Button* o
  end();
}
```

Position relative:
```
UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
  Fl_Group(0, 0, W, H, L)
{
  { new Fl_Button(20, 20, 63, 20, "button");
  } // Fl_Button* o
  position(X, Y);
  end();
}
```

Position and size relative:
```
UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
  Fl_Group(0, 0, 329, 224, L)  // 329 and 224 is the original size from Fluid
{
  { new Fl_Button(20, 20, 63, 20, "button");
  } // Fl_Button* o
  resize(X, Y, W, H);
  end();
}
```

Please let me know if this solves your issues.

etorth:
It won't work if your embedded widgets has larger size than embedding widget.


> Sorry for the very late reply.
> 
> OK, so I see what you are doing here, and I believe that we can do a much easier implementation. If you set relative position, the Widget Class is generate at 0, 0 first and later moved, so the Fl_Group::position() method does all the work. We can do the same for resizing. We just let Fl_Group::resize() do all the work for us. So these would be the code snippets for all three options:
> 
> Default:
> 
> ```
> UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
>   Fl_Group(X, Y, W, H, L)
> {
>   { new Fl_Button(20, 20, 63, 20, "button");
>   } // Fl_Button* o
>   end();
> }
> ```
> 
> Position relative:
> 
> ```
> UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
>   Fl_Group(0, 0, W, H, L)
> {
>   { new Fl_Button(20, 20, 63, 20, "button");
>   } // Fl_Button* o
>   position(X, Y);
>   end();
> }
> ```
> 
> Position and size relative:
> 
> ```
> UserInterface::UserInterface(int X, int Y, int W, int H, const char *L) :
>   Fl_Group(0, 0, 329, 224, L)  // 329 and 224 is the original size from Fluid
> {
>   { new Fl_Button(20, 20, 63, 20, "button");
>   } // Fl_Button* o
>   resize(X, Y, W, H);
>   end();
> }
> ```
> 
> Please let me know if this solves your issues.

MatthiasWM:
Why not? The `resize()` will update your embedded window to fit to the requested size, resizing all its children as well. Maybe I am missing something here. The original message had sample code. Could you please post a .fl where original Fluid generates code that does not work for you, so I can try out my approach? Thanks!

> It won't work if your embedded widgets has larger size than embedding widget.
> 
> > Position and size relative:
> > ```
> >   ...
> >   resize(X, Y, W, H);
> >   ...
> > ```

Albrecht-S:
> Could you please post a .fl where original Fluid generates code that does not work for you, so I can try out my approach?

@MatthiasWM On behalf of the OP, it's all there.
1. The original test.fl file is inlined in this https://github.com/fltk/fltk/pull/481#issuecomment-1214219738
2. My modified .fl file can be found as a link in this https://github.com/fltk/fltk/pull/481#issuecomment-1214401757 together with my notes what was "wrong" in the original file (widget sizes). Direct link: [pr-481_resize_rescale_v3.fl.txt](https://github.com/fltk/fltk/files/9333563/pr-481_resize_rescale_v3.fl.txt)

I just tested both `.fl` files with current `fluid` and you can easily see the difference. Obviously the OP's proposal is intended to "fix" the code generated by `fluid` from their original `test.fl` file so it works like the file modified by me.

The discussion continued in https://github.com/fltk/fltk/pull/481#issuecomment-1214443071 where @etorth explained why his patch would be useful because you can't always "know" the exact size of a "class definition" if it has been developed in another project or by another team.

@MatthiasWM That said, I'm not sure if it's worth the effort and I wasn't sure whether the proposed code would be correct. That's why I asked for your review. I'm +/- 0 on this change and would like to leave the decision to you.

MatthiasWM:
@etorth @Albrecht-S Ok, so I implemented the feature in PR #531 . The implementation works as expected with the give .fl file, and I am pretty sure it will work well for other cases too. It's actually a nice feature that makes widget classes more useful.

PR #502, opened 2022-09-22
Fixes fullscreen_active not being correctly set

ggarra13:
When fullscreen() or fullscreen_off() were called the FULLSCREEN flag would not get set when the window was shown (the normal behavior), leading to fullscreen_active() always returning 0.

### DIFF
--- a/src/Fl_Window_fullscreen.cxx
+++ b/src/Fl_Window_fullscreen.cxx
@@ -42,7 +42,6 @@ void Fl_Window::fullscreen() {
   no_fullscreen_h = h();
   if (shown() && !(flags() & Fl_Widget::FULLSCREEN)) {
     pWindowDriver->fullscreen_on();
-  } else {
     set_flag(FULLSCREEN);
   }
 }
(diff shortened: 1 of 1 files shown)


Albrecht-S:
@ggarra13 Thanks for the PR, the patch looks good, according to your description but I'm not going to merge it because that's something Manolo knows better.

@ManoloFLTK What do you think, would you please take (assign) this PR and merge it if it's OK? Please set labels accordingly as well. Thanks

ManoloFLTK:
The lack of setting the FULLSCREEN flag was occurring only for the X11 platform, whereas the flag was correctly set and unset under other platforms. This is now corrected in the git repo (eb0bba4). This PR is not completely appropriate because it's not X11-platform-specific, and no longer useful.

ggarra13:
I verify it is now working properly at least on macOS and Linux.

ggarra13:
> El 23 sep. 2022, a las 09:30, ManoloFLTK ***@***.***> escribió:
> 
> 
> The lack of setting the FULLSCREEN flag was occurring only for the X11 platform, whereas the flag was correctly set and unset under other platforms. This is now corrected in the git repo (eb0bba4). This PR is not completely appropriate because it's not X11-platform-specific, and no longer useful.

I was seeing the problem on macOS, not just Linux.

—
Gonzalo Garramuno
***@***.***

ManoloFLTK:
@ggarra13  The FULLSCREEN flag is set and cleared under macOS inside functions Fl_Cocoa_Window_Driver::fullscreen_on() and Fl_Cocoa_Window_Driver::fullscreen_off()
which call Fl_Widget->_set_fullscreen() and Fl_Widget->_clear_fullscreen().

Therefore, I believe the problem raised here to be resolved.

ggarra13:
Correct.  I am closing the issue.

PR #503, opened 2022-09-22
This commit fixes the white corners of menus under macOS.

ggarra13:
MacOS menus would have two pixel white corners that would show up when the background was set to a dark color.  This commit fixes that display glitch.

### DIFF
--- a/src/fl_gleam.cxx
+++ b/src/fl_gleam.cxx
@@ -63,6 +63,9 @@ static void gleam_color(Fl_Color c) {
 */
 
 static void shade_rect_top_bottom(int x, int y, int w, int h, Fl_Color fg1, Fl_Color fg2, float th) {
+  // unused background - it is white on macOS but black on Linux and Windows
+  gleam_color(fl_color_average( FL_BLACK, fg1, 0.5 ) );
+  fl_rectf(x, y, w, h);
   // calculate background size w/o borders
   x += 2; y += 2; w -= 4; h -= 4;
   // draw the shiny background using maximum limits
--- a/src/fl_gtk.cxx
+++ b/src/fl_gtk.cxx
@@ -32,6 +32,12 @@ static void gtk_color(Fl_Color c) {
 }
 
 static void gtk_up_frame(int x, int y, int w, int h, Fl_Color c) {
+  gtk_color(fl_color_average(FL_BLACK, c, 0.5));
+  fl_xyline(x, y, x + w-1);
+  fl_yxline(x, y, y + h-1);
+  fl_xyline(x, y+h-1, x + w-1);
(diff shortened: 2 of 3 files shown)


Albrecht-S:
Thanks for the patch. I'm not sure though if drawing extra lines and/or a rectangle is the correct solution in general and particularly in this platform agnostic way because the title say that the issue is macOS specific. However, I can believe that we can accept this anyway.

Since I can't test and verify this right now I'm leaving this to others (e.g. @ManoloFLTK ?) to decide.

ManoloFLTK:
I'm not keen to fix a macOS-speciic issue with platform-independent code that will also be activated on non macOS platforms.

ggarra13:
> El 23 sep. 2022, a las 09:33, ManoloFLTK ***@***.***> escribió:
> 
> 
> I'm not keen to fix a macOS-speciic issue with platform-independent code that will also be activated on non macOS platforms.

Agreed.  But the problem is also there on other platforms. You just don’t see it as windows are created blank by default there.

I can add:

#ifdef __APPLE__
#endif

Guards to the code, but I find that uglier.

—
Gonzalo Garramuno
***@***.***

Albrecht-S:
@ggarra13 wrote:
> … the problem is also there on other platforms. You just don’t see it as windows are created blank by default there.

I agree that this might be an issue on other platforms as well (pixels that are never drawn). Is it possible to make a short and complete demo program that would show the issue on all platforms, for instance by setting a background color explicitly? One option to do so is using the command line switch `-bg "color"`, e.g. `-bg '#554433'` for a dark background.

Another **_important_** point to consider: the proposed patch changes box drawing code which is applied all over the FLTK GUI, not only for menus. Backgrounds can have more than just one color, for instance the tiled stipple pattern of the `plastic` scheme or even a user defined image. In such cases the patch might cause other unwanted artifacts.

That all said, IMO the proposed patch is not a proper solution of this issue. A correct patch would draw the extra background pixels only:
- on the macOS platform (maybe) and
- in the context of menus (menu windows)

Otherwise unwanted artifacts would likely cause another regression.

Disclaimer: I didn't test anything, these are just my thoughts while reading about this issue.

ggarra13:
I am closing this pull request as I have a better fix for the menus on macOS.  Please refer to the PR menuwindow_macos.

ggarra13:
> El 24 sep. 2022, a las 05:26, Albrecht Schlosser ***@***.***> escribió:
> 
> Disclaimer: I didn't test anything, these are just my thoughts while reading about this issue.
> 

You guys are right.  You forced me to create a better fix.  Please see the new PR entitled:

This is a better fix to the white menu corners on macOS.  #507.

—
Gonzalo Garramuno
***@***.***

PR #504, opened 2022-09-22
Fixes incorrect install of fluid in macOS Applications directory.

ggarra13:
This requires sudo permissions to do it, and breaks ExtenalProject_Add.  The install on the /Applications directory should be done by creating a macOS bundle (ie. a .dmg file) if this is so much desired.


Albrecht-S:
Thanks for the fix and the PR.

I see this as a temporary fix to work around the issue but I'm still considering a better fix. However, this may take a while...

I'm not yet setting the `fixed` label because this is only a workaround.

PR #507, opened 2022-09-24
This is a better fix to the white menu corners on macOS.

ggarra13:
It fixes only menu windows on macOS, without touching the buttons code.

### DIFF
--- a/src/Fl_Menu.cxx
+++ b/src/Fl_Menu.cxx
@@ -530,14 +530,15 @@ void menutitle::draw() {
 
 void menuwindow::draw() {
   if (damage() != FL_DAMAGE_CHILD) {    // complete redraw
-#ifdef __APPLE__
-    if ( Fl::is_scheme( "gtk+" ) || Fl::is_scheme( "plastic" ) ||
-         Fl::is_scheme( "gleam" ) )
+    if ( box() != FL_FLAT_BOX && ( Fl::is_scheme( "gtk+" ) ||
+                                   Fl::is_scheme( "plastic")  ||
+                                   Fl::is_scheme( "gleam" ) ) )
     {
+      // Draw a FL_FLAT_BOX so on Apple we avoid the white corners of the
+      // menus.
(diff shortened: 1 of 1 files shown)


ggarra13:
Find a demo program to test the change on macOS.
[white_corners.zip](https://github.com/fltk/fltk/files/9638776/white_corners.zip)

src/Fl_Menu.cxx
```diff
@@ -530,7 +530,16 @@ void menutitle::draw() {
 
 void menuwindow::draw() {
   if (damage() != FL_DAMAGE_CHILD) {    // complete redraw
-    fl_draw_box(box(), 0, 0, w(), h(), button ? button->color() : color());
+#ifdef __APPLE__
+    if ( Fl::is_scheme( "gtk+" ) || Fl::is_scheme( "plastic" ) ||
+         Fl::is_scheme( "gleam" ) )
+    {
```
Albrecht-S, line 536: The condition above should also check `&& box() != FL_FLAT_BOX`, otherwise the same box type would be drawn twice.

That aside, we try not to use `#ifdef __APPLE__` (i.e. any platform specific coding) in platform independent code parts (files), but this is at least a proof of concept. Thanks.

Albrecht-S:
@ggarra13 Thanks for the update.

Actually I didn't mean to remove the `__APPLE__` condition (but please don't bother to add it again). It's tricky. What I meant was that we don't want to use '#if <platform>' all over the code. This technique turned out to be badly scalable for adding new platforms. However, in some cases we still need it, particularly for "variants" of a platform like X11 with or w/o Xft, Pango, etc..

I'm not sure what to do here. Adding a "driver method" for this tiny issue seems to be inappropriate and doing it on all platforms would likely not hurt much, so maybe the current solution would be fine?

@ManoloFLTK What do you think?

ManoloFLTK:
Thanks @ggarra13 for this fix which has been committed to the repo (3ba0904).

PR #518, opened 2022-10-19
Fixed 0,0 resize of widget which could lead to drawing issues...

ggarra13:
… when widgets were hidden and shown later.

### DIFF
--- a/src/Fl_Flex.cxx
+++ b/src/Fl_Flex.cxx
@@ -242,7 +242,7 @@ void Fl_Flex::end() {
 void Fl_Flex::set_size(Fl_Widget *w, int size) {
   if (size <= 0)
     size = 0;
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Thanks for the PR. The patch looks correct at a first glance but I'm not sure that there are no other possible unwanted effects.

I'll look into this PR when I'll work further on Fl_Flex which is not yet finished and on my to-do list.

MatthiasWM:
Fixed as part of PR #523 in [93ea93e](https://github.com/fltk/fltk/commit/93ea93ee2b424ded25358d87529b1160cdb4f80c).

MatthiasWM:
Thanks @ggarra13 and @Albrecht-S

PR #523, opened 2022-10-24
Fluid support for Fl_Flex

MatthiasWM:
This PR adds Fl_Flex support for Fluid (90% done). There are two bugs fixes for `Fl_Flex::set_size(Fl_Widget*, int)` that should be applied to master in any case.

Features:
- new icon and menu entry
- code generation, all settings saved in .fl file
- additional settings in Widget dialog per Fl_Flex and per child of Fl_Flex

Fl_Flex has a new line of parameters for setting the gap and margins:
Screenshot 2022-10-25 at 01 08 19

Children of Fl_Flex will be positioned or sized by Fl_Flex, so the size line is replace by a single Flex Size field. Enter a valid integer, and the Widget will be fixed in that height or width (vertical or horizontal Flex). Enter zero, and Fl_Flex will automatically resize the widget:
Screenshot 2022-10-25 at 01 08 37

Fixes:
- Fl_Flex::set_size would not move the last entry when deleting a previous entry (off-by-one)
- Fl_Flex::set_size would continue adding the same child when called multiple times, even if the child was already in the array


Albrecht-S:
@MatthiasWM I would like to work further on Fl_Flex and if you don't mind pick your fixes and push them in my next commit. You would have to rebase and merge your own branch then. Would it be OK for you if I committed your fixes?

OTOH, how far is your fluid support? I won't be able to test much today, I'll try to find some time tomorrow. Would you be ready to merge the fluid support part soon (together with your fixes)?

src/Fl_Flex.cxx
```diff
@@ -252,25 +252,31 @@ void Fl_Flex::set_size(Fl_Widget *w, int size) {
     }
   }
 
-  // remove from array ?
+  // remove from array, if we want the widget to be flexible, but an entry was found
   if (size == 0 && idx >= 0) {
-    for (int i = idx; i < set_size_size_ - 2; i++) {
+    for (int i = idx; i < set_size_size_ - 1; i++) {
       set_size_[i] = set_size_[i+1];
```
Albrecht-S, line 258: Thanks for finding this, looks correct.

src/Fl_Flex.cxx
```diff
@@ -252,25 +252,31 @@ void Fl_Flex::set_size(Fl_Widget *w, int size) {
...
   if (size == 0)
     return;
 
-  // add to array of fixed size widgets
-  if (set_size_size_ == set_size_alloc_) {
-    set_size_alloc_ = alloc_size(set_size_alloc_);
-    set_size_ = (Fl_Widget **)realloc(set_size_, set_size_alloc_ * sizeof(Fl_Widget *));
+  // if we have no entry yet, add to array of fixed size widgets
+  if (idx==-1) {
+    if (set_size_size_ == set_size_alloc_) {
```
Albrecht-S, line 270: Nitpicking, sorry: this should read:
```
if (idx == -1) {
```
for coding style consistency.

Albrecht-S:
@MatthiasWM FYI: I added two comments to the Fl_Flex diffs. I didn't review the fluid part in any way - and I'm not sure if I will understand the code but I'll try to **test** it to get a feeling how it works.

Thank you very much for adding fluid support of Fl_Flex which covers something I would likely not have been able to do without investing much more time.

Albrecht-S:
@MatthiasWM Another question: you moved the resize statement
```
  // if w is meant to be fixed, set its new size
  w->resize(0, 0, size, size);
```
(and added the comment) to an if clause where it is only executed "`if w is meant to be fixed`". The original code did this always, right at the beginning, which I didn't change (I only added the check to prevent `size < 0`). Is there a specific reason to change this? I'm asking because I'd like to understand this and because there's PR #518 that wants to change this statement (in its original position) to `w->size(size, size);` because it "could lead to drawing issues when widgets were hidden and shown later".

Maybe these two issues are related?

src/Fl_Flex.cxx
Albrecht-S: Fixed in commit 6b650866f2fcf0ffbb64b9f90219aedcd8ec45b1.

Albrecht-S:
Hi @MatthiasWM, I took the freedom to fix the formatting issue in src/Fl_Flex.cxx (see one of my comments) and pushed it to your branch.

Albrecht-S:
@MatthiasWM I also tested the new fluid from your branch and created a working demo layout with one horizontal and one vertical Fl_Flex widget, a few buttons, and a working `quit` button.

Everything went smoothly except handling of fixed and non-fixed widgets. The tooltip of the "Flex Parent" size field says "`... or 0 to make it flexible`" children but there's also the check button `"fixed"` and have a size greater than 0 but still flexible if the check button is off. This is not a bug but a little bit unintuitive.

For the record: here is the example fluid file I created: [flex1.fl](https://github.com/fltk/fltk/files/9893892/flex1.fl.txt).

Albrecht-S:
Note that I also noticed some new compiler warnings when building fluid:
```
$ ninja fluid
[206/221] Building CXX object fluid/CMakeFiles/fluid.dir/Fl_Group_Type.cxx.o
../../fluid/Fl_Group_Type.cxx: In member function ‘virtual void Fl_Flex_Type::copy_properties()’:
../../fluid/Fl_Group_Type.cxx:196:23: warning: unused variable ‘gp’ [-Wunused-variable]
  196 |   int lm, tm, rm, bm, gp;
      |                       ^~
../../fluid/Fl_Group_Type.cxx: In member function ‘void Fl_Flex_Type::change_subtype_to(int)’:
../../fluid/Fl_Group_Type.cxx:316:7: warning: unused variable ‘dx’ [-Wunused-variable]
  316 |   int dx = Fl::box_dx(f->box());
      |       ^~
../../fluid/Fl_Group_Type.cxx:317:7: warning: unused variable ‘dy’ [-Wunused-variable]
  317 |   int dy = Fl::box_dy(f->box());
      |       ^~
[221/221] Linking CXX executable bin/fluid
```

These should be fixed and then I believe that this PR can be merged. Thank you very much for fluid support of `Fl_Flex`!

src/Fl_Flex.cxx
MatthiasWM: Yes, hadn't seen that.

MatthiasWM:
The Fluid part uses a lot of trickery because, well, Fluid is just not very clean to begin with. I have done several attempts to clean things up. The funny thing is, the more I clean it up, the more it seems to grow (in number of files), but it's actually just the unravelling of the very densely packed source files. For the time being, Fluid works. I will every once in a while go back into it and clean up a bit here and there.

As for the resize statement: I moved the resize() to the end of set_size(), because set_size(w, 0) is the only way to remove a widget from the lookup array, but if the widget is already deleted, resize() will crash (this can happen when FLTK eagerly delete all children of a widget while deleting the group widget, but I forgot the details). resize() does not serve much of a purpose in this general format here anyway. I will change that, so it also satisfies PR #518:

```
  // if the child size is meant to be fixed, set its new size
  if (horizontal())
    child->size(size, h()-margin_top_-margin_bottom_-Fl::box_dh(box()));
  else
    child->size(w()-margin_left_-margin_right_-Fl::box_dw(box()), size);
```

MatthiasWM:
As for Fluid and the size / check button combo: I did change my original code and forgot to update the tooltips. The size always shows the current size of the widget, and the check box indicates if it is fixed. I don't like the association that 0 means not fixed. 

Also, if you drag the edge of one of the children of Fl_Flex, the size entry in Fluid changes with it, and the widget is marked as "fixed".

MatthiasWM:
As for the unused variables, I removed them. I guess I should also update the warnings flags for Xcode in CMake.

Albrecht-S (approved):
Good to go, IMHO.
Please rebase (if necessary) and squash to merge. Thanks.

PS: tried to make a formally correct review, still learning "GitHub".

PR #526, opened 2022-11-01
virtual Fl_Group::on_insert, on_remove, on_move

MatthiasWM:
I have added the virtual methods above to give groups that are derived from Fl_Group a chance to react when the children array_ changes.

The need comes from Fl_Flex_ that stores pointers to widgets that are fixed in size. Fl_Flex must remove such pointers if the child Is removed, or the the pointer will be invalid, referencing a widget that is no longer a child, or may not even exist anymore.

A few notes:

Fl_Group::insert() can also change the order of children within group. It recognises that, but still removes the child and adds it again. This is bad, because Fl_Flex will first remove the "fixed" marker and then not add it again. I change the code to avoid remove/add and instead just fixed the array_ and called on_move() to preserve the additional data.

Fl_Gorup::clear() calls on_remove() before actually removing the widgets. There is an optimisation that reverses the array_ before removing the widgets, so on_remove() functions can not rely on the index order. I don't like that, but I documented it.

on_insert and on_move can return a different destination index. This is kind of dangerous, but can be a better way to solve the Fl_Scroll kludge to keep the decoration scrollbars at n-1 and n-2, making Fl_Scroll::fix_scrollbar_order() obsolete.

Happy to get feedback, as always.


Albrecht-S:
Hi Matthias, before I start a formal review, please rebase this branch on current master and force-push it so we have it on a clean base w/o all these additional commits.

Some preliminary comments:

I have a working alternative with only one virtual method but the mechanism is similar. I do also allow the insertion index to be changed to fix the Fl_Scroll issue. I'm not sure which one I find better but yours looks good as well. I'll try to post an alternate PR so we can discuss this further.

> Fl_Gorup::clear() calls on_remove() before actually removing the widgets. There is an optimisation that reverses the array_ before removing the widgets, so on_remove() functions can not rely on the index order. I don't like that, but I documented it.

Actually I removed this reversal in my patch. This reversal of child indices was introduced by me a long time ago and it might have fixed an issue which is probably no longer necessary. The reversal has nothing to do with the optimization itself and IMHO it is wrong in my view of today. What it does is:
1. reverse all children
2. delete all children starting with the last child

This effectively removes the first child first, then the second child, and finally the last child. Usually you would remove the last added child first which would be the "natural order": first in - last out.

Hence I suggest to disable the reversal of children in general. This makes this new feature (either yours or mine or a combination of both) clear because the child index is not changed for the derived class before removal of children.

> Fl_Group::insert() can also change the order of children within group. It recognises that, but still removes the child and adds it again. This is bad, because Fl_Flex will first remove the "fixed" marker and then not add it again. I change the code to avoid remove/add and instead just fixed the array_ and called on_move() to preserve the additional data.

My version is similar but since I use only one virtual method the subclass can notice this issue (it gets both indices, the current one and the one the child is to be inserted in) and - in case of Fl_Flex - can do nothing. But I like your idea to modify the array of children directly rather than doing remove and add.

Generally I think it's a matter of taste which version we may choose or how we combine our ideas.

MatthiasWM:
Generated #527. I believe I caught everything.

PR #529, opened 2022-11-04
Fluid now stores set sizes for Fl_Flex.

MatthiasWM:
[fltk.coredev](https://groups.google.com/g/fltkcoredev/c/2JA-CcTbrX4)

Fluid stored only the indices of the fixed widgets for Fl_Flex and derived the sizes from the widgets. This is not enough because automatic resizing may scramble widgets sizes before their intended size is read. So now we store the index *and* the requested size in the .fd file.

### DIFF
--- a/fluid/Fl_Group_Type.cxx
+++ b/fluid/Fl_Group_Type.cxx
@@ -214,10 +214,12 @@ void Fl_Flex_Type::write_properties()
     if (f->set_size(f->child(i)))
       nSet++;
   if (nSet) {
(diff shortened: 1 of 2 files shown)


Albrecht-S:
Please go ahead, it's now the best time to change this format because FLTK 1.4 has not yet been released and we are in the test phase of fluid's Fl_Flex support. I did not yet test this though but I don't think that this is necessary (I'll try to do it anyway later).

:+1: on applying this PR.

PR #530, opened 2022-11-04
Wayland/X11 hybrid: use "bool fl_disable_wayland;" declaration

ManoloFLTK:
This commit solves the problem of ensuring that platform-independent as well as X11-specific source code conceived for the X11 platform of FLTK 1.3 is compatible with FLTK 1.4 built with -DOPTION_USE_WAYLAND (or ./configure --enable-wayland). This is ensured provided this single declaration
`FL_EXPORT bool fl_disable_wayland = true;`
appears somewhere in the source code.

### DIFF
--- a/FL/Fl.H
+++ b/FL/Fl.H
@@ -61,8 +61,6 @@ extern FL_EXPORT const char* fl_local_ctrl;  ///< string pointer used in shortcu
 extern FL_EXPORT const char* fl_local_meta;  ///< string pointer used in shortcuts, you can change it to another language
 extern FL_EXPORT const char* fl_local_shift; ///< string pointer used in shortcuts, you can change it to another language
 
-extern FL_EXPORT void fl_disable_wayland();
-
 /** \defgroup  callback_functions  Callback Function Typedefs
 
   \brief Typedefs defined in <FL/Fl.H> for callback or handler functions passed as function parameters.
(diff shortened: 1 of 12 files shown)


Albrecht-S:
@ManoloFLTK I have only two simple suggestions:
1. we should check the value of the defined variable fl_disable_wayland (true or false)
2. we should cache the existence of the symbol rather than calling dlopen_or_dlsym() multiple times

Reasoning:
1. If we define a `bool` variable it would be intuitive for the user to set it to `false` or `true`, for instance for testing. It would be unexpected if setting it to `false` would still **disable** the wayland backend.
2. I noticed that a simple program like test/hello.cxx would call attempt_wayland() three times if `fl_disable_wayland == true`. I don't know how expensive calling dlopen_or_dlsym() is but it can easily be avoided to call it more than once.

Here's my patch proposal: [attempt_wayland.patch](https://github.com/fltk/fltk/files/9944092/attempt_wayland.patch.txt).

@ManoloFLTK If you agree to apply this patch I believe we're good to go.

Albrecht-S:
Sorry, the previously posted patch contains test statements and can be used to demonstrate that `attempt_wayland()` is called three times as described above.

The (hopefully) final patch is this: [attempt_wayland_v2.patch](https://github.com/fltk/fltk/files/9944111/attempt_wayland_v2.patch.txt)

Sorry for the noise.

Issue #153, opened 2020-11-03
Fl_Help_View <OL> don't number items.


bart9h:
`<OL>` works just like `<UL>`, using dots as the prefix for the list items, instead of numbers.

Albrecht-S:
Thanks for your report.

FTR: This is a known issue, see [STR 3290](https://www.fltk.org/str.php?L3290). Unfortunately the posted patch is not conforming to the FLTK standard (uses templates) - we'd need to find a better solution. Should certainly be doable …

Note: although I marked this issue as "bug" it's low priority.

wcout:
I felt free to take a stab on this, starting out with the patch in the STR.
This patch does not apply any longer. After manually adapting it, I found that it works in principle, but has one major flaw: because the changes have been made in the `Fl_Help_View::draw()` method, which draws only the visible part of the HTML, this leads to the numbering getting lost (switching back to bullets) when scrolling down an ordered list beyond the top of the visible area.

The changes have to made in the `Fl_Help_View::format()` method in order to get that right.

The second flaw is of course the usage of STL, which FLTK keeps to not include in its standards.
But in order to support nested ol/ul lists it is required to use some sort of dynamic containers.
So I started a _basic_ and _simple_ `fl_data_container `class that does not depend on STL. No templates - more typing, no way around...

I have added also support for the `start` property of the OL tag, which sets the first number in the list.

Here is my final and working patch:
[ol.diff.txt](https://github.com/fltk/fltk/files/5780975/ol.diff.txt)

Albrecht-S:
@wcout Many, many thanks for the patch, this is very much appreciated. I did not yet the time to test but I'll try to do this soon. Adding the container is a good solution; I also added a string (buffer) type to Fl_Help_View before as you may have noticed.

I believe that this patch has good chances to be included in FLTK.

Some minor (formatting) observations though: We're using two column indentation and we're no longer using tabs to indent code, so your code (particularly the container stuff) is not [CMP compliant](https://www.fltk.org/cmp.php#CS_GENERAL_CODING_STYLE). But this can easily be achieved by using clang-format in our source tree. I can do that if you're not familiar with it, but if you like a newly formatted patch would be appreciated as well.

Albrecht-S:
@wcout BTW, are you "Chris", the author of some good stuff (comments, patches) in our STR system?

wcout:
[...]

PR #532, opened 2022-11-06
Add Fl_Int_Vector, an STL-ish vector without templates

Albrecht-S:
We need a vector of `int`'s in several places to simplify our code. Since we can't use std::vector I propose to add a new official class `Fl_Int_Vector` for a vector of `int`'s without templates.

The original code was taken from class `Fl_Table` which used an internal class `IntVector`, copied in its own files, and reformatted.

The second commit of this PR (branch) removes the internal class definition `IntVector` from `Fl_Table` and uses `Fl_Int_Vector` instead.

Tested and works for me.

@erco77 Could you please take a look at the copied and modified code, including the copyright note? Thanks in advance.

### DIFF
--- a/FL/Fl_Int_Vector.H
+++ b/FL/Fl_Int_Vector.H
@@ -1,0 +1,83 @@
+//
+// An STL-ish vector without templates for the Fast Light Tool Kit (FLTK).
+//
+// Copyright 2002 by Greg Ercolano.
+// Copyright 2022 by Bill Spitzak and others.
+//
+// This library is free software. Distribution and use rights are outlined in
+// the file "COPYING" which should have been included with this file.  If this
+// file is missing or damaged, see the license at:
+//
+//     https://www.fltk.org/COPYING.php
+//
+// Please see the following page on how to report bugs and issues:
+//
+//     https://www.fltk.org/bugs.php
+//
+
+#ifndef Fl_Int_Vector_H
+#define Fl_Int_Vector_H
+
+#include <FL/Fl.H>
+
+class FL_EXPORT Fl_Int_Vector {
+  int *arr_;
+  unsigned int size_;
+  void init() {
+    arr_ = 0;
+    size_ = 0;
+  }
+  void copy(int *newarr, unsigned int newsize);
+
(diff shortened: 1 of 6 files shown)


Albrecht-S:
FYI: I'd like to use the new `Fl_Int_Vector` class to fix issue #153: "Fl_Help_View <OL> don't number items".

MatthiasWM:
The hoops we jump through to avoid even the lowest C++ standard... . C++03 in case anyone cares ;*)

Albrecht-S:
> The hoops we jump through to avoid even the lowest C++ standard... . C++03 in case anyone cares ;*)

OT: Yes, I'd say we need FLTK 4.0 which enables usage of some more advanced features like containers. ;-)
This could be the next minor/major version after 1.4.x (to avoid the burned version numbers 2.0 and 3.0).

erco77:
Sure -- looks good to me.

We can add other methods as needed, e.g. erase() to remove elements.
Right now to clear the array it appears one would use size(0).
If you want, I can add doxygen docs for this after you commit.

Albrecht-S:
> Sure -- looks good to me. ...
> If you want, I can add doxygen docs for this after you commit.

Merged.

@erco77 Thanks for your quick review. I'd appreciate very much if you could add the doxygen docs.

erco77:
Done; see commit 8b72f0c66.

In doing the docs, I noticed some todo items that need fixing for proper public use.

For now in the commit I simply documented the code as written, citing those as todo and warning items. But I'll follow up shortly with some code modifications that should address the worst of those todo items.

erco77:
OK, made a few followup commits, I'll probably stop there.

There's still some todo items, but those can be done on an as-needed basis and are not important.

Wasn't sure if I should add assert()s to range check subscripts, as that could arguably slow things down "unnecessarily" unless NDEBUG is set, so I left those checks out for speed. If for sure we don't want that, I can remove the \todo items for those assert() recommendations.

Would be nice to have a regression test for this class. Albrecht: any suggestion/future plans for a ./regression/ subdir?

PR #536, opened 2022-11-08
Fluid Fl_Widget::image() path fixes.

MatthiasWM:
When reading images while interpreting a `.fl` design file, the image path was seen relative to the `.cxx` file, but should have been relative to the `.fl` file. 

While fixing this, I saw the `goto_source_dir()` and `leave_source_dir()` were not recursive as they should be. Well, that is fixed too.

But now that I write this, I realise that those functions actually change to the `.fl` **design file** dir as they should. So I will fix the name of these functions in the next commit.

### DIFF
--- a/fluid/CodeEditor.cxx
+++ b/fluid/CodeEditor.cxx
@@ -2,7 +2,7 @@
 // Code editor widget for the Fast Light Tool Kit (FLTK).
 // Syntax highlighting rewritten by erco@seriss.com 09/15/20.
 //
-// Copyright 1998-2021 by Bill Spitzak and others.
+// Copyright 1998-2022 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
@@ -294,3 +294,17 @@ void CodeViewer::draw()
   CodeEditor::draw();
   Fl::set_color(FL_SELECTION_COLOR, c);
 }
+
+// ---- DesignViewer implementation
+
+/**
+ Create a DesignViewer widget.
+ \param[in] X, Y, W, H position and size of the widget
+ \param[in] L optional label
+ */
+DesignViewer::DesignViewer(int X, int Y, int W, int H, const char *L)
+: Fl_Text_Display(X, Y, W, H, L)
+{
+  buffer(new Fl_Text_Buffer);
+}
+
--- a/fluid/CodeEditor.h
+++ b/fluid/CodeEditor.h
@@ -2,7 +2,7 @@
 // Code editor widget for the Fast Light Tool Kit (FLTK).
 // Syntax highlighting rewritten by erco@seriss.com 09/15/20.
 //
-// Copyright 1998-2021 by Bill Spitzak and others.
+// Copyright 1998-2022 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
@@ -64,6 +64,14 @@ public:
 protected:
   int handle(int ev) { return Fl_Text_Display::handle(ev); }
   void draw();
+};
+
+// ---- DesignViewer declaration
+
+class DesignViewer : public Fl_Text_Display {
+public:
+  DesignViewer(int X, int Y, int W, int H, const char *L=0);
+  int top_line() { return get_absolute_top_line_number(); }
 };
 
 #endif // !CodeEditor_h
--- a/fluid/Fl_Function_Type.cxx
+++ b/fluid/Fl_Function_Type.cxx
@@ -1128,9 +1128,9 @@ void Fl_Data_Type::open() {
       if (w == data_panel_cancel) goto BREAK2;
       else if (w == data_panel_ok) break;
       else if (w == data_filebrowser) {
-        goto_source_dir();
+        goto_designfile_dir();
         const char *fn = fl_file_chooser("Load Inline Data", 0L, data_filename->value(), 1);
-        leave_source_dir();
+        leave_designfile_dir();
         if (fn) {
           if (strcmp(fn, data_filename->value()))
             set_modflag(1);
@@ -1217,9 +1217,9 @@ void Fl_Data_Type::write_code1() {
   int nData = -1;
   // path should be set correctly already
   if (filename_ && !write_sourceview) {
-    goto_source_dir();
+    goto_designfile_dir();
     FILE *f = fl_fopen(filename_, "rb");
-    leave_source_dir();
+    leave_designfile_dir();
     if (!f) {
       message = "Can't include data from file. Can't open";
     } else {
--- a/fluid/Fluid_Image.cxx
+++ b/fluid/Fluid_Image.cxx
@@ -104,9 +104,11 @@ void Fluid_Image::write_static() {
     }
     write_c("static const unsigned char %s[] =\n", idata_name);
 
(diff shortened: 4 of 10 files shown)


ManoloFLTK (commented):
1) .svgz files do work, but they are not presented in the resulting` .cxx` file as Fl_SVG_Image but as Fl_RGB_Image with their raw pixel data.

2) Running fluid as a command line with the directory of the .fl file as current dir produces two spurious error messages
`Can't chdir to (null) : Undefined error: 0`
That's because `app_work_dir` is NULL in that case. Function `leave_designfile_dir()` should may be not give an error message when `app_work_dir` is NULL

3) Gonzalo's problem is indeed fixed.

ManoloFLTK:
I suspect that the fact that FLTK 1.4 supports HighDPI displays and GUI rescaling by ctrl/+/-/0/ keys calls for a corresponding added feature to fluid as for box images. The idea is for fluid to define both sizes of an Fl_Image *img object, its data size (img->data_w() x img->data_h()) and its drawing size (img->w() x img->h()), this way a high resolution image file can be attached to a small box so the images's pixels can be used to fill the growing box's drawing size when the GUI is scaled up. The present fluid already allows to do that putting
`o->image()->scale(o->w(), o->h());`
as "Extra Code" for an image-equiped box.

What about making that easier adding in the widget's GUI property panel a "scale to box" check box next to the image input field ? When that check box would be "on", the "Extra Code" mentionned above would be automatically generated.

@MatthiasWM what do you think ?

ManoloFLTK:
@MatthiasWM 
About supporting .svgz files

I could do that, if you agree :
- move extant function of file Fl_SVG_Image.cxx
`static char *svg_inflate(const char *fname)`
to file fl_images_core.cxx and rename it
`char *gz_file_inflate(const char *fname, int &size)`
This function decompresses a gzipped file and returns the result as a char string. It also accepts a raw text file and returns its content in that case.

- in file Fluid_Image.cxx, replace
```
FILE *f = fl_fopen(name(), "rb");
……
write_cdata(data, nData);
……
fclose(f);
```
by
```
char *data = gz_file_inflate(name(), nData);
write_cdata(data, nData);
free(data);
```

MatthiasWM:
I think it is ok to store the svg in its compressed form. I can add the Fluid code to do that, no problem. What I am missing is another constructor for Fl_SVG_Image.

We currently have `Fl_SVG_Image(const char *filename, const char *inline_data_as_text)`. If filename is NULL, it uses the inline data instead. Internally it uses `strdup()` as one of the first commands in `_init()`, so binary data is not allowed.

We could do two things. Either the `Fl_SVG_Image::_init()` detects svgz data (ASCII starts with `<?xml`, svgz starts with `0x8B1F`), but the binary `unsigned char*` needs to be cast to `char*` for the constructor, or we create another constructor just for compressed data, for example `Fl_SVG_Image(const unsigned char *)`, but that would cause ambiguities if called with `void*`.

Albrecht-S:
> ... or we create another constructor just for compressed data, for example Fl_SVG_Image(const unsigned char *), but that would cause ambiguities if called with void*.

What about adding another argument `compressed = 0` to the constructor? The default value would be "not compressed" and it would be called with `compressed = 1` if it is compressed data.

MatthiasWM:
> What about adding another argument `compressed = 0` to the constructor? The default value would be "not compressed" and it would be called with `compressed = 1` if it is compressed data.

That's not needed because I can tell by reading the first two bytes if the image is compressed or not, and so does Fl_SVG_Image. I suggest `Fl_SVG_Image(const unsigned char *data, unsigned int size)`. It will still throw an error if someone calls `Fl_SVG_Image(0, 0)`, but that's a nonsense call anyway. And it would no longer need the casting for the binary data pointer.

MatthiasWM:
> Not a solution, only a hint: `Fl_Shared_Image` has all code to detect and decompress `.svgz` files.

If the user provides compressed data, it makes sense to store it compressed inside the source code. No need to unzip it within Fluid.

Albrecht-S:
> I suggest Fl_SVG_Image(const unsigned char *data, unsigned int size).

Looks good, +1 :+1:

We might want to make the size argument a signed `int` so we can use `-1` as default when no size is provided.

ISTR that we had similar issues where we needed to provide the data size in the constructor to check for buffer overrun when reading image data from memory. Unfortunately there were some image c'tors that didn't provide a `data_size` argument (and relied on correct data which would terminate the reading as needed).

> We currently have Fl_SVG_Image(const char *filename, const char *inline_data_as_text). If filename is NULL, it uses the inline data instead.

This constructor lacks the size argument and should get one anyway (with default `-1`) for the reason mentioned above: checking for buffer overrun when reading from memory. We need the default for backwards compatibility.

ManoloFLTK:
I propose this modified Fl_SVG_Image c'tor :
`Fl_SVG_Image(const char *filename, const char *svg_data = NULL, size_t length = 0);`
where the 3rd argument, when present and non null, indicates two things, 1) that  the 2nd agument is a byte buffer containing the content of a .svgz file and 2) the size of this byte buffer.

It's implemented in the attached patch, which, if we agree, will need to be also a bit further elaborated with a new member function of class Fl_System_Driver to create a temporary filename. I've also modified  Fluid_image.cxx to be able to test that it works, but that would be Matthias' part.

[SVG-gzip-ctor.patch.txt](https://github.com/fltk/fltk/files/9973961/SVG-gzip-ctor.patch.txt)

MatthiasWM:
Thanks @ManoloFLTK for your patch. Since this PR already contains a bunch of changes, I decided to merge what we have and make your patch into another PR where we can also discuss the TempFile idea.

ManoloFLTK:
@MatthiasWM Please, find attached a new patch proposed to support Fl_SVG_Image construction from a gzip'ed in-memory byte buffer. No temporary file is used anymore, even if it turned out to be a little bit tricky to do that across platforms (see "Implementation note" in source code). If we agree on that, it'll be necessary to define 2 more virtual member functions in class Fl_System_Driver.

The patch also proposes necessary changes in fluid/Fluid_Image.cxx

[SVG-gzip-ctor.patch2.txt](https://github.com/fltk/fltk/files/9995185/SVG-gzip-ctor.patch2.txt)

Albrecht-S:
@ManoloFLTK @MatthiasWM I believe there's a conflict in the way commit 7f8f7c5b851f4e15cf95c6e819bff284b7fda7ca handles the data length argument (assuming `size > 0` means _compressed data_) and issue #542 whose intention is to have all image constructors that read from memory to use the length argument for proper read termination.

This means that the assumption that `size > 0` means **compressed data** would not always be true.
From the commit log of `Fl_SVG_Image::Fl_SVG_Image(const char *filename, const char *svg_data, size_t length)`:
```
+ \param length When 0, indicates that \p svg_data contains SVG text, otherwise \p svg_data is
+  a  buffer of \p length  bytes containing GZ-compressed SVG data.
```

This needs to be checked and resolved.

MatthiasWM:
Thanks for the suggestions. I have this on the list for all image file formats. I will probably be able to verify and apply this this week.

ManoloFLTK:
@Albrecht-S @MatthiasWM I leave you decide and implement how you would like the arguments of the public Fl_SVG_Image constructor to be because I don't see the problem (I should have repeated what is documented, that, in the committed code, length > 0 implies a gzip'ed byte buffer). I notice that compatibility with 1.3 requires to accept Fl_JPEG_Image(const char * name, const unsigned char * data)

PR #539, opened 2022-11-12
Making building Fluid optional.

MatthiasWM:
Based on the great submission by @AlexMax in PR #315, this PR adds the option FLTK_BUILD_FLUID to the CMake system. The default is ON. If set to OFF, Fluid will not be built.

Examples and tests can still be built, except for those that depend on Fluid. I decided to give little warning if examples or tests are enabled, but Fluid is disable and not found.

This also solves cross-compiling where Fluid is not built because it doesn't serve much purpose, but if a host Fluid is found, all tests are still cross compiled (untested, must merge first).

### DIFF
--- a/CMake/FLTK-Functions.cmake
+++ b/CMake/FLTK-Functions.cmake
@@ -23,13 +23,10 @@
 
 function (FLTK_RUN_FLUID TARGET SOURCES)
 
-  if (NOT FLTK_BUILD_FLUID)
-    message (WARNING "Not building ${SOURCES}. FLTK_BUILD_FLUID is OFF.")
-    return ()
-  elseif (NOT FLTK_FLUID_EXECUTABLE)
+  if (NOT FLTK_FLUID_EXECUTABLE)
     message (WARNING "Not building ${SOURCES}. FLUID executable not found.")
     return ()
-  endif (NOT FLTK_BUILD_FLUID)
+  endif (NOT FLTK_FLUID_EXECUTABLE)
 
   set (CXX_FILES)
   foreach (src ${SOURCES})
--- a/CMake/FLTK-Functions.cmake
+++ b/CMake/FLTK-Functions.cmake
@@ -22,6 +22,15 @@
 # USAGE: FLTK_RUN_FLUID TARGET_NAME "FLUID_SOURCE [.. FLUID_SOURCE]"
 
 function (FLTK_RUN_FLUID TARGET SOURCES)
+
+  if (NOT FLTK_BUILD_FLUID)
+    message (WARNING "Not building ${SOURCES}. FLTK_BUILD_FLUID is OFF.")
+    return ()
+  elseif (NOT FLTK_FLUID_EXECUTABLE)
+    message (WARNING "Not building ${SOURCES}. FLUID executable not found.")
+    return ()
+  endif (NOT FLTK_BUILD_FLUID)
+
   set (CXX_FILES)
   foreach (src ${SOURCES})
     if ("${src}" MATCHES "\\.fl$")
@@ -36,6 +45,7 @@ function (FLTK_RUN_FLUID TARGET SOURCES)
     endif ("${src}" MATCHES "\\.fl$")
   endforeach ()
   set (${TARGET} ${CXX_FILES} PARENT_SCOPE)
+
 endfunction (FLTK_RUN_FLUID TARGET SOURCES)
 
 #######################################################################
--- a/CMake/FLTKConfig.cmake.in
+++ b/CMake/FLTKConfig.cmake.in
@@ -10,7 +10,7 @@
 #  FLTK_INCLUDE_DIRS      - FLTK include directories
 #  FLTK_LIBRARIES         - list of FLTK libraries built (not yet implemented)
 #  FLTK_FLUID_EXECUTABLE  - needed by the function FLTK_RUN_FLUID
-#                          (or the deprecated fltk_wrap_ui() CMake command)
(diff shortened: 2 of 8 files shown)


MatthiasWM:
Building Fluid and having Fluid are two different things. I have been trying to differentiate those by hovering a user request `FLTK_BUILD_FLUID=off` by not building Fluid, but still building most of the tests and examples, if they don't need Fluid. FLTK does not export Fluid if it isn't built, but the scripts still try to offer a version of Fluid if it found elsewhere.

I am hoping that I implemented this so that it doesn't impact existing build systems. And I am hoping that life may be a little bit easier when cross compiling. Still not perfect, but working on it.

Last thing, I changed the macro CREATE_EXAMPLE into a function so it can abort a build of a single app if Fluid is missing. It doesn't seem to have an impact on the build process. CMake experts? Is that ok?

MatthiasWM:
Sorry for pushing the patch to `upstream` instead of `origin`. It's the macOS default, and it's highly annoying.

Albrecht-S:
@MatthiasWM: So far a quick glance over the diffs did not exhibit anything wrong except some changes in `fl_create_example().` I'm not (yet) doing a full "review" though because I need some more time to test.

One question first though: why did you change the `macro` to a `function`? I'm not against this change but I'd like to understand.... 

Oh, looking closer at the code: it seems this was necessary to use `return`. Is this correct?

I believe that this part doesn't do what we want:
```
  if (flsrcs)
    if (NOT FLTK_BUILD_FLUID)
      message(STATUS "Example app \"${NAME}\" will not be built (set FLTK_BUILD_FLUID=ON to build).")
      return ()
    elseif (NOT FLTK_FLUID_EXECUTABLE)
      message(STATUS "Example app \"${NAME}\" will not be built. FLUID executable not found.")
      return ()
    else ()
      FLTK_RUN_FLUID (FLUID_SOURCES "${flsrcs}")
      list (APPEND srcs ${FLUID_SOURCES})
      unset (FLUID_SOURCES)
    endif (NOT FLTK_BUILD_FLUID)
  endif (flsrcs)
```

This doesn't build the example programs that use `fluid` if we don't build `fluid` but we are cross-compiling (and have a host `fluid` executable). In this case the first condition would be `false` and programs that use `fluid` would not be built although we **could** build them.

Here is an alternate version that works for me when cross-compiling (Linux to Windows/MinGW) with option `FLTK_BUILD_FLUID=OFF`:
```
  if (flsrcs)
    if (NOT FLUID_PATH AND NOT FLTK_BUILD_FLUID)
      message (STATUS "Example app \"${NAME}\" will not be built (set FLTK_BUILD_FLUID=ON to build).")
      return ()
    endif ()
    if (NOT FLTK_FLUID_EXECUTABLE)
      message (STATUS "Example app \"${NAME}\" will not be built. FLUID executable not found.")
      return ()
    endif ()
    FLTK_RUN_FLUID (FLUID_SOURCES "${flsrcs}")
    list (APPEND srcs ${FLUID_SOURCES})
    unset (FLUID_SOURCES)
  endif (flsrcs)
```

Please check and verify, this was a quick shot.

1. Condition 1 is true if we don't have a host `fluid` and don't build `fluid`. The latter wouldn't help when cross-compiling but ...
2. Condition 2 is true if we are cross-compiling and don't have a host `fluid`.

But I'm not sure if this is all we need.

Albrecht-S:
> I am hoping that life may be a little bit easier when cross compiling. Still not perfect, but working on it.

I don't understand this entirely. In the past we could cross-compile and build `fluid` (important!) and use a host `fluid` executable for building the test/demo programs. So what was the problem? Was it about **installing** the built programs or ... ?

Note: in earlier days (maybe fltk 1.3.5 or so) we did not build `fluid` when cross-compiling but I fixed this part long ago. Maybe the CMake code is a little confusing, setting variables and using them much later...

Albrecht-S:
> Last thing, I changed the macro CREATE_EXAMPLE into a function so it can abort a build of a single app if Fluid is missing. It doesn't seem to have an impact on the build process. CMake experts? Is that ok?

Yes, I think it's OK. I missed that statement when reading earlier, hence I asked why you did it in my previous comment. But this is now answered.

Albrecht-S:
> Sorry for pushing the patch to upstream instead of origin. It's the macOS default, and it's highly annoying.

Yes, indeed, it's annoying. It's not a problem for me and likely other devs but I would really like to keep the "upstream" FLTK repository "clean" so other users don't need to download lots of commits with unfinished modifications. There may be users with limited bandwidth and the entire repo grows with each commit.

Albrecht-S:
OK, after further looking at my proposed code it seems that the message of the first condition **can** be misleading:
> Example app \"${NAME}\" will not be built (set FLTK_BUILD_FLUID=ON to build)

If this happens when cross-compiling then setting `FLTK_BUILD_FLUID=ON` would not help. It's only a minor issue but anyway, there may be more cases we need to distinguish and write appropriate warning messages. ... Or we can do it like Microsoft: "Example ... will not be built because `fluid` might not  be available for one of many possible reasons." ;-)

MatthiasWM:
> > I am hoping that life may be a little bit easier when cross compiling. Still not perfect, but working on it.
> 
> I don't understand this entirely. In the past we could cross-compile and build `fluid` (important!) and use a host `fluid` executable for building the test/demo programs. So what was the problem? Was it about **installing** the built programs or ... ?

It seemed just like a waste of time when you just want to test out the examples. Most other cases can actually be solved by setting the `--target` options in the camp build phase.

AlexMax:
Fantastic!  Thanks for bringing this over the finish line. 👍

Issue #543, opened 2022-11-14
Fluid: warnings when recompiling .fl files


Albrecht-S:
**Describe the bug**
Fluid outputs warnings "**_Can't chdir to (null) : Success_**" when recompiling some fluid (`.fl`) files with `fluid -c`.
The same happens when "updating" fluid files with `fluid -u`.
This does not happen for all fluid files. For examples see log below.

**To Reproduce**
1. Checkout a fresh FLTK library
2. Build the library with `make` (no options necessary)
3. Change directory to `fluid`
4. Run e.g. `./fluid -c about_panel.fl`

**Expected behavior**
Silent exit, no warnings (and files updated).

**Screenshots**
```
fltk/fluid $ for x in "-c" "-u"; do for f in *.fl;do echo; echo "--- ./fluid $x $f ---"; echo; ./fluid $x $f; done; done

--- ./fluid -c about_panel.fl ---

Can't chdir to (null) : Success
Can't chdir to (null) : Success

--- ./fluid -c alignment_panel.fl ---


--- ./fluid -c function_panel.fl ---


--- ./fluid -c print_panel.fl ---

Can't chdir to (null) : Success
Can't chdir to (null) : Success

--- ./fluid -c template_panel.fl ---


--- ./fluid -c widget_panel.fl ---


--- ./fluid -u about_panel.fl ---

Can't chdir to (null) : Success

--- ./fluid -u alignment_panel.fl ---


--- ./fluid -u function_panel.fl ---


--- ./fluid -u print_panel.fl ---

Can't chdir to (null) : Success
Can't chdir to (null) : Success

--- ./fluid -u template_panel.fl ---


--- ./fluid -u widget_panel.fl ---
```

**FLTK Version**
Please complete the following information and delete non-applicable lines:
 - Version: 1.4.0
 - Git branch: master
 - Commit: 85af0036f9fc733b2e8ad3ae62fd00bc783db738

**FLTK Configure / Build Options**
See above, just `make` with default config.

**Operating System / Platform:**
 - OS: Linux
 - OS Version: Mint 20

**Additional context**
The same happens in directory `src` for `Fl_File_Chooser.fl` but not for `Fl_Help_Dialog.fl`.

Albrecht-S:
FWIW: this bug appeared in commit 56cba05672faa953f3e078de6a2db2992dd888b9.

MatthiasWM:
Fixed in #545

Issue #514, opened 2022-10-14
fluid -c crashes on Linux


MoAlyousef:
**Describe the bug**
Invoking `fluid -c` without arguments from the command-line causes a segmentation fault.

**To Reproduce**
Call fluid -c without arguments:
```
$ fluid -c
```

N/A

**Expected behavior**
A file dialog opens on macos for example. Or maybe a help menu should be printed to the console.

**Screenshots**
N/A

**FLTK Version**
Please complete the following information and delete non-applicable lines:
 - Version: 1.4.0
 - Built from source via CMake

**FLTK Configure / Build Options**
 - cmake -Bbin -DCMAKE_BUILD_TYPE=Release -DFLTK_BUILD_EXAMPLES=OFF -DFLTK_BUILD_TEST=OFF && cmake --build bin --target install

**Operating System / Platform:**
Linux: Ubuntu 20.04 on WSL

**Linux/Unix Runtime, if applicable:**
 - X11

ManoloFLTK:
This should be fixed with commit 531ab61. 
@MoAlyousef : Please, close if you confirm.

MoAlyousef:
Hi @ManoloFLTK 
Thanks, this is indeed fixed.

Albrecht-S:
Sorry, this quick fix is good in itself (no need to revert it) but it does not really fix the issue "`fluid -c` w/o argument" because ...

... the file chooser dialog opens a file for "**saving**" which is unexpected and unintuitive.
- If you select an existing file you are asked "Do you want to replace it?" and fluid would overwrite the file if answered with "yes".
- If you select a non-existing file fluid writes an (almost) empty file and the corresponding .cxx and .h files.

A better fix would be to reject the request with a help message that an input file is required as @MoAlyousef suggested.

Reopening this issue.

Albrecht-S:
The same would very likely be true if `fluid -u` is used w/o argument.

PR #545, opened 2022-11-14
Improve Fluid argument handling and relative paths.

MatthiasWM:
This commit fixes the warnings from Issue #543 and the crash in issue #514 .

#543 was a bug that was probably there for a while, but finally detected by an error message that I added in a recent commit. I rewrote the function that make sure that all paths within .fl files are seen relative to the file itself.

#514 was caused by an incomplete handling of invalid command line arguments. The new handler adds more error handling, especially if batch mode is requested, but no filename is given. It also outputs a brief usage hint when the user types `fluid -h`. The old parameter syntax `fluid -c -h .hxx project.fl` still works.

### DIFF
--- a/fluid/Fl_Function_Type.cxx
+++ b/fluid/Fl_Function_Type.cxx
@@ -1128,9 +1128,9 @@ void Fl_Data_Type::open() {
       if (w == data_panel_cancel) goto BREAK2;
       else if (w == data_panel_ok) break;
       else if (w == data_filebrowser) {
-        goto_designfile_dir();
+        enter_project_dir();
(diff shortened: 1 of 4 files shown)


Albrecht-S (approved):
Code looks good.
Note that I fixed some typos in an additional commit (pushed to your branch).

Note also that Manolo's recent commit introduced merge conflicts.

Albrecht-S:
Note also: I tested `fluid -u -c` with all available .fl files, this PR fix issue #543. Thanks.

MatthiasWM:
@Albrecht-S thanks for checking, sorry for the typos.

Issue #541, opened 2022-11-13
Reminder: Fl_Scroll::add() must use on_move() when indicated


MatthiasWM:

MatthiasWM:
Ran several test adding and moving widgets around.

PR #546, opened 2022-11-15
Fix rearranging children in Fl_Scroll

MatthiasWM:
Implement Fl_Scroll::on_move()
Fix indexing in Fl_Group::on_move()


MatthiasWM:
See #541.

PR #548, opened 2022-11-18
Have OPTION_USE_WAYLAND / --enable-wayland set to ON by default

ManoloFLTK:
Please, report whether you agree to apply this PR that makes the hybrid Wayland/X11 platform become the default platform built with CMake or with configure+make under Linux and Unix hosts.

The CMake code checks for presence of software modules necessary for Wayland, and falls back to the usual X11 platform if any module is missing. That takes care of older Linux configurations which don't support Wayland.

Platform-independent source code for FLTK 1.3 is compatible with the 1.4 Wayland/X11 hybrid libfltk because 1.4 is API-compatible with 1.3.

X11-specific source code for FLTK 1.3.x is compatible with the 1.4 Wayland/X11 hybrid libfltk (this PR included) provided environment variable FLTK_BACKEND is defined and set to "x11" when FLTK apps are run. Alternatively, adding
  `FL_EXPORT bool fl_disable_wayland = true;`
to any 1.3 source code is enough to make any 1.3 source code compatible with the 1.4 hybrid Wayland/X11 libfltk with that single change.


imaclmaca:
On the basis that the behaviour for "legacy" X11 based systems is the same as before, then I don't see a problem with making this change.

At present most (I think possibly all... I have not explicitly checked all machines) my *nix boxes are X11-only; though some could have Wayland, I have not yet installed it anywhere. 
I do also have some X11 specific code, in some apps, that is not likely to be Wayland compatible without further work, but since I generally only build under X11 I do not expect that to be a problem in the short-term.

So in the immediate period, I do not expect that this change will affect my code all that much - and that in the future I will have to "embrace" Wayland support, at which point this proposed change will be a clear advantage.

CMake/options.cmake
```diff
@@ -194,7 +194,9 @@ set (HAVE_LIBPNG 1)
 if (UNIX)
   option (OPTION_CREATE_LINKS "create backwards compatibility links" OFF)
   list (APPEND FLTK_LDLIBS -lm)
-  option (OPTION_USE_WAYLAND "support Wayland and X11 backends" OFF)
+  if (NOT APPLE)
+    option (OPTION_USE_WAYLAND "support Wayland and X11 backends" ON)
```
imaclmaca, line 198: I wonder if this ought to read "support BOTH Wayland and X11 backends" or similar, to make it clearer that this is the hybrid case?

CMake/options.cmake
```diff
@@ -203,9 +205,13 @@ if (UNIX)
...
+      message (STATUS "Consequently, OPTION_USE_WAYLAND is set to OFF.")
+      unset (OPTION_USE_WAYLAND CACHE)
+      set (OPTION_USE_WAYLAND 0)
     endif (NOT(DBUS_FOUND AND WLDCLIENT_FOUND AND WLDCURSOR_FOUND AND WLDPROTO_FOUND AND XKBCOMMON_FOUND))
+  endif (OPTION_USE_WAYLAND)
+    
+  if (OPTION_USE_WAYLAND)
     option (OPTION_WAYLAND_ONLY "support Wayland backend only" OFF)
     set (FLTK_USE_WAYLAND 1)
     if (NOT OPTION_WAYLAND_ONLY)
```
imaclmaca, line 217: Does line 215 setting OPTION_WAYLAND_ONLY to OFF mean that this line (217) can never be true?

configure.ac
```diff
@@ -978,7 +978,7 @@ AS_CASE([$host_os_gui], [cygwin* | mingw*], [
         THREADS="threads$EXEEXT"
     ])
 
-    AS_IF([test x$enable_wayland = xyes], [
+    AS_IF([test x$enable_wayland != xno], [
```
imaclmaca, line 981: Is this a null-change?
This seems to change "== yes"  into "!= no", which is the same thing?
Or am I missing the point here?

CMake/options.cmake
ManoloFLTK: Good suggestion. Will do.

CMake/options.cmake
ManoloFLTK: Line 215 does not set the option to off. It creates the option and its associated tooltip text, and declares its default value is OFF. The user can then turn it on or off, and line 217 controls what happens next.

configure.ac
ManoloFLTK: This would be a null change if `enable_wayland` could have only 2 values (yes or no). But a 3rd situation is possible where `enable_wayland` is empty, arising when neither `--enable-wayland` nor `--disable-wayland` are present in the `configure` command line. The code in master treats this as  `--disable-wayland` whereas the PR has it treated as `--enable-wayland`.

CMake/options.cmake
ManoloFLTK: Done.

.github/workflows/build.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       # Note the current convention is to use the -S and -B options here to specify source
       # and build directories, but this is only available with CMake 3.13 and higher.
       # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
-      run: cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+      run: cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DOPTION_USE_WAYLAND=0
```
Albrecht-S, line 36: I'd prefer boolean option values ON or OFF rather than 1 or 0, respectively.

The correct (full) syntax would be `-D OPTION_USE_WAYLAND:BOOL={ON|OFF} where `:BOOL` can be omitted.

Note: There may be other existing '1' settings which I'd like to be changed as well. Thanks.

Albrecht-S (approved):

Albrecht-S:
+1 :+1: on merging this with changes pointed out by my review.

configure.ac
imaclmaca: OK.
Might be useful to have a comment somewhere to remind folks this is not a binary option, but instead tertiary...
I will certainly forget!

imaclmaca:
After the clarifications to my comments, I'm OK for this now.
+1

imaclmaca (approved):
OK from me...

.github/workflows/build.yml
ManoloFLTK: Done.

MatthiasWM (approved):
I did not get too much into the Wayland branch, but I guess it's ok to set this hybrid as a default, as long as it's documented well, so that building FLTK does not get more complicated for the user.

Albrecht-S:
@ManoloFLTK We have +3 votes (+4 including you), hence we should merge it. Are you ready to merge?

ManoloFLTK:
> @ManoloFLTK We have +3 votes (+4 including you), hence we should merge it. Are you ready to merge?

@Albrecht-S  Will do. I'm thrilled to see the Wayland branch become the default FLTK configuration for Linux!

ManoloFLTK:
Have OPTION_USE_WAYLAND / --enable-wayland set to ON by default

This makes the hybrid Wayland/X11 library the default configuration of FLTK under Linux.

Albrecht-S:
@ManoloFLTK Great, thank you very much.

Issue #412, opened 2022-03-15
Selecting text with keyboard in Fl_Text_Editor stops working sometimes


gnuwimp:
If you select text with mouse and then move/drag that text to another position, then selecting text with keyboard stops working. And you have to select some text with the mouse to get it working again.

It might be because of dragType is set to DRAG_START_DND and not turned off again after moving text.

Tested on Ubuntu 21.10 with FLTK from 2022-03-15.

MatthiasWM:
Not happening on macOS. I will try to replicate this on Ubuntu when I have access to a machine.

Albrecht-S:
Testing to add some observations, I'm not trying to fix anything.

1. I confirm the issue on Linux (Mint, based on Ubuntu) with **X11** backend
2. I confirm the issue on Linux (Mint, based on Ubuntu) with **Wayland** backend
3. I also confirm that the issue doesn't exist on **macOS** with default config (**Quartz**?)
4. However, I can replicate it on **macOS** with `./configure --enable-x11` (**X11** backend, running with XQuartz)
5. The problem doesn't exist on Windows AFAICT (cross-compiled on Linux with `wine`, not tested on a native Windows box)

My test scenario:
- launch `test/editor test.txt`
- select text with mouse **OR** keyboard (`Shift` + `cursor keys`)
- drag the selected text to another place
- press `Shift` + `cursor keys` as before: doesn't work on some platforms (see above)

The OP @gnuwimp wrote:
> ... you have to select some text with the mouse to get it working again.

In my tests it's been enough to click with the mouse anywhere in the editor to position the cursor to a new place.

HTH

MatthiasWM:
I think that resetting the drag state to none in ::handle FL_DRAG after the Fl::dnd() call should do the trick. Cocoa seems to send FL_RELEASE after the dnd which resets the state, X11 does not.

Albrecht-S:
> Cocoa seems to send FL_RELEASE after the dnd which resets the state, X11 does not.

If that's true, then one of the platforms needs to be fixed, independent of this particular issue. Events should be as close as possible on all platforms (compared to other platforms). Note that I don't know which one is correct, i.e. sending FL_RELEASE or not.

But at least this explains the difference.

ManoloFLTK:
The issue should be fixed in the Wayland backend with commit d215dde.

MatthiasWM:
Thanks Manolo!

PR #549, opened 2022-11-21
Add missing FL_RELEASE event after dnd()

MatthiasWM:
On X11, FLTK would not send a required FL_RELEASE
event to the data source window.

This is an attempt at fixing #412 without in-depth knowledge of X11. fl_dnd_x.cxx seems to go through various attempts of correctly delivering a DND package. The patch seems to work well at finalising the DND with an FL_RELEASE event at what was most likely the source widget for the drag and drop operation.

### DIFF
--- a/src/fl_dnd_x.cxx
+++ b/src/fl_dnd_x.cxx
@@ -202,6 +202,7 @@ int Fl_X11_Screen_Driver::dnd(int unused) {
   }
 
   fl_local_grab = 0;
+  Fl::handle(FL_RELEASE, source_fl_win);
   source_fl_win->cursor(FL_CURSOR_DEFAULT);
   return 1;
 }


MatthiasWM:
Thanks for looking at that, @Albrecht-S.

Albrecht-S:
@MatthiasWM Reminder: please update the `master` branch in your fork before creating a new PR (i.e. a new branch to create a PR). Working on outdated stuff can be problematic. I rebased my local branch `matt-dnd_release_event` and `git rebase` shows:
```
On branch matt-dnd_release_event
Your branch and 'matt/dnd_release_event' have diverged,
and have 20 and 1 different commits each, respectively.
```

Would you like to fix this, rebase your branch, and force-push it or shall I do this for you? It would be only one git command for me...

MatthiasWM:
Thank, yes, when I created the PR it was current. I will rebase it alter tonight.

ManoloFLTK:
1) The issue should be fixed for the Wayland backend with commit d215dde.

2) I believe the proposed fix for X11 is correct because the macOS platform that doesn't show the issue does send an FL_RELEASE event at the end of the drop operation.

Thus, I would propose to commit this PR and close issue #412.

Issue #196, opened 2021-03-01
Extending selected text with keyboard in Fl_Text_Editor


gnuwimp:
When you select text in the buffer and then try to extend it with the shift and keyboard it selects the wrong text range.

Example:
run "editor editor.cxx"
press ctrl + f and enter "https"
then press shift + right and it selects all text from pos 0 until "https:"

It might have something to do with Fl_Text_Display::dragPos which is not set when you select text with the Fl_Text_Buffer::select().
Anyway I have no idea how to fix it.

MatthiasWM:
I can replicate that. I will fix that hopefully this week.

Albrecht-S:
Maybe this is related to #412?

MatthiasWM:
Selecting text after the "find" action does not seem to set the cursor position correctly. I am traveling RN, but I will look at that tonight.

MatthiasWM:
Ok, so this is more complicated than I thought. Programmatically selecting text (Editor->Find, for example), does not set dragPos, which is the base for changing the selection with shifted arrow keys. Not sure how to solve this as dragPos/mCursorPos shadow the buffer()->selection(), which should probably not be part of the buffer anyway.

This will get complicated!

MatthiasWM:
Some more details. We have two places where we store the text selection. The first selection is associated with the text buffer, and the second one is associated with the text display. The buffer selection is the visible one, but the display selection is used for shifted movements and for drag'n'drop. 

If the buffer selection is created by mouse or cursor movements, both selections coincide. But if the buffer selection isdone programatically (e.g. by using "test/Editor->Find"), the display and buffer selections diverge. We *show* the buffer selection, but we *use* the display selection.

This gets even more complex when we have multiple displays using the same buffer... .

Spontaneously, I would remove the buffer selection altogether, but that interferes with existing software. Making buffer selection change all associated displays is feasible, but the buffer does not know which displays are connected, and keeping back pointers is dangerous if a display is deleted without unlinking the buffer.

So, um, I am open for suggestions.

erco77:
Heh, you sure seem attracted to the tough problems; fluid menus, and now the text editor X^D
All that you need to complete your migraine is to start probing around in the HTML viewer code.

I'm not looking at the code, but from what I remember, I'm going to guess that if a text buffer is shared by both an editor and a display, the idea would be any edits in the Fl_Text_Editor would be displayed by the associated Fl_Text_Display too.

[...]

PR #550, opened 2022-11-22
Fix selection extension in Fl_Text_*, issue 196

MatthiasWM:
Selecting a text range programmatically would not sync some variables with the actual selection. This also fixes a crash bug in macOS when dragging text that was selected by buffer()->select() only (*1).

This PR fixes issue #196 .

*1) this is actually funny in a way. Initiating a DND operation did not copy the text to be dragged to the clipboard first as Fl:dnd(1) expects. On macOS this would crash the app. This does not happen if a selection was mad with the cursor keys or the mouse. Replicate: open any text with test/Editor, before using any mouse buttons or keystrokes, use Find to select something, then drag that selected text somewhere. macOS will crash before the DND operation is initiated.

### DIFF
--- a/src/Fl_Text_Display.cxx
+++ b/src/Fl_Text_Display.cxx
@@ -3961,12 +3961,13 @@ void Fl_Text_Display::draw(void) {
   fl_pop_clip();
 }
 
-// GitHub Issue #196: internal selection nd visible selection can run out of
+// GitHub Issue #196: internal selection and visible selection can run out of
 // sync, giving the user unexpected keyboard selection. The code block below
 // captures that and fixes it.
-// set pos to the drag target postion or -1 if we don;t know
-// if pos is -1, and key is not -1, key can be set to indicate a direction (e.g. FL_Left)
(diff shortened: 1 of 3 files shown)


src/Fl_Text_Display.cxx
```diff
@@ -3961,7 +3961,41 @@ void Fl_Text_Display::draw(void) {
   fl_pop_clip();
 }
 
-
+// GitHub Issue #196: internal selection nd visible selection can run out of
```
Albrecht-S, line 3964: should read:
"internal selection and visible selection"

Albrecht-S:
I'm not sure what to test and what to confirm. However, I notice that this commit does not fix issue #412 (see my [test scenario](https://github.com/fltk/fltk/issues/412#issuecomment-1322093434)).

MatthiasWM:
This is fixing issue #196. Issue #412 is fixed in PR #549, but not for Wayland. I would need someone who's good with X11 check if #549 is ok, and with Wayland knowledge to port the fix over there. I also must check the docs.

Issue #262, opened 2021-07-20
Add functions for drawing rounded rectangles to the drawing api 


wyhinton:
Rounded rectangles have become a critical part of modern GUI design. Having a method for drawing them would be very handing. While it is nice to be able to set the global radius size for rounded frames, we might also want to have rounded frames with different radiuses.  So, I think we could save people time by providing a method for drawing rounded rectangles a custom radius. 

Such a method already exists in fl_rounded_box:
```C
static void rbox(int fill, int x, int y, int w, int h) {
  int i;
  int rs, rsy;
  rs = w*2/5; rsy = h*2/5;
  if (rs > rsy) rs = rsy; // use smaller radius
  if (rs > RS) rs = RS;
  if (rs == 5) rs = 4;  // use only even sizes for small corners (STR #2943)
  if (rs == 7) rs = 8;  // note: 8 is better than 6 (really)

  if (fill) fl_begin_polygon(); else fl_begin_loop();
  for (i=0; i<RN; i++)
    fl_vertex_r(x + offset[RN-i-1]*rs, y + offset[i] * rs);
  for (i=0; i<RN; i++)
    fl_vertex_r(x + offset[i]*rs, y + h-1 - offset[RN-i-1] * rs);
  for (i=0; i<RN; i++)
    fl_vertex_r(x + w-1 - offset[RN-i-1]*rs, y + h-1 - offset[i] * rs);
  for (i=0; i<RN; i++)
    fl_vertex_r(x + w-1 - offset[i]*rs, y + offset[RN-i-1] * rs);
  if (fill) fl_end_polygon(); else fl_end_loop();
}
```

MatthiasWM:
With the current API, we do not have any additional parameters for box types besides the color. So storing the radius is only possible globally. It is however possible to add custom box types that could have different radii. See `Fl::set_boxtype()`.

If that would solve the issue at hand, I'll take a look at the API and check if there is a way to make this option more reachable, maybe add an example.

MatthiasWM:
See #553 for an implementation of `fl_rrect(int x, int y, int w, int h, int radius)` and `fl_rrectf(int x, int y, int w, int h, int radius)`. If I don't get any negative comments, this will be in 1.4.0 in a day or two.

Albrecht-S:
Please refer to my request for better names in https://github.com/fltk/fltk/pull/553#issuecomment-1325177388.

MatthiasWM:
[...]

PR #553, opened 2022-11-23
Refactor code to make rounded rectangles accessible

MatthiasWM:
This adds fl_rrect and fl_rrectf (note the additional `r`) so the user can draw rounded rectangles. This uses existing and optimised code that is rearranged.

As mentioned in issue #262, we already implemented drawing code for rounded rectangles, but we did not have an API to access the code besides via box types. I also optimised to code to ignore drawing matrices, do a faster lookup, and avoid searching the graphics driver multiple times.

### DIFF
--- a/FL/Fl_Graphics_Driver.H
+++ b/FL/Fl_Graphics_Driver.H
@@ -265,8 +265,8 @@ public:
   virtual void focus_rect(int x, int y, int w, int h);
   virtual void rectf(int x, int y, int w, int h);
   virtual void _rbox(int fill, int x, int y, int w, int h, int r);
-  virtual void rrect(int x, int y, int w, int h, int r);
-  virtual void rrectf(int x, int y, int w, int h, int r);
+  virtual void rounded_rect(int x, int y, int w, int h, int r);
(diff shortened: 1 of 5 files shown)


Albrecht-S:
> This adds `fl_rrect` and `fl_rrectf` (note the additional r) ...

I didn't look at the code yet, but IMHO these function names can too easily be confused with their "normal" drawing functions (which might cause long searches for coding errors).

Can we please change the function names to `fl_round[ed]_rect()` and `fl_round[ed]_rectf()` or other names that are better to distinguish?

Albrecht-S:
Regarding requested name changes: I believe it's important for the public API but it would be helpful to change the internal driver function names as well to remain consistent.

MatthiasWM:
The internal naming absolutely should reflect the public API. Fixed that.

Albrecht-S:
Note: I pushed a commit to your PR (branch) to remove an unused variable.

Tested with test/boxtype, rounded boxes still look fine. I didn't test explicitly calling the new functions though. Anyway, +1 from me.

PR #554, opened 2022-11-24
Check boundary conditions in Cocoa scroll code

MatthiasWM:
This fixes scrolling in Cocoa if the requested scroll source area is outside of the window, and no source pixels would exist.

[google groups coredev report](https://groups.google.com/g/fltkcoredev/c/sAxgewLotNM) including code where this fails.


ManoloFLTK (approved):
OK and thanks for these changes.

PR #555, opened 2022-11-24
Add i18n to test/preferences

MatthiasWM:
To test the output of FLUID code and demonstrate
i18n, preferences emulates GNU gettext.

The FLUID setup uses GNU gettext for internationalisation (i18n). FLUID finds the texts that need to be translated and writes them into .po files that can be processed by the GNU gettext tools. FLUID produces source code that will translate all text into the current locale when generating the UI. In this small example, 'getttext' is only emulated.

### DIFF
--- a/test/preferences.fl
+++ b/test/preferences.fl
@@ -24,7 +24,7 @@ will translate all text into the current locale when
 generating the UI.
 
 
-In this small example, 'getttext' is only emulated.} {selected in_source not_in_header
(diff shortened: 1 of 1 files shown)


MatthiasWM:
After a bit back and forth I decided to merge this into master. The code is really harmless, but it illustrates nicely the possibilities and limits of i18n with FLTK .

PR #556, opened 2022-11-25
Handle Fluid project settings better

MatthiasWM:
Project settings were not completely cleared when a new project was loaded, leaving i18n settings and others behind.

Using Fl_String in some cases to simplify code.

Fixes bug in Fl_String that would keep it from expanding the buffer.


MatthiasWM:
Tested with Valgrind on Linux.

Issue #349, opened 2021-12-19
Fluid: separate system settings (global and user) to another program


Albrecht-S:
Fluid is a developer tool and the menu
- Edit
- .. Global FLTK Settings

should be accessible to users and system admins w/o having to use such a developer tool.

AFAICT we added these global and user system preferences to fluid because we needed a quick solution for all the new `Fl::option()` settings.

I propose to add a new tool that users can use to set their own preferences and system admins can use to set the system preferences (maybe only once at installation time).

I propose the name "`flups`" for "`FL`TK `U`ser `P`reference`s`" but I'm asking for opinions and better proposals.

- Shall we add such a tool?
- Which name should that tool have?
- Does anybody know of conflicts with other (Unix/Linux) tool names if we used `flups` or any other proposal?

MatthiasWM:
- + 1 👍 on adding that tool
 - other name option is Floss (FLTK operational system settings), Flux (FLTK User eXperience is probably taken)

ManoloFLTK:
+1 to separate the control of FLTK options from fluid.
This tool would control what the API calls  [Fl::Option](https://fltk.gitlab.io/fltk/classFl.html#a43e6e0bbbc03cad134d928d4edd48d1d). Its name should be based on "option",
possibly `floptions`.

MatthiasWM:
I have made a PR #560 for this.

MatthiasWM:
About to close #560

erco77:
Late to the party, yeah +1 for floptions and not flups.

We have to keep in mind that executables like fltk-config, fluid, etc. are going to be part of the linux distributions appearing in their bin directories, and therefore in all user's PATH when installed, so we want to play nice with the rest of the world's executable names.

MatthiasWM:
2 minutes after the merge ;-) . Anyway, it's `fltk-options` by majority vote.

erco77:
Oh, well that's good too!

PR #560, opened 2022-11-27
Move global FLTK options into new app fltk-admin

MatthiasWM:
As discussed many times, FLTK global settings should be in an app separate from FLUID.

I added `fltk-admin` in the `test` directory that works as a command line app and as a GUI app with pretty much the same abilities as the old FLUID implementation.


Albrecht-S:
> I added fltk-admin in the test directory that works as a command line app and as a GUI app with pretty much the same abilities as the old FLUID implementation.

@MatthiasWM Great, awesome, wonderful! This was really needed.

I didn't test it yet but I will do as soon as possible.

One thing came to mind recently: ISTR that running Wayland GUI programs with `root` permission is not allowed and tried it [1]. I conclude that we definitely need a commandline tool to set **_system_** preferences/options. It's great that we will get one for other reasons as well (e.g. scripting system setup). Thanks!

[1] I tried `sudo fluid` and it launched the `X` version of `fluid` whereas just `fluid` runs the `Wayland` version of the `X11/Wayland` hybrid build. If I disable `X11` (e.g. `unset DISPLAY`) `sudo fluid` doesn't start (error message `Can't open display`) as expected. The commandline version (`fluid -c ...`) works even with `sudo` (qed).

I'll review and test soon...

MatthiasWM:
I used the same basic idea from FLUID this as well. The command line launch will not open the Display at all. Without any command line args, we get the interactive version. I doubt that anyone will try to run in GUI mode as superbster. Admins tend to use command lines. fl_preferences will not give an error though if it can't write the system settings. May I should add that at some point.

Albrecht-S:
First test results and my "wishlist":
- [ ] Unrecognized command line options should issue an error message and exit rather than starting interactive mode.
- [ ] We should test whether we have write permissions to the system options (file) and if not ...
- [ ] (a) issue an error message if a system option should be set (command line)
- [ ] (b) gray out (deactivate) system options in interactive mode
- [ ] Output of system or user options in commandline mode should also output the option name [1]
- [ ] Commandline option `--help` should be supported (not only `-h`)
- [ ] Maybe add an option to list all system + user options for info (e.g. `fltk-admin -l` ?)

Notes:
 - I tested setting system options w/o root permission and it failed silently.
 - My first intuitive test was to use `--help` which started the interactive mode. :-(
 - fltk-admin -l could also support an option name which would list only this option (system + user)
 - output of system or user options in commandline mode should also output the respective option name

[1] potential improvement:
```
$ bin/test/fltk-admin -UVisibleFocus -SVisibleFocus
1
-1
```
could be changed to something like:
```
$ bin/test/fltk-admin -LVisibleFocus
VisibleFocus: User: 1, System: -1 
```

Other than that: my tests on Linux were fine so far (no code review yet). Thank you!

Albrecht-S:
> Makefile/makedepend (@Albrecht-S)

I assume and (if not) suggest to create a separate folder for `fltk-admin` after the test phase (it should IMHO not be kept in the test folder and I don't think this was your intention, right?

- If this is done I can volunteer to integrate this in `CMake` (`CMakeLists.txt`).
- I don't feel comfortable with making a new `Makefile` with all options though [1]. Someone else?
- Updating dependencies (`makedepend`) is easy, I can do this once we have the `Makefile`.

[1] Basically we could start copying `fluid/Makefile` and build `fltk-admin/Makefile` from that. Maybe I can try this.

Albrecht-S:
> I doubt that anyone will try to run in GUI mode as superbster. Admins tend to use command lines.

Yes, I agree, and that's why we need the command line. However, for a quick look and a one-click-action it might be possible that an admin uses the GUI for system settings. As far as I could test this works with X11 as root (see above) but in the far future, when there are systems that don't support X11 - and if Wayland still refuses to run GUI programs as root - then there is really no other choice than using the command line.

erco77:
Matt: note that on Windows we had to create a separate `fluid-cmd` exe built with `/subsystem:console` for the command line version of fluid so that it works properly in Makefiles.. not because of the GUI stuff, but because a `/subsystem:windows` will (a) background itself and thereby can't return exit codes, and (b) can't properly interact via stdin/stdout with the dos window terminal the program was invoked from.

Not sure if the new admin tool falls into this category, but sounds like it might if it's something that could be called from a scripting context like a batch file, Makefile, perl/python/etc, so it should probably generate a separate xyz-cmd.exe on windows too, similar to fluid.

Building with `/subsystem:console` isn't a solution either because then when users click on that app, a separate DOS window provided by the OS is opened (for stdin/out), so we really had to go with separate binaries.

Albrecht-S:
@erco77 Good points.

More thoughts:
1. Should we add a default scheme (name) as another option?
2. The current GUI woudn't scale well if we added more options. Should we create separate tabs for groups of options (maybe as in `unittests`) or put everything in an `Fl_Scroll` group?

I would say "yes" to both points and would prefer a resizable window with an Fl_Scroll group for convenient access. Better ideas for the UI?

I don't want to make it more complicated than necessary, but if we touch it now anyway we can likely make it easier for future extensions.

Albrecht-S:
@ManoloFLTK made a good point in [this comment](https://github.com/fltk/fltk/issues/349#issuecomment-999353228).

At a first look I thought `fltk-admin` is OK but after thinking about it and reading Manolo's argument I'm not sure that `admin` is the best choice. In many cases users will want to edit **their** preferences and having a program named `..-admin` sounds wrong for "normal users".

Maybe `fltk-prefs` or, as Manolo suggested, something with "option" would be better, maybe `fltk-options`. Given these alternate choices I would propose `fltk-prefs` because it would be short and it would say what it does (changes) for FLTK **users**.

erco77:
Regarding Wayland and running GUIs as root, it sure seems like they're not gonna allow it. That was a design decision. Apparently on linux there's something called `polkit` (aka. 'policy kit') that allows GUI apps to do things as root without running as root. Some sort of `admin://` prefix on files and I'm not sure what else, but see the [wikipedia page on polkit](https://en.wikipedia.org/wiki/Polkit) for an intro.

I'm just looking into this myself, as I was going down the rabbit hole of 'wayland guis can't run as root' in this thread:
https://bugzilla.redhat.com/show_bug.cgi?id=1274451
..which points in lots of directions, but sounds important as apparently this Wayland transition is being treated as the "line in the sand", the end of continuing to allow GUIs to run as root, the official edict being _fix your broken code already_.

erco77:
+1 for `fltk-options`, unless the tool might grow to affect more than fltk options. (Not sure what that would be tho, so fltk-options sounds good)

MatthiasWM:
Thanks for the intense feedback.

> Unrecognized command line options should issue an error message

Ah, it did, but I messed up. It's changed back.

> Output of system or user options in command line mode should also output the option name

I already added a `-v` option for that.

> Command line option --help should be supported 

I just added that

> Maybe add an option to list all system + user options

`-h` already lists all available options. The current settings can now be listed with `-L`, `-LS`, or `-LU`.

> I assume and (if not) suggest to create a separate folder for fltk-admin after the test phase

Maybe. We will get the request to disable building it with yet another cmake option. Admins who need it can just use `cmake ... --target fltk-admin`.

> note that on Windows we had to create a separate fluid-cmd

Sigh. I read all the discussions about it. Is that really the only way? Is the rest of the world doing it like that as well?

> Should we add a default scheme (name) as another option?

The app understands all FLTK command line options, like `-scheme oxy`. If we want to do a new default scheme, shouldn't it be in all apps, or actually be in the library?

> The current GUI wouldn't scale well if we added more options

Over the last 10 years, two options were added and two were removed. I am not too worried that we will run out of space, but in any case, the internal code is very easy to expand (just a lookup array), and the UI is quickly adjusted. This is btw why giving an unsupported option is generating a warning and no error, so existing scripts can continue to set the known options.

I will add the grading out next. Should be merged tonight.

Albrecht-S:
> The app understands all FLTK command line options, like -scheme oxy.

That was not what I meant. Sure, it's good to know that you can use it for this app as well.

> If we want to do a new default scheme, shouldn't it be in all apps, or actually be in the library?

I meant to add a new default scheme as system and user preference - just like all the other options. Changing the library to use another default is independent of setting a system or user preference. We can do this, but since the scheme is a matter of taste it should IMHO be a preference (aka option) users can set for themselves and/or the system manager for the system. This would give the following precedence (lowest priority first):

1. Library default, currently 'none' or 'base', lowest priority. I don't think we need to change this.
2. System default set with this app by an admin.
3. User default set with this app by the user.
4. Environment variable `FLTK_SCHEME`: overrides the default set in options above.
5. Commandline option `-scheme` of program invocation: almost highest priority, overrides environment variable.
6. Hard coded Fl::scheme("...") in program: should really be avoided because it would override everything.

All points except 2. and 3. exist already, points 2. and 3. would be added by such an option in this app. If all user programs are coded correctly, i.e. using Fl_Window::show(argc, argv) so the commandline works and **not** using a hard coded scheme (6.) then users can set their preferred scheme and forget about it.

Environment variable `FLTK_SCHEME` (4.) and the commandline option (5.) should only be used for special cases, for instance testing. The environment variable is used in test/demo to set the selected scheme for all launched test programs.

Note that points 1. - 5. would all be implemented in the initialization part of the library and would not require code in user programs.

MatthiasWM:
Point 6 would depend on the author, and if they put the command before or after the Fl::args() call. The correct position would be before.

MatthiasWM:
+1 @erco77 `fltk-options` is probably the better name

erco77:
Matt: I cloned your repo and checked out fltk-admin and did a clean build with `make` (which used configure defaults).. seemed not to build anything called 'fltk-admin' in tests, so I tried `( cd test && make fltk-admin )` and got an error:
```
[erco@harris] 1740 : make fltk-admin
Compiling fltk-admin.cxx...
fltk-admin.cxx:18:10: fatal error: fltk-admin-ui.h: No such file or directory
   18 | #include "fltk-admin-ui.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [../makeinclude:172: fltk-admin.o] Error 1
```

I loaded up the fl and hit File -> Write Code, and it got a little further but failed with a missing function:

```
[erco@harris] 1742 : !make
make fltk-admin
Compiling fltk-admin.cxx...
Compiling fltk-admin-ui.cxx...
Linking fltk-admin...
/usr/bin/ld: fltk-admin.o: in function `main':
fltk-admin.cxx:(.text.startup.main+0x15f): undefined reference to `create_options_editor(int, int)'
collect2: error: ld returned 1 exit status
make: *** [../makeinclude:163: fltk-admin] Error 1
```

MatthiasWM:
@erco77 Please try the CMake setup. I only slammed some code into the Makefile unchecked. See checkbox 3 in my first list ... .

erco77:
To Albrecht's point, a small UI sugg, perhaps for later once the app is stable:

An Fl_Browser might be a good way to provide for more options in the future. A UI something like the following might also allow text descriptions, and with separate scrolls for the 'options' and 'settings', it'd allow for multiple sub options and long text descriptions that might scroll off the page. It'd also make it easier for the user to find stuff if the options were in alphabetic order.
```
    Options                       Settings
     _______________________    ______________________________________________________________________________
    | Keyboard focus      | |  |                                              :   System    :    User       | |
    | Tooltips            | |  |                                              :             :               | |
    | Drag+drop           | |  |----------------------------------------------------------------------------| |
    | Native File Chooser | |  |  Visible Keyboard Focus                      :             :               | |
    | Print dialog        | |  |                                              :             :               | |
    | Scaling factor      | |  |  This controls if a dotted line box          :  _________  :  _________    | |                         
    |                     | |  |  appears around input fields indicating      : |_default_| : |_default_|   | |                         
    |                     | |  |  the field has keyboard focus.               :             :               | |
    |                     | |  |  The default is 'on'                         :             :               | |
    |                     | |  |______________________________________________:_____________:_______________| |
    |                     | |  |                                              :             :               | |
    |                     | |  |  Arrow Keys Move Focus                       :             :               | |
    |                     | |  |                                              :             :               | |
    |                     | |  |  In early versions of fltk, the arrow keys   :  _________  :  _________    | |
    |                     | |  |  allowed navigation of keyboard focus, but   : |_default_| : |_default_|   | |
    |                     | |  |  standards since then use Tab and Shift-Tab  :             :               | |
    |                     | |  |  to navigate keyboard focus.                 :             :               | |
    |                     | |  |  The default is xxx.                         :             :               | |
    |_____________________|_|  |______________________________________________:_____________:_______________|_|

                           ^                                                                                ^
                           Scrollbar                                                                        Scrollbar

```

Albrecht-S:
> An Fl_Browser might be a good way to provide for more options in the future ...

... or something embedded in an Fl_Grid once it's ready (RSN). ;-)

ManoloFLTK:
> Regarding Wayland and running GUIs as root, it sure seems like they're not gonna allow it. That was a design decision. 

Under Debian, I've turned root with `su` and then ran a Wayland-enabled FLTK app. The app runs without apparent trouble. `ps -ef` shows `root` is the user of the corresponding process. `xwininfo` shows the FLTK app's window is not managed by X11.

ManoloFLTK:
> Under Debian, I've turned root with `su` and then ran a Wayland-enabled FLTK app. The app runs without apparent trouble. 

I notice now an interesting detail : when run as root, the app looses the minimize button of its titlebar. See attached where the same binary file runs as regular user at left and as root at right :
Wayland-as-root

erco77:
I think if you dig thru that bugzilla link you'll see there are some x11 oriented workarounds in place to support the transition, but from what I read the eventuality is those doors will be closed at some point, the argument being that gui code is not designed to run as root (think web browsers and movie players) which run enormous amounts of GUI library code (media libraries etc) that are not designed to be security hardened. Even polkit itself had a rather large [privilege escalation vulnerability](https://www.bleepingcomputer.com/news/security/linux-system-service-bug-gives-root-on-all-major-distros-exploit-released/) that apparently had been in there for >10 yrs before it was recognized.

Albrecht-S:
> ... the argument being that gui code is not designed to run as root ...

_OT:_ ... the **absurd** (current) **situation** being: if you start a "hybrid" `Wayland/x11` GUI program (FLTK, Qt, ... more?) as root Wayland refuses to start and you get an even **less secure** X11 version running. Tested for instance with `gedit` and some more apps.

I understand that this is a "temporary" effect until all systems are delivered w/o X11. ;-)
_/OT_

Albrecht-S:
@MatthiasWM First of all, thank you for implementing my entire "wishlist"!

> [Albrecht]: I suggest to create a separate folder for fltk-admin after the test phase
> [Matt] We will get the request to disable building it with yet another cmake option.

Yes, and this is OK (we can't avoid it, see below). But this would be much simpler than the same option for fluid because we don't need fltk-admin for our build process.

> Admins who need it can just use cmake ... --target fltk-admin.

No, that doesn't work. If it is included in the test programs (`test` folder) and you disable building the test programs (which would be the only reason to build `fltk-admin` separately) then the target `fltk-admin` will never be created and you can't build it.

> [regarding separate console app on Windows]
> Is that really the only way?

The only other option I can imagine is to use a commandline argument to force creating a console window but this wouldn't play well with a script (if the executable is a GUI program) for reasons Greg mentioned.

> Is the rest of the world doing it like that as well?

I don't know what the rest of the world are doing. We didn't create an extra console version of fluid for more than 20 years because we were not aware of the issues, maybe because nobody complained. When Greg found it out the hard way we noticed and fixed it.

> [setting Fl::scheme("...") in the application program ...]
> ... would depend on the author, and if they put the command before or after the Fl::args() call. The correct position would be before.

I agree and I believe this needs some advice in the documentation, e.g. when to call it and what this would imply. That is, what is a good programming convention to set a "default scheme per program" w/o overriding users' choices. But this is OT and should be handled separately.

ManoloFLTK:
> 
> _OT:_ ... the **absurd** (current) **situation** being: if you start a "hybrid" `Wayland/x11` GUI program (FLTK, Qt, ... more?) as root Wayland refuses to start and you get an even **less secure** X11 version running. Tested for instance with `gedit` and some more apps.
> 

I'm getting lost here. What I see is that Wayland-enabled FKTK apps do run as root.
I see that with Debian sid, and also with Ubuntu 22.04 under which it's necessary to set
an environment variable `export XDG_RUNTIME_DIR=/run/user/1000` for apps to start as root.

erco77:
@ManoloFLTK On rocky9 (similar to RHEL 9) with the wayland window manager, if I'm logged in as a normal user, I get an error trying to run wayland enabled FLTK apps in a `su - root` shell, similar to the errors in the bugzilla page.
```
Mon 11/28/22 11:09:20 /usr/local/src/fltk-1.4.x-git/build/bin/test
[root@rocky] 506 # export FLTK_BACKEND=wayland

Mon 11/28/22 11:09:28 /usr/local/src/fltk-1.4.x-git/build/bin/test
[root@rocky] 507 # ./hello
error: XDG_RUNTIME_DIR not set in the environment.
Error: no Wayland connection available, FLTK_BACKEND = 'wayland'
```
Pretty sure this is intentional and the intentional behavior going forward.
I can only guess that what you might be encountering is an X11 fallback, or some kind of workaround allowing the behavior, as for sure some utilities have not yet switched over to this polkit and whatever else is needed to get actual work done as root.

MatthiasWM:
Since we are drifting in all funny directions, a `/subsystem:windows` app can be convinced to reconnect to the console that launched it:
```
  FreeConsole();
  AttachConsole(ATTACH_PARENT_PROCESS);
  freopen("CONOUT$", "w", stdout);
  puts("This will go to the parent's console no matter what process creation flags were used");
```
The app launches asynchronously though. To get the return value in `cmd.exe`, we must run:
```
start /wait .\fluid.exe
echo %errorlevel%
```
and on `powershell`, that could be
```
.\adjuster.exe  | Out-Default
$lastexitcode
```
So maybe we can stop generating `fluid-cmd` and don't need `fltk-admin-cmd`?

Albrecht-S:
> What I see is that Wayland-enabled FKTK apps do run as root.
I see that with Debian sid, and also with Ubuntu 22.04 under which it's necessary to set
an environment variable export XDG_RUNTIME_DIR=/run/user/1000 for apps to start as root.

I can only tell what I see on my system (Linux Mint 20, based on Ubuntu 20.04 LTS). That said, setting `XDG_RUNTIME_DIR=/run/user/1000` is AFAICT calling for trouble. Since the default user id under Debian/Ubuntu/Mint is 1000 you ask the process running as root (sudo) to write into the user's runtime directory. I'm not sure, but wouldn't this write files with owner root into the user's runtime directory?

Maybe current Debian systems are setting a policy (with `polkit`, as Greg said)? I don't know...

erco77:
Regarding wayland+root, someone correct me if I'm wrong, it would appear that wayland interaction is done through socket(s) in the /run/user/<UID> directory, and it would seem if you have permission to write to that socket (e.g. root), you can interact with wayland. My understanding is wayland's permission model is to use file system permissions, which I take to mean on such socket(s), and perhaps other things (like shared memory?)

I imagine with selinux and ACLs, the distros _could_ start tightening things down so such sockets are unreadable even to root - I know there's security like this in MacOS, where even root can't do things like remove or change files in /bin and /usr/bin no matter how hard you try, in the name of security.

erco77:
@MatthiasWM the `start /wait` has had in the past some terrible bugs, like not pass exit codes through, and bad issues with stdio redirection/pipeing, so it's much more stable to go with a console app.

Albrecht-S:
OK, here's something I can do on my system which shows what happens: wayland apps open a socket to connect to the wayland compositor. I open a `weston` instance with socket `weston-123` as my user with user id 1000.
```
$ weston -S weston-123
```
Weston opens its own window on my desktop.
In another terminal session:
```
$ ls -l /run/user/1000/weston-123
srwxrwxr-x 1 albrecht albrecht 0 Nov 28 23:05 /run/user/1000/weston-123
```
OK, the `weston` socket has been created.

Then, as user root (`sudo su - root`) I can start the editor `xed` (or `gedit`) which opens our `README.txt` and connects to the Weston display `weston-123`:
```
/fltk/master# XDG_RUNTIME_DIR=/run/user/1000  WAYLAND_DISPLAY=weston-123  xed README.txt 
```
The result is that `xed` runs as `root` and opens a `Wayland` GUI in my `weston` compositor window. `xed` has the advantage that it clearly shows that it's running with "Elevated Privileges":

wayland-root

The main window is the wayland compositor (weston) and the inner window is the editor xed.

So, it "works" but as I wrote before, now a `root` application uses the runtime directory of a normal user with all consequences. This is just a proof of concept, I don't recommend doing it this way.

MatthiasWM:
So I rewrote pretty much the entire app to have a better and extensible database of all available options. I stuck mostly with the old UI plus a scroll bar that show up if the window is resized. 

There are still a few things that I have on my list:
 - [ ] Should the app get its own directory like FLUID? Maybe not.
 - [ ] Should the app install like fltk-config? Probably yes.
 - [ ] Should there be fltk-config-cmd for MS Windows? Probably yes.
 - [ ] Do I have to write a Doxygen page? Sigh, yes.
 - [ ] Do I really have to write a man page? I am afraid so.
 - [ ] Should the app be available via test/demo? Probably yes.
 - [ ] Should I really go to the dentist tomorrow? Will it hurt? Certainly!

MatthiasWM:
The top design is my current design. The window resizes vertically and shows a scrollbar if the list no longer fits or options are dded later.

The second design is a quick-and-dirty test of Eric's suggestion @erco77 . It can be activated by changing the `#define FO_GUI_LAYOUT 0` line in the source code to 1. The list on the right would also scroll, as would the browser on the left, but the screen size itself is fixed.

Any votes on this?


Screenshot 2023-01-18 at 14 56 36

MatthiasWM:
Apart from the UI layout question, I am done with the app and would like to merge it into master soon.

erco77:
While the first looks less busy, I do like being able to read the descriptions easily, though as Manolo says, tooltips are a good way to data hide in this case. (I do suggest in the final app to sort the order of the browser alphabetically.)

If you were to go with the second design (who's Eric? :^D), I'd suggest changing the layout just a little bit for clarity:
image
Namely:
- De-emphasize the text descriptions by making it a lighter gray instead of black
- Text description above, chooser widgets below
- Dividers with 'white space' above + below
- I'd suggest in this UI making the scrollbar's width small, like 9 instead of the default 15
- The text can be reduced -- I can help if you want. Text like "If visible focus is switched on.." to just "If visible focus is on.." I find small tweaks like this throughout can really shorten walls of text; the shorter the better. In this context "terse-speak" is probably apropos, where you get right to the point, kinda like writing tweets lol.

That all said, your first layout would be fine too, with tooltips for descriptions.

MatthiasWM:
- [ ] don't forget to close #349
- [ ] add Zenity file chooser option
- [ ] use Greg's suggestions for second style GUI (sorry, Eric ;-) )

PR #565, opened 2022-11-30
Adding a unit test for drawing complex shapes

MatthiasWM:
As we know, OpenGL does not deal with non-convex polygons, giving us render errors in complex shapes.

This PR is to show the issue, but not to fix it.

This adds `fl_load_matrix()` and `fl_load_identity()` for completeness of matrix operations.

This improves the render quality of small Bézier curves.


MatthiasWM:
CI successful.

PR #566, opened 2022-11-30
Fluid: Keep widget panel open during undo/redo

MatthiasWM:
Using undo and redo in FLUID used to always close the widget panel


MatthiasWM:
CI returns no errors

Issue #569, opened 2022-12-05
Fl_win32.cxx RegisterDragDrop cause memory leak


xhawk18:
line 2021 in Fl_win32.cxx , there's RegisterDragDrop  called there.
after removed this function call, the memroy leak goes away.

Albrecht-S:
Thank you for the report.

Given the line number 2021 in `Fl_win32.cxx` you are using FLTK 1.3.x which will very likely not be updated anymore. The same code is in [line 2246 of Fl_win32.cxx](https://github.com/fltk/fltk/blob/a8923a0fd4424a07fff247fe1167909264f8a117/src/Fl_win32.cxx#L2246) in FLTK 1.4.0 (git).

Since calling RegisterDragDrop is done per window I agree that this causes a memory leak and this should be fixed in FLTK 1.4.0. According to Microsoft docs we should probably call [RevokeDragDrop](https://learn.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-revokedragdrop) when the window is destroyed.

Commenting out or removing this call is not an option because it is required for drag and drop operations. 

@xhawk18: How did you find the memory leak? Can you give us (me) some advice so we can test and verify a potential solution?

xhawk18:
> @xhawk18: How did you find the memory leak? Can you give us (me) some advice so we can test and verify a potential solution?

thank you for the reply!  
please try test code below, and may found memory grows slowly.

```
#include <FL/Fl.H>
#include <FL/Fl_Window.H>

void hide(void *arg) {
    Fl_Window *win = (Fl_Window *)arg;
    Fl::remove_idle(hide, win);
    win->hide();
}

int show() {
    Fl_Window win(220, 90);
    win.show();

    Fl::add_idle(hide, &win);

    return Fl::run();
}

int main() {
    while (true) {
        show();
    }
}
```

Albrecht-S:
I believe I found a possible solution, see attached patch [issue-569_v1.diff](https://github.com/fltk/fltk/files/10171124/issue-569_v1.diff.txt). It's maybe not the last version but I'm asking for feedback.

@xhawk18 This patch is for FLTK 1.4 (git) exclusively. Can you please try FLTK 1.4 and this patch and report if it solves the problem for you? I tried your demo program and it solves at least the memory leak. Please do not use in production software, it's still not tested thoroughly.

xhawk18:
@Albrecht-S I test the patch, and no memory leak found now! thank you!

[...]

PR #583, opened 2022-12-09
Windows: fix memory leak caused by RegisterDragDrop (#569)

Albrecht-S:
We need to call RevokeDragDrop() to fix the leak but we must also prevent premature deletion of the FLDropTarget object.


Albrecht-S:
Please see details at issue https://github.com/fltk/fltk/issues/569

PR #589, opened 2022-12-10
Bind image to a widget

MatthiasWM:
So the widget will release/delete the images when the widget itself is deleted, see #3 .


MatthiasWM:
Calling bind_image() with a new image will delete the old image if it was bound, and then set the new image, and bind that. If old and new image are the same, nothing happens.

Calling image() with a new image will delete the old image if it was bound, and set the new image without binding it. If old and new are the same, the image will not be deleted, but it will be unbound.

Calling image() or bind_image() with NULL will delete the old image if it was bound and not set a new image.

Shared images are handled.

src/Fl_Widget.cxx
```diff
@@ -18,6 +18,7 @@
 #include <FL/Fl_Widget.H>
 #include <FL/Fl_Group.H>
 #include <FL/Fl_Tooltip.H>
+#include <FL/Fl_Shared_Image.H>
```
Albrecht-S, line 21: This should no longer be needed since the second commit doesn't use it.

Should be removed.

Albrecht-S:
Code looks good (untested). I'm not sure (and don't have time to check) that all "special cases" mentioned in [this comment](https://github.com/fltk/fltk/pull/589#issuecomment-1345324777) are also documented. It's important for the user docs. Particularly "If old and new are the same, the image will not be deleted, but it will be unbound" might be unexpected (but it's OK).

Thank you for taking the original PR and making such a good solution (IMHO).

wcout:
My thoughts, at  a quick glance:

- Isn't there missing an important part, namely the `release()` of the images in the widget destructor `~Fl_Widget()` , or does this happen somewhere else?

- And should
`````cxx
void bind_image(int f) { set_flag(IMAGE_BOUND); }
`````
not be:
`````cxx
void bind_image(int f) { f ? set_flag(IMAGE_BOUND) : clear_flag(IMAGE_BOUND); }
`````
or something like that?

Same with `bind_deimage()`.

Albrecht-S:
@wcout Good points, thank you for reviewing the code!

MatthiasWM:
> * Isn't there missing an important part, namely the `release()` of the images in the widget destructor `~Fl_Widget()` , or does this happen somewhere else?

`~Fl_Widget` is calling `image(NULL);` and `damage(NULL);` which will delete the images if bound to the widget.

> * And should
> 
> ```c++
> void bind_image(int f) { set_flag(IMAGE_BOUND); }
> ```
> 
> not be:
> 
> ```c++
> void bind_image(int f) { f ? set_flag(IMAGE_BOUND) : clear_flag(IMAGE_BOUND); }
> ```

Oops ;-) . it's fixed in the recent commit.

MatthiasWM:
> Thank you for taking the original PR and making such a good solution (IMHO).

Ah, sure. It's fun.

Issue #542, opened 2022-11-13
Reminder: image readers need to use size argument to avoid illegal memory access


MatthiasWM:

Albrecht-S:
Note: the intention of this issue is to **_add a length argument to all image constructors that read from memory_** in the first place and then use this length in the image readers for proper read termination at "end of memory" to avoid data overrun, just like "end of file" would do.

Please refer to my comment [here](https://github.com/fltk/fltk/pull/536#issuecomment-1315195854) and check commit 7f8f7c5b851f4e15cf95c6e819bff284b7fda7ca for compatibility.

MatthiasWM:
Personal note, API's to check:

- [ ] base classes Fl_Image, Fl_Bitmap, Fl_Pixmap, Fl_RGB_Image, Fl_Shared_Image, Fl_Tiled_Image
- [ ] Fl_XBM_Image
- [ ] Fl_GIF_Image
- [ ] Fl_XPM_Image
- [ ] Fl_BMP_Image
- [ ] Fl_JPEG_Image
- [ ] Fl_PNG_Image
- [ ] Fl_PNM_Image
- [ ] Fl_SVG_Image
- [ ] Fl_ICO_Image

Albrecht-S:
Reopening for reference until #592 is merged.

PR #592, opened 2022-12-11
Adding length checks for in-memory image data (see #542)

MatthiasWM:


MatthiasWM:
Personal note, check the following APIs.

base classes:
- [x] `Fl_Image` : no initialisation from memory
- [x] `Fl_Bitmap` : *fixme*, can initialise from memory, but no length check
- [ ] `Fl_Pixmap` : *fixme*, can initialise from memory, but no length check
- [ ] `Fl_RGB_Image` : *fixme*, can initialise from memory, but no length check
- [x] `Fl_Shared_Image` : no initialisation from memory
- [x] `Fl_Tiled_Image` : no initialisation from memory
image type implementations
- [ ] `Fl_XBM_Image` : *todo* add initialisation from memory as an option?
- [ ] `Fl_GIF_Image` : *todo* verify, already has a constructor with length
- [ ] `Fl_XPM_Image` : *todo* add initialisation from memory as an option?
- [ ] `Fl_BMP_Image` : *todo* verify, already has a constructor with length
- [ ] `Fl_JPEG_Image` : *fixme*, can initialise from memory, but no length check
- [ ] `Fl_PNG_Image` : *todo* verify, already has a constructor with length
- [ ] `Fl_PNM_Image` : *todo* add initialisation from memory as an option?
- [ ] `Fl_SVG_Image` : *todo* verify, already has a constructor with length, but check binary vs. text data 
- [ ] `Fl_ICO_Image` : *todo* verify, already has a constructor with length

Albrecht-S:
Matthias, my intention in #542 was to
1. enable the user program (e.g. fluid) to tell us the length of the in-memory data
2. check this memory size when reading the image data at runtime to avoid reading over the memory bounds.

This should prevent reading over the allocated memory size. Imagine a user program (or fluid) reading an image from a broken file into memory and then loading the image from memory. Provided the program sets the read size correctly, this should return a read error at "end of memory".

I'm wondering if it is good (correct) to reallocate the data in memory before reading the image if the given size is smaller than the calculated size. As you did it (reallocate and fill with 0's) this will be unexpected (another data address in memory). My gut feeling is that we should not do this, I just have a bad feeling with this.

What I wanted was the same behavior we have when reading from files: we get an EOF flag set when **reading** beyond the end of the file. The same should happen when we read from memory.
`Fl_Image_Reader` has already been improved to support this (commit bc0d18c1bb6ce205f034161ea7816638b5d7f5b7) and IIRC we're using this new feature in the GIF reader and maybe elsewhere.

Implementation note: `Fl_Image_Reader` sets the error (EOM) flag when reading over the memory limits and returns zero for all subsequent reads. It's the responsibility of the caller to check the error status from time to time, i.e. at critical points.

It was clear that we didn't support the memory size in many of the image constructors that read from memory. This should be fixed and then it should be used only to terminate reading if we read beyond EOM (end of memory) if and only if the size was provided by the constructor.

BTW: A logical consequence would be that fluid uses the additional size argument for all in-memory images, but that's another point for later.

MatthiasWM:
I apologize that I messed around quite a lot in the SVG source code. This version decompresses SVGs in memory without the need for a pipe, which should be faster. It tests for compression by checking the signature, so misnamed SVGs are still decompressed correctly. There is also an option to make an SVG into a shared image with the in-memory constructor.

MatthiasWM:
> Matthias, my intention in #542 was to
> 
> 1. enable the user program (e.g. fluid) to tell us the length of the in-memory data

Yes, all image readers have now an option to add a length for binary data. The user program know the length of the data (either by compiling binary data into the app, or by reading a file into memory), and can pass it on, and FLTK will not read beyond that point.

> 2. check this memory size when reading the image data at runtime to avoid reading over the memory bounds.

That's what all the readers do now, assuming the correct length was passed for binary data. Text data is still expected to be terminated with a NUL.

> This should prevent reading over the allocated memory size. Imagine a user program (or fluid) reading an image from a broken file into memory and then loading the image from memory. Provided the program sets the read size correctly, this should return a read error at "end of memory".

Yeah, that's what it does. I may be setting `ERR_FORMAT`, but I can of course change that to `ERR_FILE_ACCESS` or add a new `ERR_MEMORY_ACCESS`.

> I'm wondering if it is good (correct) to reallocate the data in memory before reading the image if the given size is smaller than the calculated size. As you did it (reallocate and fill with 0's) this will be unexpected (another data address in memory). My gut feeling is that we should not do this, I just have a bad feeling with this.

It happens only if the size is too small which would be an error. We could trigger an assert when in DEBUG mode, and if you prefer, I make it throw an error instead of adding 0s. That reduces complexity.

> What I wanted was the same behavior we have when reading from files: we get an EOF flag set when **reading** beyond the end of the file. The same should happen when we read from memory. `Fl_Image_Reader` has already been improved to support this (commit [bc0d18c](https://github.com/fltk/fltk/commit/bc0d18c1bb6ce205f034161ea7816638b5d7f5b7)) and IIRC we're using this new feature in the GIF reader and maybe elsewhere.

Yes, I saw that code. Nice!

> It was clear that we didn't support the memory size in many of the image constructors that read from memory. This should be fixed and then it should be used only to terminate reading if we read beyond EOM (end of memory) if and only if the size was provided by the constructor.

That's what my new code is doing. In Fl_JPEG_Image, for example, if called with no length argument, we can read far into user memory. If used *with* length, we utilise the builtin JPEG memory reader that does not read beyond available data. Same for compressed SVG.

> BTW: A logical consequence would be that fluid uses the additional size argument for all in-memory images, but that's another point for later.

It is doing that for SVG now. Still needs to be done for JPEG. PNGs interestingly are stored and read as RGBA data which should be changed to mirror JPEG and SVG. I have not tested GIF or ICO yet.

Albrecht-S:
I need to review the current code (tomorrow), I'm now replying only to this: regarding reallocating the user's data and adding zero's. Note, this is only a "management view", not a code review!

You wrote:
> It happens only if the size is too small which would be an error.

The question is: why? Is this really necessary? I believe reallocating should always be avoided. If the user gives us an address, we should not reallocate the data in memory. Let us just postpone error detection until we read beyond the given memory - as we would do from a file. No reallocation, no length check before reading, nothing like that.

> We could trigger an assert when in DEBUG mode, and if you prefer, I make it throw an error instead of adding 0s. That reduces complexity.

I hope it is possible to check for out-of-bounds reads in all image readers. If this is possible we should definitely not add zeroes to (reallocated) data. If we did it (and read that data later) it could lead to unrecognized errors because we would read zeroes rather than terminating with an error.

It would IMHO be fine if we terminated loading the image early even before starting to read it if we knew that the data is too short if this is easier to implement than catching the error later when reading.

Albrecht-S:
> PNGs interestingly are stored and read as RGBA data which should be changed to mirror JPEG and SVG.

Maybe this was implemented before we had the PNG image constructor that reads from memory?

Let's use the KISS principle. All image file types in fluid should just store the image in memory and call the "memory constructor" of the respective image type. That should likely simplify the fluid code (at least of the PNG image type). It would also save memory since the PNG would be stored in compressed format and decompressed at runtime.

FL/Fl_SVG_Image.H
```diff
@@ -147,14 +147,15 @@ private:
...
+  void init_(const char *name, const unsigned char *filedata, size_t length);
   Fl_SVG_Image(const Fl_SVG_Image *source);
 public:
   /** Set this to \c false to allow image re-scaling that alters the image aspect ratio.
    Upon object creation, proportional is set to \c true, and the aspect ratio is kept constant.*/
   bool proportional;
-  Fl_SVG_Image(const char *filename, const char *svg_data = NULL, size_t length = 0);
+  Fl_SVG_Image(const char *filename);
+  Fl_SVG_Image(const char *sharedname, const char *svg_data);
+  Fl_SVG_Image(const char *sharedname, const uchar *svg_data, size_t length);
```
Albrecht-S, line 158: We should IMHO be consistent WRT the type of image data. Is it char, unsigned char, or uchar. The latter needs a typedef (exists in our codebase) but should IMHO not be used for the public API. I'd prefer `const unsigned char *` for all image constructors.
Other image types should be checked and made consistent as well.

PS: Do we maybe need constructors for both signed and unsigned char, or do we rely on the compiler "converting" the type? Or would that risk compiler warnings?

src/Fl_SVG_Image.cxx
```diff
@@ -53,23 +54,76 @@ static double strtoll(const char *str, char **endptr, int base) {
...
-/** The constructor loads the SVG image from the given .svg/.svgz filename or in-memory data.
- \param filename Name of a .svg or .svgz file, or NULL.
- \param svg_data A pointer to the memory location of the SVG image data.
- This parameter allows to load an SVG image from in-memory data, and is used when \p filename is NULL.
- \param length When 0, indicates that \p svg_data contains SVG text, otherwise \p svg_data is
-  a  buffer of \p length  bytes containing GZ-compressed SVG data.
- \note In-memory SVG data is parsed by the object constructor and is not used after construction.
- When \p length > 0, parameter \p svg_data may safely be cast from data of type <em>const unsigned char *</em>.
+
+/** Load an SVG image froma file.
```
Albrecht-S, line 58: should read: "... from a ..." (mising space)

src/Fl_SVG_Image.cxx
```diff
@@ -90,143 +144,165 @@ float Fl_SVG_Image::svg_scaling_(int W, int H) {
...
+  while (chunk) {
+    next_chunk = chunk->next;
+    delete chunk;
+    chunk = next_chunk;
+  }
+
+  return err == Z_STREAM_END ? Z_OK :
+  err == Z_NEED_DICT ? Z_DATA_ERROR  :
+  err == Z_BUF_ERROR && stream.avail_out ? Z_DATA_ERROR :
+  err;
```
Albrecht-S, line 222: missing indentation

src/Fl_SVG_Image.cxx
```diff
@@ -90,143 +144,165 @@ float Fl_SVG_Image::svg_scaling_(int W, int H) {
...
-    if (fp) {
-      fseek(fp, 0, SEEK_END);
-      long size = ftell(fp);
-      fseek(fp, 0, SEEK_SET);
-      filedata = (char*)malloc(size+1);
-      if (filedata) {
-        if (fread(filedata, 1, size, fp) == size) {
-          filedata[size] = '\0';
+  // yes, this is a const cast to avoid duplicating user supplied data
+  uchar *data = (uchar*)in_data;
```
Albrecht-S, line 238: Then we should IMHO use `const_cast` explicitly in new code, according to recent discussion in fltk.general and the new CMP (will be updated very soon).

src/Fl_SVG_Image.cxx
```diff
@@ -90,143 +144,165 @@ float Fl_SVG_Image::svg_scaling_(int W, int H) {
...
+        data[length] = 0;
+      }
     }
-  } else if (copy_source) {
-    w(copy_source->w());
-    h(copy_source->h());
+  }
+#endif // HAVE_LIBZ
+
+  // now our SVG data should be in text format in `data`, terminated by a NUL
```
Albrecht-S, line 288: Sorry for this: but what if not? What if HAVE_LIBZ is not defined? It **should** always be defined, but ...

Albrecht-S:
Note: I just skimmed over the changes. My comments are not a real review, I commented just a few things I noticed.

I'll stop further actions until this is no longer a draft (to avoid wasting time), or until there's more progress, or ... ? Please let me know when you think it's ready to review.

FL/Fl_SVG_Image.H
MatthiasWM: Ok, no problem.

MatthiasWM: We *do* have signed char for text data and unsigned char plus length for compressed data. FLUID outputs it like that as well.

src/Fl_SVG_Image.cxx
MatthiasWM: Copied from zlib. I will see if I can indent that in a useful way.

src/Fl_SVG_Image.cxx
MatthiasWM: For good measure, I added an emoji too, since we do that now.

src/Fl_SVG_Image.cxx
MatthiasWM: In line 250, `ld` is preset with `ERR_FORMAT`. If we deliver compressed data without zlib present, `nsvgParse()` will error out and `ERR_FORMAT` will be returned.

MatthiasWM:
Thanks for looking through the code. I will apply the type changes and remove the expanding array on uncompressed image data readers, and return `ERR_FORMAT` instead (or a similar, maybe more meaningful error). I probably won't do this before Thursday though.

MatthiasWM:
> > PNGs interestingly are stored and read as RGBA data which should be changed to mirror JPEG and SVG.
> 
> Maybe this was implemented before we had the PNG image constructor that reads from memory?
> 
> Let's use the KISS principle. All image file types in fluid should just store the image in memory and call the "memory constructor" of the respective image type. That should likely simplify the fluid code (at least of the PNG image type). It would also save memory since the PNG would be stored in compressed format and decompressed at runtime.

Well, storing PNG data as RGBA in Fluid is a way to avoid adding image libraries to the final executable. I can't just remove that for back compatibility reasons. It must still be possible to inline a compressed PNG though. I'll get to FLUID later.

src/Fl_SVG_Image.cxx
MatthiasWM: I did add some error handling for compressed data and no zlib. It's much safer.

Albrecht-S:
> storing PNG data as RGBA in Fluid is a way to avoid adding image libraries to the final executable.

Good point, I didn't think of this consequence. So let's keep it.

erco77:
> +  return err == Z_STREAM_END ? Z_OK :
> +  err == Z_NEED_DICT ? Z_DATA_ERROR  :
> +  err == Z_BUF_ERROR && stream.avail_out ? Z_DATA_ERROR :
> +  err;
> 
> Copied from zlib. I will see if I can indent that in a useful way.

FWIW, I sometimes line up the `?` and `:` vertically this way to avoid walking off the right margin:

```
return (err == Z_STREAM_END)
         ? Z_OK
         : (err == Z_NEED_DICT)
           ? Z_DATA_ERROR
           : ((err == Z_BUF_ERROR) && stream.avail_out)
             ? Z_DATA_ERROR
             : err;
```

MatthiasWM:
> FWIW, I sometimes line up the `?` and `:` vertically this way to avoid walking off the right margin:

Aaaah, that looks much better. I copied that. Thanks.

src/Fl_Bitmap.cxx
```diff
@@ -33,6 +33,17 @@ void Fl_Bitmap::draw(int XP, int YP, int WP, int HP, int cx, int cy) {
 }
 
 
+/** The constructors create a new bitmap from the specified bitmap data.
+ If the provided array is to small to contain all the image data, the
+ constructor will not generate the bitmap to avoid illegale memeory read
```
Albrecht-S, line 38: s/to small/too small/
s/illegale memeory/illegal memory/

src/Fl_Bitmap.cxx
```diff
@@ -33,6 +33,17 @@ void Fl_Bitmap::draw(int XP, int YP, int WP, int HP, int cx, int cy) {
...
+ access and instead set \c data to NULL and \c ld to \c ERR_MEMORY_ACCESS.
+ \param bits bitmap data, one pixels per bit, rows are rounded to the next byte
```
Albrecht-S, line 40: s/pixels/pixel/

src/Fl_Bitmap.cxx
```diff
@@ -43,16 +54,26 @@ Fl_Bitmap::Fl_Bitmap(const uchar *bits, int bits_length, int W, int H) :
...
   }
-  data((const char **)&array, 1);
 }
 
 
+/** The constructors create a new bitmap from the specified bitmap data.
+ If the provided array is to small to contain all the image data, the
+ constructor will not generate the bitmap to avoid illegale memeory read
+ access and instead set \c data to NULL and \c ld to \c ERR_MEMORY_ACCESS.
+ \param bits bitmap data, one pixels per bit, rows are rounded to the next byte
```
Albrecht-S, line 71: See (three) proposed changes above (copy and paste)

src/Fl_Image.cxx
```diff
@@ -383,8 +383,8 @@ Fl_RGB_Image::Fl_RGB_Image(const uchar *bits, int W, int H, int D, int LD) :
 /**
  The constructor creates a new image from the specified data.
 
- If the provided array is to small to contain all the image data, the
- constructor will not generate the image to avoid illegale memeory read
+ If the provided array is too small to contain all the image data, the
+ constructor will not generate the image to avoid illegal memeory read
```
Albrecht-S, line 387: Sorry, "memeory" is still misspelled (at least three times in different files).

MatthiasWM: Sorry, bad muscle memeory... ;-) . Fixed.

MatthiasWM:
I decided to push these changes now. I feel that everything is complete on the image side of things, and that the FLUID implementation should be in a new PR, because it is a bit more involved than originally thought.

PR #594, opened 2022-12-13
Fl_Flex updates

Albrecht-S:
See discussion in fltk.coredev, thread ["Fl_Flex method name question"](https://groups.google.com/g/fltkcoredev/c/0pHCzc9m0eI/m/a5FUpouMBAAJ).

Please note that this branch also changes fluid in an incompatible way, i.e. besides changing the code I also changed the .fl file syntax. More below...

### DIFF
--- a/FL/Fl_Flex.H
+++ b/FL/Fl_Flex.H
@@ -78,8 +78,6 @@
   Fl_Flex containers can be nested so you can create flexible layouts with
   multiple columns and rows. However, if your UI design is more complex you
   may want to use Fl_Grid instead.
-  At the time of this writing (Aug 7, 2022) Fl_Grid is not yet available
-  but will be added before FLTK 1.4.0 gets released.
 
   Example:
   \image html Fl_Flex_simple.png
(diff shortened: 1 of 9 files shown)


Albrecht-S:
Further options we might want to consider (according to the thread in fltk.coredev):
- Allow fixed size `0` and use `-1` (negative) for changing a widget back to `flexible`
- add a size() method that returns the current width or height depending on type() (horizontal or vertical)
- add new methods:
  - `int flexible(Fl_Widget *) const` to query whether a widget is flexible or fixed
  - `void flexible(Fl_Widget *, int)` to set a widget flexible (1) or fixed (0)

I'm not really keen on adding these `flexible()` methods but anyway, here they are as an option.

Albrecht-S:
@MatthiasWM Regarding `fluid`: I changed
```diff
-  "set_size_tuples" <word> : this Word is written with printf "{%d", where %d
+  "fixed_size_tuples" <word> : this Word is written with printf "{%d", where %d
```
in `fluid/README_fl.txt` and in `fluid/Fl_Group_Type.cxx`. I did this for consistency but this is disputable. It would break existing `fluid` (`.fl`) files that have been created with the unreleased 1.4.0 fluid.

Albrecht-S:
@MatthiasWM Hmm, any more info than "thanks"? Maybe ...
- [ ] OK, yes, please merge
- [ ] I'll review later
- [ ] There are things I'd like to be changed ...

I know it's not easy to decide, sorry for asking for further info.

MatthiasWM:
Sorry, busy day at work today. I'll probably get to review this tonight.

MatthiasWM:
Thanks for also fixing FLUID and the comments. I like it.

Albrecht-S:
Welcome, thanks for checking. Merged.

Issue #596, opened 2022-12-15
Accelerate `Fl_Text_Display` and `Fl_Text_Buffer`


MatthiasWM:
This is a personal reminder to check if appending many short strings to `Fl_Text_Buffer` can be accelerated.

Also, when utf'ifying `Fl_Text_Display` I seem to remember that there was a way to reduce the time to calculate `Fl_Text_Display::warp()` substantially, but I never implemented that. So, now is the time (hint: binary search).

erco77:
See also related issue #300.

MatthiasWM:
Ok, so I dug through Fl_Text_Display, and I am sorry to say that I can't accelerate that without a wide ranging rewrite, and I wouldn't;t even know if the result is better under all circumstances. My acceleration idea works, but wasn't the time waster I suspected after all. Disabling `wrap()` is the one true accelerator, and if Fl_Simple_Terminal is limited to appending text, inserting a `\n` every 80 or so characters would give a typical terminal effect. But it is somewhat of a hack.

erco77:
@MatthiasWM Thanks for taking a look. Yeah, no one seems to want to touch Fl_Text_xxx stuff, lol.
I may have to do something like disable wrap() for the test programs, as some really are unusable due to the slowness of FST. test/tree and opening a large menu and then doing a large selection shows the problem, such as doing a shift-select over a hundred or so lines repeatedly.

erco77:
OK, the problem with disabling `wrap_mode()` is it causes Fl_Text_Display to force showing a horizontal scrollbar on the screen at all times (whether it needs one or not) once the vertical scrollbar appears.

Which in the case of the all the test apps is kinda a bad problem because the terminal windows at the default app size are necessarily small (just a few lines) the hscrollbar takes up a large percentage of the display and is kinda distracting.

I have an XXX comment in the FST constructor about this:
```
  :
  // XXX: We use WRAP_AT_BOUNDS to prevent the hscrollbar from /always/
  //      being present, an annoying UI bug in Fl_Text_Display.
  wrap_mode(Fl_Text_Display::WRAP_AT_BOUNDS, 0);
  :
```

I think I considered trying to hide the hscrollbar in FTD, but I don't think the widget is designed for that, so had to fallback to enabling wrap to prevent it.

Perhaps the code can be modified so that if the hscrollbar is hidden, it will not try to resize the text area, and can perhaps even save itself some calculation time.

MatthiasWM:
[...]

Issue #595, opened 2022-12-15
`Fl_Simple_Terminal` should not use `wrap()` mode, but instead wrap at a fixed boundary by inserting '\n'


MatthiasWM:
**Is your feature request related to a problem? Please describe.**
Fl_Text_Display, the base of Fl_Simple_Terminal, can become quite slow if wrap mode is activated. I am planning on improving that ASAP, but the basic problem will remain. Fl_Simple_Terminal can become relatively slow when large amounts of text (in particular, text without '\n' in them) are logged. This can be accelerated a lot if wrap() mode is disabled.

**I am using FLTK**
1.4.0

**Describe the solution you'd like**
 I suggest to wrap automatically at character 80, or at a user-settable character x position, because Fl_Simple_Terminal will likely use a fixed font and likely log output from other apps. It should still be possible to enable wrap mode if needed.

**Describe alternatives you've considered**
Qt, but FLTK is so much nicer. And smaller. And without license fees. And the developers are just awesome.

_[Edit by @Albrecht-S: fixed a few typos.]_

Albrecht-S:
I'm not sure if I understand the request correctly. I would not want to **insert** `\n` in the actual data (FLTK should not change user data) but I agree that the **display** should (or could) wrap at a given position.

erco77:
@MatthiasWM -1 for inserting \n's or in any way changing the user's text, aside from ansi() parsing and special character handling like backspace (I'm recently adding this btw).

I think it's important that long lines wrap for display purposes only, and unwrap when the widget is made larger, so long output can be read continuously. It'd be OK to have clipping be an option and a horizontal scrollbar appear only if anything clips. (Not sure if the base widget handles that well tho -- I seem to have memory of the horiz scrollbar appearing when it shouldn't under certain circumstances)

erco77:
BTW, it appears I'll be modifying FST's ansi() parsing code quite a bit to support some new features that are needed (my recent comment sin issue #577), so if we're both working on that code, there might be conflicts.

Issue #300, opened 2021-11-27
test/tree: hitting 'select all' hangs the application


erco77:
Investigating..

erco77:
Hmm, it does recover after about 30 or more seconds.

Seems to be the Fl_Simple_Terminal() code taking a really long time to handle all the info print messages triggered by select all (there's 500 items in one of the tree's children) Still, that shouldn't be a problem.

Possibly platform specific; this is linux, and according to gdb, apparently xft calls are involved during the printf() operations, probably doing label measurements. Still investigating.

erco77:
Yikes, yeah, this is actually a problem in Fl_Simple_Terminal.

I modified examples/simple-terminal.cxx to just print 300 dots and a crlf every second, e.g.
```
    for ( int t=0; t<300; t++ ) G_tty->printf("."); 
    G_tty->printf("\n");
```
..and that's causing really large delays in e.g. resizing and such.

This _doesn't_ seem to be xft specific; even rebuilding FLTK with --disable-xft still causes severe slowness.

erco77:
Hmm, I copied Fl_Simple_Terminal's code over to 1.3.x (this widget is new in 1.4.x), and am /not/ seeing slowness, so perhaps something has changed in FLTK 1.4.x that's slowing down drawing text.

I checked OSX, and I see the problem there too, but to a lesser extent. On an old mac mini, I'm seeing a ~10 sec hang using test/tree and hitting the "Select All" button.

So this seems to be a cross platform issue.

Might need help with this one -- has anything changed in 1.4.x recently that might affect text drawing speed? I'll try to determine if this is a regression with bisect.. might take a while.

erco77:
Yeesh, I've gone back to a commit from late 2017 and still seeing slowness with test/tree "Select All" so it's not a recent thing. But it's still more efficient somehow in 1.3.x?

Decided not to go back any further in 1.4.x and see if I can optimize Fl_Simple_Terminal's operation. It inherits from Fl_Text_Display which is a touchy beast, so looking to design around Fl_Text_Display's behavior.

ManoloFLTK:
Adding
`tty->wrap_mode(Fl_Simple_Terminal::WRAP_NONE, 0);`
after line #2069 of file test/tree.cxx (after it was generated by fluid),
has the slowness disappear.

Thus, the slowness seems connected to computations that support line wrapping 
repeated each time a line is written to the terminal. And, this gives at least a
temporary fix for the issue.

[...]

PR #603, opened 2022-12-17
Added support for ESC sequences to straddle append()

erco77:
This is basically a rewrite of the ESC sequence handler, so as
    to keep state information within class, so that ESC sequences
    can be continued between calls to `append()`, `printf()` et al.

    A new private class, `Fl_Escape_Seq` handles parsing character-at-a-time,
    maintaining state info. It also has careful bounds checking during parsing.

    This includes the new backspace handling, weird sequences like `ESC[;m`,
    and private variables renamed from `_xxx` to `xxx_` as per the CMP.

### DIFF
--- a/FL/Fl_Simple_Terminal.H
+++ b/FL/Fl_Simple_Terminal.H
@@ -137,12 +137,12 @@ private:
     static const int completed = 1;   // multi-step operation completed successfully
   private:
     char  esc_mode_;        // escape parsing mode state
-    char  buf[maxbuf];      // escape sequence being parsed
-    char *bufp;             // parsing ptr into buf[]
-    char *bufendp;          // end of buf[] (ptr to last valid buf char)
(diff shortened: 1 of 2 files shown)


erco77:
This does NOT address any of the slowness issues that AFAIK are with the base classes, and are a side issue still open as related issues #300 (which has I think the most analysis of the slowness issue so far), and new related issues from matt: #595 and #596.

MatthiasWM:
> This branch does NOT address any of the slowness issues that AFAIK are with the base classes, and are a side issue still open as related issues #300 (which has I think the most analysis of the slowness issue so far), and new related issues from matt: #595 and #596.

Thanks, yes, that's a completely separate issue. I will look at that in the next days.

erco77:
Closing this pull request, committed to master directly.

PR #604, opened 2022-12-17
Improving FLUID support for inline image data (see #542, #592)

MatthiasWM:
$542: image readers need to use size argument to avoid illegal memory access
$592: Adding length checks for in-memory image data (see #542) #592

FLUID support for length in binary formats, choice of original or raw format
- [x] `Fl_SVG_Image` : works in compressed mode and uncompressed
- [x] `Fl_GIF_Image` : works compressed and uncompressed
- [x] `Fl_BMP_Image` : works compressed and uncompressed
- [x] `Fl_XBM_Image` : inlines raw Bitmap data which is probably an ok choice
- [x] `Fl_JPEG_Image` : works compressed and uncompressed
- [x] `Fl_PNG_Image` : works compressed and uncompressed
- [x] `Fl_PNM_Image` : is not a registered image format
- [x] `Fl_ICO_Image` : so far not supported by FLUID
- [x] `Fl_Pixmap_Image` : always prints its text version

Error codes:
- [x] verify that all `ld()` error codes are meaningful, probably need to add some


MatthiasWM:
I did not change `ld()` existing error codes. When reading compressed formats from memory, the error will still be `ERR_FILE_ACCESS` instead of `ERR_MEMORY_ACCESS` because it's consistent to see compressed image data as an in-memory dump of a file. `ERR_MEMORY_ACCESS` stays reserved for uncompressed direct pixel access fails.

MatthiasWM:
Ok, so we have a new option in FLUID that allows us to store images in their native, given, compressed format inside a source code file, which saves a ton of space, but requires linking with fltk_images, and we have the option to save uncompressed RGB, Pixmap, or Bitmap data, depending on the source file format.

Albrecht-S:
> ... [storing compressed images] ... requires linking with fltk_images ...

This is a great new feature! I like it.

Would it also require to call `fl_register_images()` at runtime? If yes, this should be documented (or shown in a tooltip, or whatever).

MatthiasWM:
Thank. Yes, it must be documented in the FLUID handbook. Before the patch, JPEG was always stored compressed, and PNG always as RGB, so I hope for not too many surprises.

PR #609, opened 2022-12-21
Mark all overridden methods with `override`

MatthiasWM:
- [x] libraries
- [x] FLUID
- [x] tests

- [x] macOS
- [ ] X11
- [ ] Wayland
- [ ] MSWindow


MatthiasWM:
Bill suggests to use `override` instead of `FL_OVERRIDE`
Also suggests to shorten the #if/else even more

Albrecht-S:
I believe this PR is obsolete, the current version work is done in PR #611. Is this correct?

If yes, can we close this PR?

PR #611, opened 2022-12-22
Alternative format for `override`

MatthiasWM:
This version uses `void draw() override` instead of `virtual void draw() FL_OVERRIDE`.


MatthiasWM:
Ah well, we actually have a function named `Fl_Window::override()`, so using `#define override` will not work. So we can go with `#define fl_override override`, or just assume that all C compilers today support `override` and not go with any `#define`.

Albrecht-S:
I'm for keeping `FL_OVERRIDE` (uppercase) because it is clear that it is a macro. Using a `#define` for `'override'` sounds like a bad idea anyway because `#define` is not context sensitive (as has been shown with our own override method. It would also affect user code which is not what we want.

MatthiasWM:
Ok, so VisualStudio neither sets _cplusplus to anything meaningful, nor does the compiler care much if virtual or override or nothing is set, not even if I try the MS version of Clang for ARM64. So I can't test or fix this for Windows. I will give it a few hours of the other devs replies and then probably apply this PR.

CHANGES_1.1.txt
```diff
@@ -2449,7 +2449,7 @@ CHANGES IN FLTK 1.1.0b4                                 BETA: ??? ?? ????
 
   - Updated the flags_ member of Fl_Widget to be an
     integer instead of uchar, to support the new
-    FL_OVERRIDE flag for Fl_Window.
+    OVERRIDE flag for Fl_Window.
```
Albrecht-S, line 2452: Please revert this change.
This file (CHANGES_1.1.txt) should NOT be changed, it's for reference only.

Albrecht-S:
@MatthiasWM I'm testing this branch.

I found more methods that should be marked override (`[-Wsuggest-override]`):

- a total of about 200 methods in my Linux (X11 + Wayland builds)
- 326 methods in Windows build (cross-compiled under Linux)

I can post a list with warnings or I can try to help and fix (at least the Windows) warnings myself.

How do you want to proceed?

Albrecht-S:
FYI: I combined all warnings in one file, sorted by method name. I had to remove the source file name and line numbers, but the sorted list contains each method name only once (`sort -u`) with a total of 529 lines methods.

[override_sorted.txt](https://github.com/fltk/fltk/files/10298825/override_sorted.txt)

If you want me to help with the Windows methods I could do it by cross-compiling on Linux which should find all Windows specific overrides w/o using Visual Studio!

Albrecht-S:
FYI: I started looking into the Windows implementation of `FL_OVERRIDE` and I confirm that Visual Studio doesn't define `__cplusplus` to a sensible value (it's `199711L`).

I believe we can use `_MSC_VER` instead, which defines the Visual Studio version number. I made a patch that uses this to define `FL_OVERRIDE` = `override` starting with **Visual Studio 2015** which is an arbitrary but IMHO sensible value. It doesn't matter much anyway, but I would appreciate if we used `override` in the latest Visual Studio versions.

I propose to apply the attached patch [override2_attr.diff.txt](https://github.com/fltk/fltk/files/10300403/override2_attr.diff.txt).

Note: this patch also fixes some `'//'` comments in parts that might be compiled with a C compiler. I left `'//'` comments in C++ parts (where `__cplusplus` is defined).

I tested this patch with Visual Studio 2019 and I can at least confirm that it issues an error message if `FL_OVERRIDE` is applied to a **_non-virtual_** method.

Testing now with mingw-w64 (MSYS2) on Windows...

CHANGES_1.1.txt
MatthiasWM: Oha, I just downloaded 1.1 and the flag was indeed FL_OVERRIDE. I did not remember that and thought I fixed old docs.

MatthiasWM:
I applied stuff for Linux and your patch. Thanks very much for that. Now I will give Windows a shot.

MatthiasWM:
Ok, I think that this is complete now.

MatthiasWM:
Thanks for revisiting this. Sorry for generating additional work for you. I did find one single false `virtual` keyword in existing code (not a big deal at all, but at least the system is working).

Albrecht-S (approved):
OK, I believe that this branch is complete and OK, as far as I can tell, if my two commits to your fork are included.

Built on Windows (in Virtualbox) with MinGW (MSYS2) and Visual Studio 2019.

Albrecht-S:
There's only one minor issue in one of the commit messages (sorry, nitpicking):
"Define FL__OVERRIDE for VisualC > 2015" should read "... >= 2015" (VS 2015 is included).

I assume you'll see this line when you squash/merge on GitHub and can edit it in the commit message of the squashed commit.

Please feel free to go ahead and squash/merge, and thank you **VERY** much for doing this mass edit!

Albrecht-S:
Oh, and please edit (remove) obsolete lines from the final commit message like that of my last commit "Restore 'override' in bundled png library" which will be useless when all the commits are squashed because "it never happened". ;-)

PR #620, opened 2022-12-30
Remove commented fl_beep calls

chyla:
I don't think so the code would be needed some day. Commented out 11 years ago with 1fba7d0f4694fdabc3db67f489d9e3e3956035da


Albrecht-S:
Thanks for the PR, although this is not strictly something necessary. However I agree that these fl_beep calls will probably never be needed again. I'll merge it, and this will close the PR.

Issue #612, opened 2022-12-23
Build failure case of FLTK with MSYS2+MinGW-W64


rageworx:
I am not sure is it a bug, maybe failure by environment, but anyway it happended, so I report to here.
In case of auto configured with ./autogen.sh, fltk/fltk fails to bilbuild libfltk_png.a by libz reference of MinGW64.
```
Configuration Summary
-------------------------------------------------------------------------
    Directories: prefix=/mingw64
                 bindir=${exec_prefix}/bin
                 datadir=${datarootdir}
                 datarootdir=${prefix}/share
                 exec_prefix=${prefix}
                 includedir=${prefix}/include
                 libdir=${exec_prefix}/lib
                 mandir=${datarootdir}/man
                 MSys docpath=/local/share/doc/fltk
Cross-compiling: no
       Graphics: GDI+
Image Libraries: JPEG=System
                 PNG=Builtin
                 ZLIB=System
    Large Files: YES
         OpenGL: YES
        Threads: YES
configure: creating ./config.status
config.status: creating makeinclude
config.status: creating fltk.list
config.status: creating fltk-config
config.status: creating fltk.spec
config.status: creating FL/Makefile
config.status: creating config.h
config.status: creating FL/fl_config.h
```
And build stops as ...
```
=== making png ===
make[1]: Entering directory '/d/Projects/fltk-origin/png'
make[1]: warning: -j4 forced in submake: resetting jobserver mode.
Compiling png.c...
Compiling pngset.c...
Compiling pngget.c...
Compiling pngrutil.c...
In file included from pngrutil.c:17:
pngpriv.h:920:4: error: #error ZLIB_VERNUM != PNG_ZLIB_VERNUM "-I (include path) error: see the note
s in pngpriv.h"
  920 | #  error ZLIB_VERNUM != PNG_ZLIB_VERNUM \
      |    ^~~~~
In file included from pngget.c:15:
pngpriv.h:920:4: error: #error ZLIB_VERNUM != PNG_ZLIB_VERNUM "-I (include path) error: see the note
s in pngpriv.h"
  920 | #  error ZLIB_VERNUM != PNG_ZLIB_VERNUM \
      |    ^~~~~
In file included from png.c:14:
pngpriv.h:920:4: error: #error ZLIB_VERNUM != PNG_ZLIB_VERNUM "-I (include path) error: see the note
s in pngpriv.h"
  920 | #  error ZLIB_VERNUM != PNG_ZLIB_VERNUM \
[...]

PR #621, opened 2022-12-30
Make sure that zlib and png are both local or system

MatthiasWM:
Implemented in autoconf and CMake.

This checks if libs and png system libraries are requested and available. If any of the checks fail, we fall back to the local version, so that version compatibilities are maintained. #612 .

### DIFF
--- a/CMake/options.cmake
+++ b/CMake/options.cmake
@@ -103,10 +103,10 @@ if (OPTION_USE_SYSTEM_LIBPNG)
   find_package (PNG)
 endif ()
 
-# If we use the system zlib, we must also use the png zlib and vice versa
-# If either of them if not available, we fall back to using both local libraries
+# If we use the system zlib, we must also use the system png zlib and vice versa
+# If either of them is not available, we fall back to using both local libraries
 if (OPTION_USE_SYSTEM_LIBPNG AND NOT (OPTION_USE_SYSTEM_ZLIB AND ZLIB_FOUND))
-    set (PNG_FOOUND FALSE)
+    set (PNG_FOUND FALSE)
     set (OPTION_USE_SYSTEM_LIBPNG FALSE)
     message (STATUS "Local z lib selected: overriding png lib to local for compatibility.\n")
 endif ()
--- a/CMake/options.cmake
+++ b/CMake/options.cmake
@@ -92,12 +92,33 @@ else ()
 endif ()
 
 #######################################################################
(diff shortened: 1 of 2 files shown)


CMake/options.cmake
```diff
@@ -92,13 +92,34 @@ else ()
...
 if (OPTION_USE_SYSTEM_ZLIB)
   find_package (ZLIB)
 endif ()
 
+if (OPTION_USE_SYSTEM_LIBPNG)
+  find_package (PNG)
+endif ()
+
+# If we use the system zlib, we must also use the png zlib and vice versa
+# If either of them if not available, we fall back to using both local libraries
```
Albrecht-S, line 107: TYPO: ... **is** not available ...

CMake/options.cmake
```diff
@@ -92,13 +92,34 @@ else ()
...
+if (OPTION_USE_SYSTEM_LIBPNG AND NOT (OPTION_USE_SYSTEM_ZLIB AND ZLIB_FOUND))
+    set (PNG_FOOUND FALSE)
```
Albrecht-S, line 109: TypOO, should read: PNG_FOUND

configure.ac
```diff
@@ -727,7 +704,44 @@ AS_IF([test x$enable_localpng != xyes], [
     ])
 ])
 
-# The following is executed if the lib was not found usable or if local lib is required explicitly
+# If we use the system zlib, we must also use the png zlib and vice versa
```
Albrecht-S, line 707: as above: "we must also use the **system png lib**"

configure.ac
```diff
@@ -727,7 +704,44 @@ AS_IF([test x$enable_localpng != xyes], [
...
+# If either of them if not available, we fall back to using both local libraries
```
Albrecht-S, line 708: s/if/is/ (as above)

Albrecht-S:
@MatthiasWM I just looked quickly over the changes and flagged some typos, one of which would be relevant (code), the others being comments only.

FYI: I can't tell if these CMake mods are correct, and I didn't read/check the configure changes.

Albrecht-S (approved):
CMake code looks good.
Tested with cmake-gui, changing options, worked as expected.

I didn't test the configure part though, leaving this up to others.

MatthiasWM:
Thanks for checking this. I will have a another round of testing with autoconf tomorrow.

PR #624, opened 2023-01-01
Add Fl_Scheme_Choice widget

Albrecht-S:
This widget offers the choice of all known FLTK schemes as a simple widget based on `Fl_Choice`.

Note: initial version, needs more work (WIP)

Todo:
 - [x] redraw all open windows when the scheme is changed
 - [ ] add feature to add new schemes -- postponed --
 - [ ] update status (value) when the scheme is changed by `Fl::scheme("...")` -- postponed --
 - [x] update `Fluid`: replace `Fl_Choice` with `Fl_Scheme_Choice` to remove hardcoded scheme names
 - [x] make sure `Fluid` doesn't open the display in batch mode
 - [x] provide a list of known scheme names (suggested by @erco77 below)

Future:
 - [ ] register schemes to auto-update `Fl_Scheme_Choice` widget
 - [ ] maybe more...

Edit (Jan 04, 2023): I changed the simple lists (bullets) to task lists and edited some items for clarity.

### DIFF
--- a/fluid/alignment_panel.fl
+++ b/fluid/alignment_panel.fl
@@ -38,6 +38,9 @@ decl {\#include <FL/Fl_Text_Display.H>} {public local
 decl {\#include <FL/filename.H>} {public local
 }
 
+decl {\#include <FL/Fl_Scheme_Choice.H>} {public local
+}
+
 decl {\#include <FL/Fl_Preferences.H>} {private global
 }
 
@@ -47,11 +50,14 @@ decl {\#include <FL/fl_ask.H>} {private global
 decl {\#include <string.h>} {private global
 }
 
+decl {void init_scheme(void);} {
+  comment {// initialize the scheme from preferences} selected public global
+}
+
 decl {extern struct Fl_Menu_Item *dbmanager_item;} {public local
 }
 
-Function {make_project_window()} {open
-} {
+Function {make_project_window()} {} {
   Fl_Window project_window {
     label {Project Settings} open
     xywh {472 246 399 298} type Double hide
@@ -168,50 +174,23 @@ set_modflag(-1, -1);}
 decl {extern void i18n_cb(Fl_Choice *,void *);} {public local
 }
 
-decl {void scheme_cb(Fl_Choice *, void *);} {public local
+decl {void scheme_cb(Fl_Scheme_Choice *, void *);} {public local
 }
 
 Function {make_settings_window()} {open
 } {
   Fl_Window settings_window {
-    label {GUI Settings}
-    xywh {442 538 360 355} type Double hide resizable
+    label {GUI Settings} open
+    xywh {701 666 360 355} type Double hide resizable
     code0 {o->size_range(o->w(), o->h());} non_modal
   } {
     Fl_Choice scheme_choice {
       label {Scheme: }
       callback scheme_cb open
       xywh {140 10 115 25} down_box BORDER_BOX labelfont 1
-      code0 {int s;}
-      code1 {fluid_prefs.get("scheme", s, 0);}
-      code2 {scheme_choice->value(s);}
-      code3 {scheme_cb(0, 0);}
-    } {
-      MenuItem {} {
-        label Default
-        xywh {0 0 35 25}
-      }
-      MenuItem {} {
-        label None
-        xywh {0 0 35 25}
-      }
-      MenuItem {} {
-        label Plastic
-        xywh {0 0 35 25}
-      }
-      MenuItem {} {
-        label {GTK+}
-        xywh {10 10 35 25}
-      }
-      MenuItem {} {
-        label Gleam
-        xywh {20 20 35 25}
-      }
-      MenuItem {} {
-        label Oxy
-        xywh {30 30 35 25}
-      }
-    }
+      code0 {init_scheme();}
+      class Fl_Scheme_Choice
+    } {}
     Fl_Group {} {open
       xywh {20 43 330 161} labelfont 1 align 0
     } {
@@ -327,7 +306,7 @@ Function {make_shell_window()} {open
         tooltip {save the design to the .fl file before running the command} xywh {82 39 136 19} down_box DOWN_BOX labelsize 12
       }
       Fl_Check_Button shell_writecode_button {
-        label {save source code} selected
+        label {save source code}
         tooltip {generate the source code and header file before running the command} xywh {82 59 120 19} down_box DOWN_BOX labelsize 12
       }
       Fl_Check_Button shell_writemsgs_button {
--- a/test/boxtype.cxx
+++ b/test/boxtype.cxx
@@ -19,6 +19,7 @@
 #include <FL/Fl.H>
 #include <FL/Fl_Double_Window.H>
 #include <FL/Fl_Box.H>
+#include <FL/Fl_Scheme_Choice.H>
 #include <FL/fl_draw.H>
 
 int N = 0;
@@ -176,6 +177,7 @@ int main(int argc, char ** argv) {
   bt("FL_GTK_ROUND_DOWN_BOX",FL_GTK_ROUND_DOWN_BOX);
   bg->end();
   window->resizable(window);
+  Fl_Scheme_Choice scheme_choice(610, 10, 150, 30, "Scheme:");
   window->end();
   window->show();
   return Fl::run();
--- a/test/unittest_schemes.cxx
+++ b/test/unittest_schemes.cxx
@@ -18,7 +18,7 @@
 
 #include "unittests.h"
 
-#include <FL/Fl_Choice.H>
+#include <FL/Fl_Scheme_Choice.H>
 
 // needed by Edmanuel's test layout
 #include <FL/Fl_Button.H>
@@ -45,16 +45,7 @@
 #include <FL/Fl_Radio_Round_Button.H>
 
 class Ut_Schemes_Test : public Fl_Group {
-  Fl_Choice *scheme_choice_;
-
-  static void scheme_choice_cb(Fl_Widget*,void *data) {
-    Ut_Schemes_Test *st = (Ut_Schemes_Test*)data;
-    const char *name = st->scheme_choice_->text();
-    if ( name ) {
-      Fl::scheme(name);         // change scheme
-      st->window()->redraw();   // redraw window
-    }
-  } // SchemeChoice_CB()
+  Fl_Scheme_Choice *scheme_choice_;
 
   static void activate_subwin(Fl_Widget *w, void *data) {
     Fl_Window *win = (Fl_Window *)data;
@@ -82,22 +73,9 @@ public:
   }
   Ut_Schemes_Test(int X,int Y,int W,int H)
     : Fl_Group(X,Y,W,H) {
-    scheme_choice_ = new Fl_Choice(X+125,Y,140,25,"FLTK Scheme");
-    scheme_choice_->add("none");
-    scheme_choice_->add("plastic");
-    scheme_choice_->add("gtk+");
-    scheme_choice_->add("gleam");
-    scheme_choice_->add("oxy");
-    scheme_choice_->value(0);
+
+    scheme_choice_ = new Fl_Scheme_Choice(X+125,Y,140,25,"FLTK Scheme");
     scheme_choice_->labelfont(FL_HELVETICA_BOLD);
-    const char *name = Fl::scheme();
-    if ( name ) {
-           if ( strcmp(name, "plastic") == 0) { scheme_choice_->value(1); }
-      else if ( strcmp(name, "gtk+")    == 0) { scheme_choice_->value(2); }
-      else if ( strcmp(name, "gleam")   == 0) { scheme_choice_->value(3); }
-      else if ( strcmp(name, "oxy")     == 0) { scheme_choice_->value(4); }
-    }
-    scheme_choice_->callback(scheme_choice_cb, (void*)this);
 
     Fl_Light_Button *active = new Fl_Light_Button(X + 300, Y, 100, 25, "active");
     active->value(1);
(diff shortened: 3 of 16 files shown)


erco77:
I'd be fine with just a method that returns a string array of the available scheme names, e.g. `const char **Fl::schemes()` or some such, so I can work them into a menu bar's preference choices.

Albrecht-S:
The intention of this PR (widget Fl_Scheme_Choice) is
1. remove hardcoded scheme names from user code (where currently scheme selection is possible)
2. enable users to add a scheme choice widget easily w/o hardcoding scheme names
3. increase maintainability

The diff stats in user code (folder `test`) show the gain (less code):
```
$ git diff --stat origin/master.. -- test/
 test/demo.cxx             | 24 +++++-------------------
 test/unittest_schemes.cxx | 31 +++++--------------------------
 2 files changed, 10 insertions(+), 45 deletions(-)
```

@MatthiasWM @erco77 @ManoloFLTK This branch (PR) is work in progress, I'll add more features (see commit comment) but I'd appreciate your comments:
- what can or should be added?
- what can be done better?
- did I miss anything?

Thanks in advance.

Albrecht-S:
@erco77 Thanks for your suggestion, that's something I thought about as well, and I believe this is easy to add. I'll check this and update the branch.

The reason I made this widget was that we have such an `Fl_Choice` in several places (see `demo.cxx` and `unittest_schemes.cxx`) and I needed a quick way to add such a scheme choice widget for testing. With this branch it's just a two-liner to add a new `Fl_Scheme_Choice` widget w/o coding scheme names and writing a callback and more.

PS: registering schemes is a future option (I'm still planning to create a `Fl_Scheme_*` hierarchy) and that's likely the point when such a register function makes sense. However, that's not planned for 1.4.0.

MatthiasWM:
Ok, here are a few observation. Nothing major, the code seems fine the way it is, but...
 - instead of "None" I would suggest "Classic". FO complete novice users, "None" has no meaning, but "Classic" IMHO reflects the default WIndows 95 look very well
 - a name list is useful. Maybe a NULL terminated list of menuitems is a good alternative, because programmers can integrate that into their main menu by simply linking it as a submenu, especially if it has already working callbacks attached. And it also provides a name list. But then those should probably be radio items. OTOH, it's only five lines or so if a scheme name list is available

Albrecht-S:
@MatthiasWM Thanks for your suggestions.

The list of possible scheme names is documented: we have the choice of "none" or "base" for the default scheme, but "classic" has never been defined AFAICT. I chose "none" because this is what I'm used to and what we have in the existing scheme choice widgets. I didn't want to "invent" yet another name.

Regarding name list, menu items, callback: I've been thinking about integrating scheme choices in menus as well but I don't have a concrete API in mind. This would come in handy for users that want to add the scheme selection to the main menu like `{File|Edit}/Preferences/Schemes/<submenu of all schemes>`. Proposals welcome.

Albrecht-S:
Commit df07cc03c7aba83b3a59b45a86fa2ebc64771087 adds the list of available schemes and updates (redraws) **all open windows** in the `Fl_Scheme_Choice` callback (the latter is untested). This is still work in progress.

Note: the method to get the list of scheme names is currently `Fl_Scheme_Choice::schemes()` which is maybe not the final name. I left it under the `Fl_Scheme_Choice::` namespace for now because I didn't want to add more to the `'Fl'` class.

Exampe code in test/demo.cxx to output the known scheme names:
```
  const char * const * scheme_names = Fl_Scheme_Choice::schemes();
  int ni = 0;
  while (scheme_names[ni]) {
    printf("scheme[%2d] = '%s'\n", ni, scheme_names[ni]);
    ni++;
  }
```
outputs:
```
scheme[ 0] = 'none'
scheme[ 1] = 'plastic'
scheme[ 2] = 'gtk+'
scheme[ 3] = 'gleam'
scheme[ 4] = 'oxy'
```

MatthiasWM:
> The list of possible scheme names is documented: we have the choice of "none" or "base" for the default scheme, but "classic" has never been defined AFAICT. 

True. It was just intuition. Maybe Base is still better than None? Not sure.

> Regarding name list, menu items, callback: I've been thinking about integrating scheme choices in menus as well but I don't have a concrete API in mind. This would come in handy for users that want to add the scheme selection to the main menu like `{File|Edit}/Preferences/Schemes/<submenu of all schemes>`. Proposals welcome.

Exactly. It could be a simple as `static Fl_Menu_item* Fl_Scheme_Choice::menu_items();`.

Albrecht-S:
FTR: I updated the branch and force-pushed it. You need to update your branch and pull the changes if you like to test.

Status report:

I renamed `Fl_Scheme_Choice::schemes()` to `Fl_Scheme_Choice::scheme_names()` but meanwhile I'm thinking about putting it into `Fl` as `Fl::scheme_names()`. In the future this will likely be `Fl_Scheme::names()` or similar. I'm still investigating whether adding a minimal `Fl_Scheme` class is useful, particularly with respect to ABI stability during 1.4.x versions.

The new `handle()` method allows the `Fl_Scheme_Choice` widget to initialize its current `value()` **_when shown_**, i.e. after or "inside" `Fl_Window::show(argc, argv)` - after the args are parsed and the scheme is set - which avoids the need to call `init_value()` in user code.

I added `Fl_Scheme_Choice` to these test programs:
- test/boxtype.cxx
- test/buttons.cxx
- test/menubar.cxx

The following test programs have been updated to use the new widget:
- test/unittest_schemes.cxx
- test/demo.cxx

This new feature helped me testing button drawing and boxtypes and I found some issues which I could already fix. More to come...

Albrecht-S:
> Maybe Base is still better than None? Not sure.

Neither am I:
- 'none' is good because it's the default, standard scheme, i.e. **no scheme** has been set
- 'base' is also good because it's the `basic scheme`. In the future this may be implemented in the base class Fl_Scheme or in Fl_Scheme_Base or Fl_Scheme_None. I can't tell yet.

> [ a menu API ] ... could be as simple as `static Fl_Menu_item* Fl_Scheme_Choice::menu_items();`

Yes, that might do it. Thinking further about it...

erco77:
On 1/2/23 11:39, Matthias Melcher wrote:

>     The list of possible scheme names is documented: we have the choice of "none" or "base" for the default scheme, but "classic" has never been
>     defined AFAICT.
>
> True. It was just intuition. Maybe Base is still better than None? Not sure.
>
     "default" comes to mind.

Albrecht-S:
> >  Maybe Base is still better than None? Not sure.

> "default" comes to mind.

In fact all **unknown** scheme names select the 'default' / 'base' / 'none' scheme, i.e. you can **use** whatever you like in your code (as long as this name is not added later as another **real** scheme name). For instance:
```C++
  Fl::scheme("unknown");
```
would select the default scheme.

However, which name do we set internally? This name would be returned by `Fl::scheme()` and could lead to "surprises" in user code. I just checked it and it is clear that `<null>` is returned if the default scheme is set, even if it was set by `Fl::scheme("base")` (which is good). It's not clearly documented that setting scheme 'none' will return `<null>` but we can easily add this to the docs.
https://www.fltk.org/doc-1.4/classFl.html#a9d070bb1d5a24c28beab53a70af095a4

Another question we'd need to answer is which one we want to use as the "official" name that would be presented (to the user) in the `Fl_Scheme_Choice` widget and menus. This should then be documented and that's it. It would be good though to use one of the already documented names.

Personaly I'm open for a new name like `'default'` as the "official" scheme name. The documentation could keep the old names as deprecated but they would still work in existing code.

Summary:
- `'none'` is kind of a misnomer (because it **is** a scheme) and should maybe be replaced by ...
- `'base'`: this would be a good replacement because it is already documented
- `'default'`: (suggested by @erco77): would IMHO be another good choice (although it's a little longer ;-) )
- `'classic'`: (suggested by @MatthiasWM): would be yet another option

My favorite names would be `'base'` and `'default'`.

Comments, anybody? Shall we discuss this (and vote) in fltk.coredev?

MatthiasWM:
You TODO has an "add to FLUID". Does that mean adding `Fl_Scheme_Choice` as something the user can add to their project? Or as a quick access to change the FLUID scheme while designing? It's in the app settings, but you have to open a dialog box for that.

Albrecht-S:
> You TODO has an "add to FLUID". Does that mean adding Fl_Scheme_Choice as something the user can add to their project?

No, this meant to replace the existing code with `Fl_Scheme_Choice` to get rid of the hardcoded scheme names.
See code around [fluid/fluid.cxx, line 1476++](https://github.com/fltk/fltk/blob/d3988f9bdcd3cf463113b43fb0f1f22724e69f42/fluid/fluid.cxx#L1476) and whatever is needed around this.

This should be easy (similar to `test/demo` and `unittests` which I did already in my branch). I can do the same in fluid when `Fl_Scheme_Choice` is committed.

I don't think that we need direct `Fl_Scheme_Choice` support in fluid for projects. For the developer it could be as easy as using `Fl_Choice` and changing the class to `Fl_Scheme_Choice`, right? Or maybe we can add it as a separate widget type (if you want to do this).

MatthiasWM:
> ...replace the existing `Fl_Choice` widget with `Fl_Scheme_Choice` to get rid of the hardcoded scheme names. See code around [fluid/fluid.cxx, line 1476++]

Oh, sure, no problem. I'd like to wait for you to push to master.

> I don't think that we need direct `Fl_Scheme_Choice` support in fluid for projects. For the developer it could be as easy as using `Fl_Choice` and changing the class to `Fl_Scheme_Choice`, right? Or maybe we can add it as a separate widget type (if you want to do this).

I think that typing `Fl_Scheme_Choice` in the Class filed of any `Fl_Choice` should be enough and can be expected by the user. Maybe I can add it under "Other".

Albrecht-S:
FTR: the latest two commits
- added a basic `Fl_Scheme` class which now provides `Fl_Scheme::names()`, the list of known scheme names
- `Fluid` now uses the new `Fl_Scheme_Choice` widget
- `Fluid` now stores the scheme preferences under the new key `scheme_name` as a character string
- backwards compatibility (reading and writing old index from key `scheme`) is maintained

Note that the Fl_Scheme class is work in progress. There are some unused declarations and probably unused code. This will be updated later. The current goal was to support `Fl_Scheme::names()` and to be open for further extensions.

@MatthiasWM I'd appreciate if you would check the fluid changes. Thanks in advance.

fluid/alignment_panel.fl
```diff
@@ -47,11 +50,14 @@ decl {\#include <FL/fl_ask.H>} {private global
 decl {\#include <string.h>} {private global
 }
 
+decl {void init_scheme(void);} {
```
MatthiasWM, line 53: Just as an aside, you can actually put the `#include` statement into the **Code** section in the widget dialog. FLUID positions it correctly in the source for you, keeping things together. If you'd ever delete the widget, you would automatically delete the include statement as well.

test/boxtype.cxx
```diff
@@ -176,6 +177,7 @@ int main(int argc, char ** argv) {
   bt("FL_GTK_ROUND_DOWN_BOX",FL_GTK_ROUND_DOWN_BOX);
   bg->end();
   window->resizable(window);
+  Fl_Scheme_Choice scheme_choice(610, 10, 150, 30, "Scheme:");
```
MatthiasWM, line 180: I don't feel that this is needed here, because the window shows most box types anyway, regardless of the selected scheme. Then again, why not.

test/unittest_schemes.cxx
```diff
@@ -82,22 +73,9 @@ class Ut_Schemes_Test : public Fl_Group {
   }
   Ut_Schemes_Test(int X,int Y,int W,int H)
     : Fl_Group(X,Y,W,H) {
-    scheme_choice_ = new Fl_Choice(X+125,Y,140,25,"FLTK Scheme");
```
MatthiasWM, line 85: Nice reduction of code here, and easier to maintain. Thanks!

MatthiasWM:
Seems fine to me. Could neither find bugs nor typos.

fluid/alignment_panel.fl
Albrecht-S: Thanks for the hint. I always have problems to insert such things like #include in the right place(s). Fluid is still a mystery to me (and I don't meant its code but its user interface). Admittedly I've never actively worked with fluid.

test/boxtype.cxx
Albrecht-S: It was really needed. It's important to be able to switch the scheme and see which of the standard boxtypes (those of the "normal" scheme are overwritten - and how.

If you actually try it you may notice that there are no "round(ed)" boxes in the `gleam` scheme. This has the effect that radio buttons are not round (circles) as they should be. Try the "buttons" demo as well.

test/unittest_schemes.cxx
Albrecht-S: My goal is still to be able to add new schemes (or let users add new schemes) without patching the FLTK code in several (many!) places. I'm making good progress - step by step - and eventually this should work. And I also hate having to find all the places if we add new schemes or similar things. Maintainability is important!

Albrecht-S:
There are still some things to do in my list but I believe I can merge the first part soon.

fluid/alignment_panel.fl
MatthiasWM: FLUID is not really hard to use, but it does some hidden things that may be useful, but are unexpected. Working on it.

Albrecht-S:
Note: I rewrote the branch (i.e. the order of commits) to separate the new `Fl_Scheme` class and the `Fl_Scheme_Choice` widget. Both are added in different commits. The third commit is an update of an example program.

The plan is to extend the `Fl_Scheme` class in the future but this is not urgent. Some smaller additions will be added before the release of 1.4.0 though.

I'm ready to merge this PR tomorrow.

PR #625, opened 2023-01-03
Fix mandelbrot demo: box and missing last line

wcout:
`mandelbrot` has hardcoded box offsets dx/dy/dw/dh 3/4/6/8. These are wrong with all current styles and produce a black rectangle around the drawing. Also the last pixel line of the image is never drawn.

Here is a video showing the effect:

https://user-images.githubusercontent.com/9895976/210402876-ff157ab1-1871-4425-8e83-1fe51fdcaca1.mp4

### DIFF
--- a/test/mandelbrot.cxx
+++ b/test/mandelbrot.cxx
@@ -1,7 +1,7 @@
 //
 // Mandelbrot set demo for the Fast Light Tool Kit (FLTK).
 //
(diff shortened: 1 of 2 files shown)


wcout:
Sorry I missed a line - will create a corrected PR...

wcout:
See the new PR #626

PR #626, opened 2023-01-03
Fix mandelbrot demo: box and missing last line

wcout:
`mandelbrot` has hardcoded box offsets dx/dy/dw/dh 3/4/6/8. These are wrong with all current styles and produce a black rectangle around the drawing. Also the last pixel line of the image is never drawn.

### DIFF
--- a/test/mandelbrot.cxx
+++ b/test/mandelbrot.cxx
@@ -1,7 +1,7 @@
 //
 // Mandelbrot set demo for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2010 by Bill Spitzak and others.
+// Copyright 1998-2023 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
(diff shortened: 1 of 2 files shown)


test/mandelbrot.cxx
```diff
@@ -78,6 +78,12 @@ void Drawing_Window::update_label() {
 }
 
 void Drawing_Area::draw() {
+  if (!dx) {
+    dx = Fl::box_dx(box());
+    dy = Fl::box_dy(box());
+    W -= 2*dx;
+    H -= 2*dy;
+  }
```
Albrecht-S, line 86: Is there a specific reason why you use `2*dx` and `2*dy` rather than `Fl::box_dw(box())` and `Fl::box_dh(box())`, respectively? Even if this would in almost all cases be the same value, it would IMHO be more consistent to use the "official" box values.

This could be used later as well (see note below).

test/mandelbrot.cxx
```diff
@@ -204,8 +210,8 @@ void Drawing_Area::new_display() {
 
 void Drawing_Area::resize(int XX,int YY,int WW,int HH) {
   if (WW != w() || HH != h()) {
-    W = WW-6;
-    H = HH-8;
+    W = WW - 2*dx;
+    H = HH - 2*dy;
```
Albrecht-S, line 214: See note above.

MatthiasWM:
Implemented in #631

PR #628, opened 2023-01-03
Add close buttons for individual tabs in Fl_Tabs

MatthiasWM:
Setting `FL_WHEN_CLOSED` for any child of `Fl_Tabs` will render an **X**
 symbol to the left of the label text in the corresponding tab. When clicked,
 the child's callback will be called with `FL_REASON_CLOSED`. It's up to the
 child to remove itself from its parent.

 Compressed tabs show no close button until they are fully expanded.

FLUID support implemented.

The callback reasons are introduced to support multiple causes for our single widget callback. The implementation of `Fl::callback_reason()` is compatible to the same interface in `Fl_Tree` and expands the functionality to all widgets and their callbacks. Unless there is a serious bug or oversight in my implementation, I would like to commit this to master soon and then add `FL_REASON_...` flags as they are needed. This has no influence on the API or ABI.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -51,6 +51,7 @@ Changes in FLTK 1.4.0                            Released: ??? ?? 2022
   - New Fl_ICO_Image class to read Windows .ico icon files.
   - New classes Fl_SVG_File_Surface and Fl_EPS_File_Surface to save any FLTK
     graphics to SVG or EPS files, respectively.
+  - Fl_Tabs now supports close buttons for individual tabs.
   - Windows platform: added support for using a manifest to set the
     application's level of DPI awareness (issue #309).
   - class Fl_Native_File_Chooser on the X11/Wayland platform relies on external
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -1,7 +1,7 @@
 //
 // Enumerations for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2022 by Bill Spitzak and others.
+// Copyright 1998-2023 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -415,8 +415,13 @@ enum Fl_Event { // events
 /*@{*/
 /** These constants determine when a callback is performed.
 
-    \see Fl_Widget::when();
-    \todo doxygen comments for values are incomplete and maybe wrong or unclear
+    Fl_When is a bit field. Some values are merely shorcuts for common bit
+    combinations. New flags may be added in the future, so it's important to
+    mask the required bit when reading via \p when().
+
+    \note Some widgets may not fully suppoert \p FL_WHEN_... flags.
+
+    \see Fl_Widget::when(); Fl::callback_reason();
  */
 enum Fl_When { // Fl_Widget::when():
   FL_WHEN_NEVER             =  0, ///< Never call the callback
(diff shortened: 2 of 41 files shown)


MatthiasWM:
So there is one issue. Historically, `when()` was documented as returning one of the `FL_WHEN_...` values, although it actually always was a bitfield. Now even we have code like `if (when() == FL_WHEN_RELEASE)` in our code base, which will fail if `FL_WHEN_CLOSED` is set. This is obviously nothing that would normally happen, but it could. Especially if we should add more when() bits one day.

So here are proposals: 
 - keep returning the low 4 bits in `when()` and add `when_bitmap()` to return all bits
 - ... and add `when_closed()` to return only the FL_WHEN_CLOSED bit
 - keep returning all bit in `when()` and hope that it has no impact

Albrecht-S:
@MatthiasWM I didn't test this yet but I believe it's a great achievement. ISTR that we had a request or even an STR that had a patch - or maybe not. It might be useful to check what we have already to combine the good ideas that might exist with your code, and eventually to close related STR's. Thanks!

Albrecht-S:
Handling of `when()` flags: I think we should implement two of your options:
1. keep returning all bits in `when(),` warn in the docs [and hope that it has no impact]
2. move the `Fl_Tabs` close button [flag/option] into a flag that is unrelated to `when()`

_**Reasoning:**_

Option 1 should be done anyway. It's important to document the fact if it isn't already. Then we should also check the FLTK core and make sure that it **always** handles the `when()` flags correctly as a bit mask. This could be done in another issue which should be part of the release milestone.

Option 2 is more a gut feeling. The `when()` flags are used to trigger callbacks in all kinds of widgets. Such a `FL_WHEN_CLOSED` bit might be ambiguous and confusing and such a special value that it should IMHO not be part of the overall widget `when()` flags.

Here's another suggestion (unless you did that already): wouldn't it be useful to add another `Fl_Tabs` specific option/flag to trigger a popup like "Do you really want to close tab <label>?" (note: with tab label) to avoid that a user closes a tab by accident. If this flag is not set it's up to the developer to handle this in their own callback (or not).

Albrecht-S:
Oh, and sorry for asking for another unrelated feature: while you're at it, could you please check related issue #630 ([STR 3314](https://www.fltk.org/str.php?L3314)) that requests making some methods virtual? This could of course be handled separately, but since there are edits in `Fl_Tabs` anyway... TIA.

MatthiasWM:
Yes, that's easy enough to do. I was thinking about porting other methods of handling an overflowing Tab Group. I have code somewhere that adds a pulldown menu for that tabs that can't be shown. Very helpful.

MatthiasWM:
implement handle_overflow() with a few flags, probably:
 - Fl_Tabs::COMPRESS for the current behaviour
 - Fl_Tabs::CLIP as the name said
 - Fl_Tabs::DRAG would make the entire tab draggable left and right
 - Fl_Tabs::PULLDOWN show a pulldown menu with the items that don't fit in the tabs

Albrecht-S:
@MatthiasWM Ping: did you notice my [comment about the Fl::when() flags](https://github.com/fltk/fltk/pull/628#issuecomment-1370921490)? 

What are your thoughts?

MatthiasWM:
> @MatthiasWM Ping: did you notice my [comment about the Fl::when() flags](https://github.com/fltk/fltk/pull/628#issuecomment-1370921490)?
> 
> What are your thoughts?

I don't think that the Fl_Tabs widget needs to do anything when the user clicks the close button, just like the desktop doesn't do anything. All requests ("Are you sure?") are done by the app, including the ability to ignore the click. The current Fl_Tabs does exactly that: it calls the callback and lets the child do the rest (including possibly removing itself).

However, it's not the Fl_Tabs callback, so we must identify it for the child. This is done with Fl::callback_reason(). Fl_Tree and Fl_Table already use a local version of that, so implementing a global version makes sense. In the current commit, you can see that now every do_callback can have a Reason (and most do).

Now, without explicitly telling Fl_Tabs that if may send a Close callback, it would be dangerous to call a child widget callback. But without explicitly allowing a Close callback, it makes no sense to show the close button. So the FL_WHEN_CLOSED is very much needed.

Putting all that together, adding yet another flag to request the Close button seems redundant.

MatthiasWM:
OK, so I have all features in that I wanted. Maybe I can mix the OVERFLOW_DRAG and OVERFLOW_PULLDOWN at a later point. Also, the OVERFLOW_DRAG could show left and right arrows to make it clear that tabs can be dragged, and they should be draggable using the horizontal scroll event.

I added features to test/tabs that show all the new features, including adding and removing tabs and choosing an overflow handler.

Other than that, I valground the heck out of this and I am pretty confident that everything works.

Albrecht-S:
@MatthiasWM Wow, that looks great, thanks for these new features!

I found one minor (?) issue when testing though.
- if pulldown overflow mode is active
- and a particular tab is not visible (i.e. in the overflow area)
- and that tab has a close button

then you can activate that tab via the pulldown menu (good) but you can't close it via its close button because it is not visible.

Did I miss anything?
Should the pulldown mode also allow to scroll (drag) the tabs?
Or should selecting a hidden tab also make it visible by scrolling the tabs, even if drag mode via mouse is not enabled?

MatthiasWM:
I made the pulldown mode tabs draggable as well. In both draggable modes, selecting a widget will ensure that the widget is visible. The pulldown menu need all tab now, because there could be invible items on the left or right of the visible tabs, and leaving out the visible ones seems somewhat inconsistant.

Albrecht-S:
Great!

PR #631, opened 2023-01-04
Mandelbrot demo: Add color mode

wcout:
Toggle between (current) greyscale  and color mode with `Ctrl+M`. Default mode is still greyscale.

This PR also includes the fixes from #626.

mandelbrot-color

### DIFF
--- a/test/mandelbrot.cxx
+++ b/test/mandelbrot.cxx
@@ -1,7 +1,7 @@
 //
 // Mandelbrot set demo for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2010 by Bill Spitzak and others.
+// Copyright 1998-2023 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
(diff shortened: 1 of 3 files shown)


MatthiasWM:
I like the fact that this tests fl_draw_image *and* fl_draw_image_mono. I don't care about the colors themselves as this is just a test app and should not get too complicated.

Albrecht-S:
> I like the fact that this tests fl_draw_image and fl_draw_image_mono.

I wouldn't mind if someone checked and merged the PR if it's good to have. As I wrote before I don't have time to do it.

test/mandelbrot.cxx
```diff
@@ -21,6 +21,47 @@
...
+Fl_Color lighter(Fl_Color c) { return fl_color_average(c, FL_WHITE, 0.98); }
+Fl_Color  darker(Fl_Color c) { return fl_color_average(c, FL_BLACK, 0.98); }
+
+void get_color(int index, uchar& r, uchar& g, uchar& b, int brightness) {
+  Fl_Color c = palette[index % palette_size] << 8;
+  for (int i=0; i>brightness; i--) c = darker(c);
+  for (int i=0; i<brightness; i++) c = lighter(c);
+  Fl::get_color(c, r, g, b);
+}
+
```
MatthiasWM, line 64: You can greatly reduce the complexity here. Just replace line 53 to 63 with
```
const size_t palette_size = sizeof(palette)/sizeof(palette[0]);

void get_color(float t, uchar& r, uchar& g, uchar& b) {
  int index = (1-t) * palette_size;
  Fl_Color c = palette[index % palette_size] << 8;
  Fl::get_color(c, r, g, b);
}
```
to avoid the two loops, then...

test/mandelbrot.cxx
```diff
@@ -87,37 +143,50 @@ int Drawing_Area::idle() {
...
         wy = 2*wx*wy + yy;
         wx = t;
         if (wx*wx + wy*wy > 4) {
           wx = t = 1-double(i)/(1<<10);
-          if (t <= 0) t = 0; else for (i=brightness; i--;) t*=wx;
-          *p = 255-int(254*t);
+          if (t <= 0) t = 0; else for (i=(use_colors?MAX_BRIGHTNESS-2:brightness); i--;) t*=wx;
+          int c = 255 - int(254*t);
+          if (use_colors) get_color(c, *p, *(p+1), *(p+2), brightness-MAX_BRIGHTNESS/2);
+          else *p = c;
```
MatthiasWM, line 182: ... replace 179 to 182 with
```
          if (t <= 0) t = 0; else for (i=brightness; i--;) t*=wx;
          if (use_colors) {
            get_color(t, *p, *(p+1), *(p+2));
          } else {
            *p = 255 - int(254*t);
          }
```
which seems a bit more cleaned up to me and keeps the steepness of the brightness ramp as it works in grayscale

wcout: Thanks for reviewing.
You are probably right, cleanliness is more important than fanciness (I was not satisfied with the look at minimal brightness, so I tried to improve that). But so it is more in sync with the greyscale version.
This would mean, that `DEFAULT_BRIGHTNESS_COLOR` thing could also be removed...

If you are going to commit this PR, would you mind to do the required changes yourself? You can also push to this PR or use your own PR and close this one.

MatthiasWM:
Implemented in #631

PR #636, opened 2023-01-06
X11: Optionally copy selection buffer to clipboard (STR 3229)

Albrecht-S:
The new method Fl::selection_to_clipboard(int) enables copying selection data to the clipboard on X11 if it is set to 1.

This feature was requested by [STR 3229](https://www.fltk.org/str.php?L3229) and the implementation was inspired by an `xterm` feature named "Select to Clipboard" which can be enabled by `ctrl` + `middle mouse button` + `Select to Clipboard` in an `xterm` window.

### DIFF
--- a/FL/Fl.H
+++ b/FL/Fl.H
@@ -1,7 +1,7 @@
 //
 // Main header file for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2022 by Bill Spitzak and others.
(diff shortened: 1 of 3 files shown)


Albrecht-S:
Note: this feature does **not** enable **selective (per widget)** copying of data to the selection buffer **or** the clipboard as discussed on [STR 3229](https://www.fltk.org/str.php?L3229) but I believe this is a good compromise.

Comments and suggestions appreciated.

Issue #69, opened 2020-04-01
Fl_Input_::position shadowing Fl_Widget::position


pwiecz:
In FLTK version 1.3 Fl_Input_::position method is shadowing Fl_Widget::position virtual method.
Maybe it's time to fix it in 1.4?

Albrecht-S:
Thanks for the heads-up.

This is true, however it would need an API change which we try to avoid as much as possible.

One possible solution: deprecate `Fl_Input_::position()` and add a new method, for instance `Fl_Input::insert_position()` to replace it (all three overloaded methods). Then we could remove `Fl_Input_::position()` in a later FLTK version, presumably 1.5 (or higher).

pwiecz:
Oh, I thought that 1.4 is not considered stable, so API changes are still possible.
But I understand, that API is already stabilised.

Albrecht-S:
Thanks for your feedback. To explain: we try to keep the API (Application Programming Interface) across all versions so developers don't need to change their program code from one FLTK version to another. You could still compile most of FLTK 1.1 code with 1.3 (with a few minor exceptions) and so you will with 1.4.

The ABI (Application Binary Interface) however is not yet final for 1.4, it can still be modified until 1.4.0 gets released, but must be constant during 1.4.x releases.

Albrecht-S:
FTR: I noticed that `Fl_Text_Display` and `Fl_Text_Editor` (by inheritance) already have an `insert_position()` method. This needs to be checked and can be either

1. good, if arguments and semantics are the same (compatible)
2. bad, if not. Then we'd need another name.

This is just a reminder - code not yet checked.

Albrecht-S:
> Fl_Tile::position -> Fl_Tile::adjust

Just for the record: isn't this `Fl_Tile::move_intersection()` in #666 ?

MatthiasWM:
Yep, I changed my mind when implementing it. Fixed.

PR #666, opened 2023-01-31
`FL_DEPRECATED` macro to mark `position()` method that shadow `Fl_Widget::position()` #69

MatthiasWM:


FL/fl_attr.h
```diff
@@ -22,83 +22,160 @@
...
+
+
+/*
+ Declare macros specific to clang
+
+ Macros may have been declared already in previous sections.
+ */
+#if defined(__clang__)
+
+#define CLANG_VERSION (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
```
Albrecht-S, line 128: It would IMHO be better to use FL_CLANG_VERSION and FL_GCC_VERSION (below) to avoid name clashes with user-defined names. I thought of #undef'ining these names at the end, but this wouldn't avoid to interfere with user's #define's (maybe it would be even worse).

FL/fl_attr.h
```diff
@@ -22,83 +22,160 @@
...
+
+
+/*
+ Declare macros specific to gcc.
+
+ Macros may have been declared already in previous sections.
+ */
+#if defined(__GNUC__)
+
+#define GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
```
Albrecht-S, line 142: See comment above: suggestion to use FL_GCC_VERSION.

Albrecht-S:
That's a great work, I like it. The restructured `fl_attr.h` looks good (but see my comments), and the new `FL_DEPRECATED` macro is a great achievement.

I wonder how you found out which MSVC version supports which "dialect" of the `deprecated` attribute. Awesome.

Note that I just **read** the diffs, I couldn't test anything yet, hence no "official review" so far. But everything I saw looks good.

FL/fl_attr.h
MatthiasWM: Done

FL/fl_attr.h
MatthiasWM: Done

MatthiasWM:
> I wonder how you found out which MSVC version supports which "dialect" of the deprecated attribute. Awesome.

Thanks. Lots and lots of Google and StackOverflow.

MatthiasWM:
PS: in this particular case, the BOOST source code...

Issue #453, opened 2022-06-24
Unicode support in ExternalCodeEditor_WIN32.cxx


ZJUGKC:
Several Windows API functions with `W` version should be used in `fluid/ExternalCodeEditor_WIN32.cxx`.

```
GetFileAttributeA -> GetFileAttributeW
PostMessage -> PostMessageW
CreateFile -> CreateFileW
DeleteFile -> DeleteFileW
GetTempPath -> GetTempPathW
RemoveDirectory -> RemoveDirectoryW
CreateDirectory -> CreateDirectoryW
CreateProcess -> CreateProcessW
```

Albrecht-S:
Good point, thanks. This really needs to be done. We **must** always use the "wide character", i.e. Unicode versions of system functions which are in most cases wrapped in `fl_*` function names. I'm not sure if just renaming the function calls is enough though (_if this was the suggestion of the OP_) - this should be investigated and verified.

@erco77 Greg, ISTR that you are the original author of the `ExternalCodeEditor` functions. Would you like to take this issue?

MatthiasWM:
PR #668

Albrecht-S:
PR #668 looks fine, good to go from me.

MatthiasWM:
Fixed in 59d3b2e9fd10bdf14592e82ced422346ecd7204e

PR #668, opened 2023-01-31
FLUID: Unicode support in ExternalCodeEditor_WIN32.cxx (#453)

MatthiasWM:
I believe I caught them all.


MatthiasWM:
@ZJUGKC @erco77 @Albrecht-S if you guys find the time to verify this. Windows is not my first platform (or second).

fluid/ExternalCodeEditor_WIN32.cxx
```diff
@@ -312,9 +337,12 @@ int ExternalCodeEditor::remove_tmpfile() {
 //     Returns pointer to static memory.
 //
 const char* ExternalCodeEditor::tmpdir_name() {
-  char tempdir[100];
-  if (GetTempPath(sizeof(tempdir), tempdir) == 0 ) {
+  wchar_t tempdirW[FL_PATH_MAX+1];
+  char tempdir[FL_PATH_MAX+1];
+  if (GetTempPathW(FL_PATH_MAX, tempdirW) == 0) { // FIXME: word
```
Albrecht-S, line 342: What does this comment "FIXME: word" mean? It should have an obvious meaning to the reader, otherwise it's useless and confusing. Like it confuses me. ;-)

erco77: "Fix Me".. like TODO or XXX

MatthiasWM: A FIXME is similar to a TODO in that it marks a point in the code that still needs to be looked at. TODOs are stuff, that the programmer wants to come back to eventually, to improve thing or add features. FIXMEs however mark bugs that could not be fixed right now, but must be fixed as soon as possible. Some editors and code quality tools recognise TODO: and FIXME: and highlight them. Xcode inserts a band-aid icon in the function list pulldown.

In this case, I marked all the functions that needed to be converted to wide characters before actually changing them. This makes it very easy to find the lines in Xcode. I simply forgot to remove that last one. It's removed now.

Albrecht-S: @erco77 @MatthiasWM Thanks for all the kind explanations, but this was not my question. I know what "FIXME" means, it was about this particular text: "word". But Matthias explained this (and removed it now), thanks.

I didn't know that Xcode flags **comments**, that's a nice feature (but I'm not using Xcode).

Albrecht-S (approved):
Other than the two "nitpicking" comments I couldn't find any missed Windows calls that should be converted to wide char calls. I cross-compiled it with MinGW on Linux and didn't get any compiler warnings. However, I didn't **test** the resultant fluid.exe.

Anyway, good to go from me.

fluid/ExternalCodeEditor_WIN32.cxx
```diff
@@ -440,9 +471,10 @@ int ExternalCodeEditor::start_editor(const char *editor_cmd,
   // Command
   char cmd[1024];
   _snprintf(cmd, sizeof(cmd), "%s %s", editor_cmd, filename);
+  utf8_to_wchar(cmd, wbuf);
   // Start editor process
-  if (CreateProcess(NULL,               // app name
-                    (char*)cmd,         // command to exec
+  if (CreateProcessW(NULL,              // app name
+                    wbuf,               // command to exec
```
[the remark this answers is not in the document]
MatthiasWM: I don't think it should. The generic form does not have an A or W appended. It's somewhat of an FLTK thing that we do that. A user is normally not concerned with it.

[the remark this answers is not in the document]
Albrecht-S: OK, I'm fine with that too.

PR #683, opened 2023-02-21
Fl_String refactoring and extension

MatthiasWM:
I am in the process of refactor Fl_String to more closely resemble what FLTK needs of the std::string interface, plus adding a few methods for convenience.

To make sure that Fl_String will work perfectly, I decided to implement a subset of the Google Test API and write a few unit tests for the current Fl_String implementation as part of `test/unittest`. Run `unittest --core` to stay in the shell and not open a GUI.

Adding more tests is really easy and can be done in any module of `unittest` after including `unittests.h`. Simply do:
```C++
#include "unittests.h"
TEST(Math, Multiplication) {
  EXPECT_EQ(3*3, 9);
  char buf[32]; 
  sprintf(buf, "5*3=%d", 5 * 3);
  EXPECT_STREQ(buf, "5*3=15");
  return true;
}
TEST(Math, Subtraction) {
  EXPECT_EQ(2-3, -1);
  return true;
}
```
Multiple TEST() are supported across all code. Tests are grouped by the suite name (first argument). Eventually, just call `RUN_ALL_TESTS();`.


Albrecht-S:
Sorry, I don't have time to investigate the proposal right now, but I have questions.

Does this introduce a dependency on "Google Test" software if the user doesn't (want to) run tests?

Do we (devs) need to install `Google Test` if we want to run the tests, or how does it work?

Albrecht-S:
Here's one thing I suggest to change: when I implemented `Fl_String` I made it possible that `Fl_String` can store a `NULL` string, i.e. its `value()` (similar to `std::string::c_str()`) can be `NULL`.

I would now prefer to always store (or at least return) a **_valid string_** which would be the empty string `""` rather than `NULL`. I believe this is what `std::string` would do, but I'm not familiar with all details of `std::string`.

MatthiasWM:
@Albrecht-S There is a problem though. `fl_input_str` returns an Fl_String, and if I interpret the docs right, its value() is "" if the user entered nothing and pressed OK, and value() is NULL if the user pressed Cancel. Now the new Fl_String can't convey that information anymore.

Any ideas?

Albrecht-S:
BTW: the reason I chose `Fl_String::value()` was to be equivalent to `Fl_Input::value()`. Do you intend to remove the `value()` methods in favor of `c_str()` - or will we have both?

MatthiasWM:
Ok, I like 1 and 2, but 3 is not needed, because it is just either 1 or 2 (setting ret to NULL in 3 would be silly if you can simply choose 1). We could mimic the fl_choice_n return values if that makes any sense?

* -2 if the dialog was closed by pushing the window close button
* -1 if the dialog was closed by hitting Escape
*  0 if the button with \p b0 text is pushed (Here "OK")

Albrecht-S:
> I like 1 and 2, but 3 is not needed, because it is just either 1 or 2 ...

Yes, I agree, (3) would only be useful if we removed (1). But if we keep (1), then I also agree that (1) + (2) is the best solution.

My vote is +1 :+1: for keeping (1) and adding (2).

Return value: I agree with your proposal because the FLTK user can remember the values easily if they are the same (or at least similar) for all of these functions.

MatthiasWM:
Ok, I implemented that. Not super happy, but good enough. I still have to tighten the unit tests for the new Fl_String and add minimal string manipulation (append, insert, remove)

MatthiasWM:
Ok, so the implementation is complete. The API documentation is minimal, but probably sufficient for those who know std::strings. The unit tests were incredibly helpful. I will stick with that concept.

The dialogs return 0 for OK and -1 for Cancel. To get more info, I would need more restructuring which is probably not necessary.

It looks like a lot of code, but must methods are only one-liners that will be inlined by the compiler anyway. I checked that with Clang on macOS and was surprised just how much optimisation is going on in the assembly code. I will optimise FLUID to make better use of Fl_String in another branch, once this is merged.

MatthiasWM:
Sorry for all the additional commit after marking it ready for review. Anyway, I am happy with code and docs now and will hopefully be done with changes.

Albrecht-S:
No problem, but I found another one: `s/trailling/trailing/' multiple times, needs a global change.

Albrecht-S:
@MatthiasWM After some "quiet hours" I took the opportunity to push commit 201e6f87807cd76aa5a6b33aaafc4fd761dc3534 to your branch which fixes a few more typos and minor glitches.

Please make sure to pull my changes before you continue to work on it (if you do).

I read the code thoroughly and didn't see any obvious bugs.

Good to go from my side. Thanks for the great work to make this class (more) complete.

Albrecht-S:
I think the current version of `Fl_String` needs methods to append a single character, for instance:
```
Fl_String str;
str.append('a');
str += 'b';
```
Usage would be in `Fl_Help_View.cxx` to rewrite the internal class `HV_Edit_Buffer` as a subclass of `Fl_String` - I'm working on it and it's almost ready but appending a single `'char'` is missing in `Fl_String`. I could implement it (or keep it) in `HV_Edit_Buffer` but it's IMHO better if it's implemented in `Fl_String`. Could you, please ... ? [1]

Note: the class `HV_Edit_Buffer` must be a **subclass** of `Fl_String` because it needs a few special methods, but most of the code can be removed by using `Fl_String` (which I already "announced" in [comments in this file](https://github.com/fltk/fltk/blob/92818939267fc7fbb6a33d86fb78576f55ce6aca/src/Fl_Help_View.cxx#L280)).
Example code:
```
HV_Edit_Buffer buf;
buf += (*ptr++); // used like this in Fl_Help_View
buf += ' '; // etc. - old code was: buf.add(' ');
```

I'm preparing the update of `Fl_Help_View` so I can commit it as soon as the new `Fl_String` class is merged.

FYI, current status (WIP) is:
```
$ git diff --stat -- src/Fl_Help_View.cxx
 src/Fl_Help_View.cxx | 195 +++++++++++-----------------------------------------------
 1 file changed, 37 insertions(+), 158 deletions(-)
```
-------------------------
[1] FWIW, here's a quick and dirty hack as a proof of concept (patch):
```diff
diff --git a/FL/Fl_String.H b/FL/Fl_String.H
index fa59897b8..69bd6ee9d 100644
--- a/FL/Fl_String.H
+++ b/FL/Fl_String.H
@@ -107,6 +107,7 @@ public:
   void pop_back();
   Fl_String &append(const char *src, int n_ins=npos);
   Fl_String &append(const Fl_String &src);
+  Fl_String &operator+=(const char c);
   Fl_String &operator+=(const char *src);
   Fl_String &operator+=(const Fl_String &src);
   Fl_String &replace(int at, int n_del, const char *src, int n_ins=npos);
diff --git a/src/Fl_String.cxx b/src/Fl_String.cxx
index 5bfd04d91..d22980a44 100644
--- a/src/Fl_String.cxx
+++ b/src/Fl_String.cxx
@@ -423,6 +423,15 @@ Fl_String &Fl_String::append(const Fl_String &src) {
   return replace_(size_, 0, src.buffer_, src.size_);
 }
 
+/**
+ Append a character.
+ \param[in] src   copy character from here
+ \return self
+ */
+Fl_String &Fl_String::operator+=(const char src) {
+  return append(&src, 1);             // FIXME: should use method append(const char)
+}
+
 /**
  Append a C-style string or data.
  \param[in] src   copy C-style string from here
```

MatthiasWM:
There is Fl_String::push_back(char); . As ugly as it sounds, it's compatible to std::string. Thanks for all the fixes!

Albrecht-S:
> There is Fl_String::push_back(char);

Yes, when I woke up this morning I "suddenly knew" this. Sometime you need to sleep to find such things out. ;-)

`operator+=` would still be nice to have though - implemented with `push_back()`, obviously. Maybe:
```
Fl_String &Fl_String::operator+=(const char src) {
  push_back(src);
  return *this;
}
```

It's not strictly necessary, but it would be useful for better readable code. What do you think?

> Thanks for all the fixes!

Welcome.

wcout:
Sorry, I don't want to open an issue for this, because this is only a tiny miss-out: when building with Makefiles the `unittest_core` module is not linked in and therefore missing:

````diff
diff --git a/test/Makefile b/test/Makefile
index f1187caf5..79e61e02e 100644
--- a/test/Makefile
+++ b/test/Makefile
@@ -47,7 +47,8 @@ OBJUNITTEST = \
        unittest_viewport.o \
        unittest_scrollbarsize.o \
        unittest_schemes.o \
-       unittest_simple_terminal.o
+       unittest_simple_terminal.o \
+       unittest_core.o
````

**Comment by MatthiasWM** [comment]
Oopsie :-) thanks

**Comment by MatthiasWM** [comment]
Hmm, this was fixed here: https://github.com/fltk/fltk/commit/9f87af8ad9a3d8112518b6bbb5b6075881a5b9a2#diff-740cb5a1689091cb894445e46683c255e39689ccba1c6c63d8a8841c8df8817d 

Could you please verify that?

**Comment by wcout** [comment]
No, `unittest_core.o` is still missing in `test/Makefile` for target `OBJUNITTEST`.

**Comment by Albrecht-S** [comment]
OK, thanks, this is now fixed in e44a988c4a0770b7a12193cc023d1a55b5ff89e7.

I chose to use
```
OBJUNITTEST = \
	$(CPPUNITTEST:.cxx=.o)
```

rather than enumerating all files twice. This should hopefully fix this issue "forever".

MatthiasWM:
Oh! Thank you, Albrecht. So there is a negative aspect to making modules self-registering. You must make sure that the object is actually linked ;-)

RokerHRO:
> @Albrecht-S There is a problem though. `fl_input_str` returns an Fl_String, and if I interpret the docs right, its value() is "" if the user entered nothing and pressed OK, and value() is NULL if the user pressed Cancel. Now the new Fl_String can't convey that information anymore.
> 
> I can add Fl_String::null() that returns true if capacity() is 0, but that's incompatible to std::string and can be misleading. Add another flag?
> 
> Any ideas?

I'd always prefer `std::string` over an own string class because every C++ developer is familiar with it and everyone knows what this thing does and how it behaves.

If there is a function that might return a string (even an empty one) or not a string, I'd use `std::optional<std::string>`. That has a clear and well-known semantics fo all C++ developers, too.

I cannot see any benefit of a self-made Fl_String class, but a lot of disadvantages.

MatthiasWM:
> I cannot see any benefit of a self-made Fl_String class, but a lot of disadvantages.

@RokerHRO Fl_String is the result of a convention that we set a long time ago to stay compatible with even the earliest versions of C++ on the smallest CPU's with a screen: no templates, no std library, only C++98. We decided to keep it that way for fltk 1.4.

Fl_String has the same interface as std::string so that when we decide to go "modern", std::string should be a drop-in replacement for Fl_String. So for the time being, I can't offer std::optional (or Fl_Optional_String LOL). 

Eventually we will have to decide (1.5, 4.0?) if we want to revamp the API to be perfect C++, or sneak in more and more C++ features (like the one you suggested) over time. Comments welcome.

MatthiasWM:
PS: there is more. Fl_Int_Vector instead of std::vector<int> for example. Or FL_METHOD_CALLBACK instead of std::function, platform functions instead of std::chrono, all the fl_filename_* that now exists in std::filesystem. And FLUID's is_a() virtual methods to avoid RTTI.

The point is, where do we find the balance between "fast and light" and modern usability.

RokerHRO:
Okay, I see.

I am curious: What is the most-limited platform / CPU / OS / architecture & compiler that FLTK still shall support? Is that documented somewhere? If yes, we have a robust base for discussion. :-)

Why I ask: some C++ features (e.g. RTTI) needs run-time support and might(!) be expensive, but templates (e.g. std::string, std::vector etc.) brings no run-time costs at all, but a huge benefit of (very limited) "developer-time". And if FLTK wants more developer power it should not repel them by unnecessarily much home-brewn classes that have better and better-known equivalents in the C++ stdlib.

Moreover: the standard containers on most modern stdlib implementations are already more optimized than any self-made container can be. (Except you are a multi-billion dollar company like Facebook/Meta who is able to implement its own efficiency-focused string class. But even they got it not much better than std::string from modern stdlibs: https://www.youtube.com/watch?v=kPR8h4-qZdk )

I'd guess: >95% of all FLTK programs already use std::string and other stuff from the stdlib, so there is no good reason to avoid them within FLTK anymore.

And FLTK 1.3 and 1.4 supports Unicode, depends on font rendering libraries, X11/Wayland etc., which uses a lot of memory (and CPU cycles), so I guess it would not be able to run on a limited embedded system with 16 KB RAM anymore either.

So I'd suggest: Let's discuss that topic on a better suited place and fetch some input from the other FLTK developers, too, so we might get a solution which makes FLTK more attractive to other C++ developers as well. :-)

Albrecht-S:
> Let's discuss that topic on a better suited place and fetch some input from the other FLTK developers, too, so we might get a solution which makes FLTK more attractive to other C++ developers as well. :-)

@RokerHRO Thanks you for your thoughts. This discussion has already been started on GitHub Issue #789 and, to sum it up (as it stands now) we will
- make Fl_String and Fl_Int_Vector private classes (not in public header files) for the time being (FLTK 1.4)
- use these classes in FLTK 1.4 *only* internally
- release FLTK 1.4 with all the new goodies we developed (you mentioned some of these above)
- move to the next gen FLTK development with more modern C++ support soon after 1.4 has been released or maybe after some more minor releases (1.4.x)

Please watch the discussion in #789 for further progress.

PR #684, opened 2023-02-23
Derive internal Fl_Help_View buffer from Fl_String

Albrecht-S:
Refactor the internal class HV_Edit_Buffer by deriving it from the the recently extended Fl_String class.

The class HV_Edit_Buffer still needs some "special" methods but the code could be simplified substantially.

### DIFF
--- a/src/Fl_Help_View.cxx
+++ b/src/Fl_Help_View.cxx
@@ -1,11 +1,11 @@
 //
-// Fl_Help_View widget routines.
+// Fl_Help_View widget for the Fast Light Tool Kit (FLTK).
 //
 // Copyright 1997-2010 by Easy Software Products.
 // Image support by Matthias Melcher, Copyright 2000-2009.
 //
 // Buffer management (HV_Edit_Buffer) and more by AlbrechtS and others.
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Hi Matthias, could you please take a look at the simplified class `HV_Edit_Buffer`, particularly if I missed anything or if it could be optimized further? Just a quick glance ... TIA.

I believe it's OK (tests look good) and would like to merge it soon.

MatthiasWM (approved):
Nice! Feels like you removed almost as many lines as I added in Fl_String. Now the functionality is public and reused. I have not compiled the code, but what I see looks very straight forward. Thank you!

Albrecht-S:
@MatthiasWM Thanks for your review.

> Feels like you removed almost as many lines as I added in Fl_String.

Well, not really ;-), but particularly the buffer allocation stuff is no longer needed, and extending the string is now "for free" (except the Unicode code point argument). I could even remove constructor and destructor and use the default ones. Less code, less potential for bugs.

Thanks for extending `Fl_String` and particularly for adding `operator+=(char)`.

PR #688, opened 2023-02-26
Fix coordinates in gl_rect

ZJUGKC:
There is a problem in function `gl_rect`, line 224 in `gl_draw.cxx`:

```
glVertex2i(x+w, y+h-1);
```

The last point will exceed the rectangle.

So this should be `glVertex2i(x+w-1, y+h-1);`.


MatthiasWM:
Thanks! Merged with two minor additions.

PR #692, opened 2023-03-03
Eliminates a few compile warnings. Fixes a couple of potential bugs

ileGITimo:
Fix use after realloc of local_buf in fl_draw.cxx (bug?)

mLineStarts is a one element array but [0] and [1] are initialized in Fl_Text_Display.cxx  (bug?)

Replace deprecated XKeyCodeToKeysym with XGetKeyboardMapping and remove pragma protection around it in Fl_x.cxx.  Tested on Fedora 37, not sure it covers all cases, other or old environments.

Worked around get_list being defined but not used by moving/duplicating a few lines in gl_draw.cxx 

buffer was used before initialized, and choices initialization was incomplete in icon.cxx

### DIFF
--- a/FL/x.H
+++ b/FL/x.H
@@ -82,7 +82,7 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
+  if (!_sgc) { XFreeGC(fl_display, fl_gc); } fl_gc = _sgc
 
 extern FL_EXPORT void fl_copy_offscreen(int x, int y, int w, int h, Fl_Offscreen pixmap, int srcx, int srcy);
 #    define fl_delete_offscreen(pixmap) XFreePixmap(fl_display, pixmap)
--- a/src/Fl_Text_Display.cxx
+++ b/src/Fl_Text_Display.cxx
@@ -149,7 +149,7 @@ Fl_Text_Display::Fl_Text_Display(int X, int Y, int W, int H, const char* l)
   mStyleTable = 0;
   mNStyles = 0;
   mNVisibleLines = 1;
-  mLineStarts = new int[mNVisibleLines];
+  mLineStarts = new int[mNVisibleLines+1];
   mLineStarts[0] = 0;
   for (i=1; i<mNVisibleLines; i++)
     mLineStarts[i] = -1;
--- a/src/fl_draw.cxx
+++ b/src/fl_draw.cxx
@@ -80,7 +80,7 @@ static const char* expand_text_(const char* from, char*& buf, int maxbuf, double
     if (o > e) {
       if (maxbuf) break; // don't overflow buffer
       l_local_buff += (o - e) + 200; // enlarge buffer
-      buf = (char*)realloc(local_buf, l_local_buff);
+      buf = local_buf = (char*)realloc(local_buf, l_local_buff);
       e = buf + l_local_buff - 4; // update pointers to buffer content
       o = buf + (o - local_buf);
       word_end = buf + (word_end - local_buf);
--- a/test/icon.cxx
+++ b/test/icon.cxx
@@ -23,17 +23,17 @@ static Fl_Double_Window *win;
 
 void choice_cb(Fl_Widget *, void *v) {
   Fl_Color c = (Fl_Color)(fl_intptr_t)v;
-  uchar buffer[32*32*3];
+  uchar buffer[32*32*3] = {};
   Fl_RGB_Image icon(buffer, 32, 32, 3);
   icon.color_average(c, 0.0);
   win->icon(&icon);
 }
 
 Fl_Menu_Item choices[] = {
-  {"Red",0,choice_cb,(void*)FL_RED},
-  {"Green",0,choice_cb,(void*)FL_GREEN},
-  {"Blue",0,choice_cb,(void*)FL_BLUE},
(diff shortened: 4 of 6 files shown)


FL/x.H
```diff
@@ -82,7 +82,7 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
+  if (!_sgc) { XFreeGC(fl_display, fl_gc); } fl_gc = _sgc
```
MatthiasWM, line 85: I would prefer to spread the macro over more lines to make the intention clear.

src/Fl_Text_Display.cxx
```diff
@@ -149,7 +149,7 @@ Fl_Text_Display::Fl_Text_Display(int X, int Y, int W, int H, const char* l)
   mStyleTable = 0;
   mNStyles = 0;
   mNVisibleLines = 1;
-  mLineStarts = new int[mNVisibleLines];
+  mLineStarts = new int[mNVisibleLines+1];
```
MatthiasWM, line 152: The `for` loop is never executed because `i<mNVisibleLines` fails for `i=1` anyway. I am pretty sure this is a leftover when `mNVisibleLines` was initialised to more than 1. `new int[mNVisibleLines+1]` is not necessary.

FL/x.H
ileGITimo: I thought macros were passé with C++.  I tried to go with the spirit of whoever created that thing.  I would spread it out, a statement per line, or replace it with a non-macro implementation.

src/Fl_Text_Display.cxx
ileGITimo: Result of early morning oversight. Drop the loop?

src/fl_draw.cxx
```diff
@@ -80,7 +80,7 @@ static const char* expand_text_(const char* from, char*& buf, int maxbuf, double
     if (o > e) {
       if (maxbuf) break; // don't overflow buffer
       l_local_buff += (o - e) + 200; // enlarge buffer
-      buf = (char*)realloc(local_buf, l_local_buff);
+      buf = local_buf = (char*)realloc(local_buf, l_local_buff);
```
MatthiasWM, line 83: Doing this is wrong. `local_buf` is set correctly later, but in line 85, it is still used to recalculate an offset, setting the pointer `o` to point into the new buffer.

test/icon.cxx
```diff
@@ -23,17 +23,17 @@ static Fl_Double_Window *win;
 
 void choice_cb(Fl_Widget *, void *v) {
   Fl_Color c = (Fl_Color)(fl_intptr_t)v;
-  uchar buffer[32*32*3];
+  uchar buffer[32*32*3] = {};
```
MatthiasWM, line 26: This seems useful.

test/icon.cxx
```diff
@@ -23,17 +23,17 @@ static Fl_Double_Window *win;
...
 
 Fl_Menu_Item choices[] = {
-  {"Red",0,choice_cb,(void*)FL_RED},
-  {"Green",0,choice_cb,(void*)FL_GREEN},
-  {"Blue",0,choice_cb,(void*)FL_BLUE},
-  {0}
+  {"Red",0,choice_cb,(void*)FL_RED,0,0,0,0,0},
+  {"Green",0,choice_cb,(void*)FL_GREEN,0,0,0,0,0},
+  {"Blue",0,choice_cb,(void*)FL_BLUE,0,0,0,0,0},
+  {0,0,0,0,0,0,0,0,0}
```
MatthiasWM, line 36: This is unnecessary. Uninitialised members of a struct are always set to 0 if at least one element is initialised.

src/fl_draw.cxx
ileGITimo: I though realloc can move the buffer elsewhere, thus it's use in line 85 is not valid,

test/icon.cxx
ileGITimo: It is necessary to shut up the compiler.

FL/x.H
MatthiasWM: I know and agree. This is old code that abuses macros by hiding the creation of _sw and _ss in fl_begin_offscreen(). To stay backwards compatible, we need to keep the macro, but a better way would be some abstract type that is returned by a better `fl_begin_offscreen()`.

src/fl_draw.cxx
ileGITimo: probably need to copy it out to another value, to appease the compiler

MatthiasWM: Yes, it can, `o` is a pointer into the old buffer. Line 85 uses the old pointer to calculate the offset of `o` to the start of the old buffer. The pointer no longer points to allocated memory, but the pointer itself is valid for some pointer math. But yes, there are better ways to solve this.

test/icon.cxx
MatthiasWM: That is one picky compiler! ;-)

src/fl_draw.cxx
ileGITimo: Just trying to shut this up:
  fl_draw.cxx:86:34: warning: pointer used after ‘void* realloc(void*, size_t)’ [-Wuse-after-free]
will rework

test/icon.cxx
ileGITimo: Come to think of it.  It may just have been flycheck, not the compiler.

ileGITimo:
reworking will resubmit

PR #693, opened 2023-03-04
Branch 1.3

ileGITimo:
Eliminates 4 of the 5 warnings I get when compiling branch-1.3 head rev. with
  gcc version 12.2.1 20221121 (Red Hat 12.2.1-4) (GCC)
after cloning and running
  NOCONFIGURE=1 ./autogen.sh; ./configure --prefix=/usr; make

### DIFF
--- a/FL/x.H
+++ b/FL/x.H
@@ -82,7 +82,7 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
+  if (!_sgc) { XFreeGC(fl_display, fl_gc); } fl_gc = _sgc
 
 extern FL_EXPORT void fl_copy_offscreen(int x, int y, int w, int h, Fl_Offscreen pixmap, int srcx, int srcy);
 #    define fl_delete_offscreen(pixmap) XFreePixmap(fl_display, pixmap)
--- a/FL/x.H
+++ b/FL/x.H
@@ -82,7 +82,8 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
+  if (!_sgc) XFreeGC(fl_display, fl_gc); \
+  fl_gc = _sgc
 
 extern FL_EXPORT void fl_copy_offscreen(int x, int y, int w, int h, Fl_Offscreen pixmap, int srcx, int srcy);
 #    define fl_delete_offscreen(pixmap) XFreePixmap(fl_display, pixmap)
--- a/FL/x.H
+++ b/FL/x.H
@@ -82,7 +82,7 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) { XFreeGC(fl_display, fl_gc); } fl_gc = _sgc
+  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
 
 extern FL_EXPORT void fl_copy_offscreen(int x, int y, int w, int h, Fl_Offscreen pixmap, int srcx, int srcy);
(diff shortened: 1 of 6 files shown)


src/fl_draw.cxx
```diff
@@ -80,10 +80,12 @@ static const char* expand_text_(const char* from, char*& buf, int maxbuf, double
...
       if (maxbuf) break; // don't overflow buffer
       l_local_buff += (o - e) + 200; // enlarge buffer
+      auto d_o = (o - local_buf);
+      auto d_word_end = (word_end - local_buf);
       buf = (char*)realloc(local_buf, l_local_buff);
       e = buf + l_local_buff - 4; // update pointers to buffer content
-      o = buf + (o - local_buf);
-      word_end = buf + (word_end - local_buf);
+      o = buf + d_o;
+      word_end = buf + d_word_end;
```
MatthiasWM, line 88: Thanks for all your effort. I apologise for nagging again. For back compatibility, FLTK does not use "auto". IIRC `ptrdiff_t` from `stddef.h` is permitted.

Albrecht-S: This has been fixed in 1.4.0 (using size_t BTW). If at all we should use the code from 1.4.0 for a fix.

Note: I explained in fltk.general that we don't accept patches for pure warnings. In this case I would say this can be an issue and would backport the 1.4.0 code.

src/gl_draw.cxx
```diff
@@ -117,30 +117,33 @@ void  gl_font(int fontid, int size) {
...
 
 #ifndef __APPLE__
-static void get_list(int r) {
-  gl_fontsize->glok[r] = 1;
 #if defined(USE_X11)
 # if USE_XFT
 // FIXME
 # else
+static void get_list(int r) {
+  gl_fontsize->glok[r] = 1;
```
Albrecht-S, line 125: This code is completely different in 1.4.0, and I don't think it's worth fixing it in 1.3.x

FL/x.H
```diff
@@ -82,7 +82,8 @@ typedef ulong Fl_Offscreen;
   fl_push_no_clip()
 #    define fl_end_offscreen() \
   fl_pop_clip(); fl_window = _sw; _ss->set_current(); \
-  if (!_sgc) XFreeGC(fl_display, fl_gc); fl_gc = _sgc
+  if (!_sgc) XFreeGC(fl_display, fl_gc); \
+  fl_gc = _sgc
```
Albrecht-S, line 86: This looks like an acceptable - although unnecessary - change.

It's unnecessary because the warning is benign.

test/icon.cxx
```diff
@@ -23,7 +23,7 @@ static Fl_Double_Window *win;
 
 void choice_cb(Fl_Widget *, void *v) {
   Fl_Color c = (Fl_Color)(fl_intptr_t)v;
-  uchar buffer[32*32*3];
+  uchar buffer[32*32*3] = {};
```
Albrecht-S, line 26: Fixed in 1.4.0 by making this variable 'static' which fixes the warning w/o an initializer (because static variables are initialized with zeroes anyway).

ileGITimo:
OK, I'll wait to see what 1.3.9 has before submitting any thing else :-)

Albrecht-S:
@ileGITimo Please don't understand our comments wrong. We try to fix legitimate warnings, and we appreciate proposals. However, as I wrote in fltk.general in reply to a post from "Henrique" aka HASM (is that you?) we can't fix all compiler warnings in the 1.3.x series because 1.3.x is end of life. New compilers tend to generate new warnings, and we don't fix pure warnings in our old (1.3) code. You wrote:

> I'll wait to see what 1.3.9 has before submitting any thing else

Please don't spend your time on 1.3.8 and/or 1.3.9. These versions will not be maintained any more!

However, suggestions for fixes in 1.4.0 (Git master) will be appreciated and accepted. Thank you!

PS: maybe we **will** fix a few of your issues in 1.3.9, but there's no guarantee.

FL/x.H
Albrecht-S, line 86: Note: fixed in commit 66b0935ed912bc531a7527cd454c6ee90bda863c

Albrecht-S (commented):
FTR: I fixed/backported all warning fixes from 1.4.0 except the warnings related to get_list() (unused function) which is benign and has been fixed differently in 1.4.0.

FL/x.H
Albrecht-S: Fixed anyway as proposed.

test/icon.cxx
Albrecht-S: Backported from 1.4.0

Albrecht-S:
Thanks to @ileGITimo for all the suggestions. I decided to backport most of the fixes from 1.4.0 (Git master) except the issues related to `get_list()` which is a benign warning.

PR #699, opened 2023-03-15
Protect GL_SILENCE_DEPRECATION from re-definition.

ramcdona:
If users have already defined GL_SILENCE_DEPRECATION before #including FLTK, they currently get a warning.  This prevents that.


Albrecht-S:
Thanks, merged.

Issue #346, opened 2021-12-19
Fluid: redesign and fix all Settings dialogs


MatthiasWM:
We need to separate Fluid settings vs. Project settings and probably also support user settings, make sure all those are correctly initialised when starting a new design, and reorder them from multiple "grown" dialogs into two cleaned up Settings dialogs.

MatthiasWM:
See [STR 2586](https://www.fltk.org/str.php?L2586).

MatthiasWM:
Some major overhaul in #703.

PR #701, opened 2023-03-17
#346 partial: fixing the alignment panel 

MatthiasWM:
This patch fixes the interactive widget alignment and introduces a whole new way to manage widget layout rules.

- interactive sizing and removing now clips correctly to a number of rules
- additional rules can be added via classes derived from `Fd_Snap_Action`
- the alignment dialog is completely reworked
- alignment rules are much more useful now
- alignment rules can be stored with user preferences, within an .fl project file, or in an .fll layout file
- layout suites can have different rules for main app windows, dialogs, and toolboxes
- documentation must still be written


src/fl_draw.cxx
MatthiasWM, line 1: Memcheck would complain that we did pointer math with a NULL pointer (and then throw away the result)

src/fl_symbols.cxx
MatthiasWM, line 1: Added symbols for import and export (share) because they do not exist in Unicode. Did not need them in the end though.

src/Fl_Menu_.cxx
MatthiasWM, line 1: Fixing `Fl_Menu_Item::stonily()` because `Fl_Menu_::setonly(Fl_Menu_Item*)` does not work on submenu-pointer menu item arrays.

FL/Fl_Widget.H
MatthiasWM, line 1: `const` variant for `Fl_Widget::as_...` calls,

MatthiasWM:
Screenshot 2023-03-17 at 02 07 37

MatthiasWM:
All changes tested with Valgrind and Memcheck.

PR #715, opened 2023-04-10
FLTK Widget Add-On's made easy

MatthiasWM:
This commit makes it very easy to locally add widgets, tests, docs, and even FLUID plugins to core FLTK by simply copying one folder to the new directory `extensions/widgets/`. 

When writing a program, application coders may want to use external widgets to make FLTK more powerful. These small changes to the build system make it possible to isolate new widgets and all the supporting code into a single subdirectory. CMake can then take care of downloading and updating extension from GitHub etc. via a single `FetchContent_Populate` statement in `CMakeLists.txt`.

I will add example CMake projects next and then add an autoconf/Makefile setup.

This PR contains an example widget in `extensions/widgets/` which would not become part of FLTK, but instead any user could offer these widgets from their repos.


extensions/README.txt
MatthiasWM, line 1: Rough first text. This will be nicely documented.

Albrecht-S: Please look for capital letters at the beginning of lines in the middle of sentences. Is this some kind of auto correction by your editor? Looks weird.

Albrecht-S:
Nice idea, looks interesting. I'm not going to do a full review now but may do this later. Anyway, here are some first thoughts:

- `CMakeLists.txt`: `file(GLOB ...)` is not recommended in `CMake` because it can't detect new files automatically (when `CMake` is rerun by the build system). I don't know of a better solution for this extensions stuff, but wanted to mention it anyway (maybe it's OK in this context).
- `extensions/widgets/Fl_Range_Slider/*`: I don't know if you intend to keep this for a future merge, but if yes, I'd like to see a much simpler example than this one. A good example might be "Fl_Highlight_Button" that just overrides the handle method to highlight and un-highlight the button on `FL_ENTER` and `FL_LEAVE` resp. or something as simple as that. Perhaps it would be a better idea to put an example elsewhere so the extensions directory doesn't contain any widget in the default distribution (or maybe it's just a good idea to have an example in the default build?).

I didn't test this yet but I'd like to do if (when) you let us know that your work is complete enough to be tested. Is it already?

MatthiasWM:
Removed for the time being until we get a new roadmap for 1.5/4.0

PR #720, opened 2023-04-17
Add const qualifiers to can_undo and can_redo

ZJUGKC:
Add const qualifiers to can_undo() and can_redo() in Fl_Input_ and Fl_Text_Buffer.


Albrecht-S:
Thanks for the PR, patch looks good.

I suggest to merge. @MatthiasWM ?

PR #729, opened 2023-05-07
Improved, yet compatible, widget callback system using macros

MatthiasWM:
I added three macros that are C99 compatible, probably even C89 which allow callbacks of methods with custom parameters and even a capable emulsion of more C++ lambdas. Calls look like this:

```
 #include <FL/FL_CALLBACK.H>
 ...
 Fl_Button *btn1 = new Fl_Button(10, 10, 100, 20, "Beep");
 FL_FUNCTION_CALLBACK(btn1, fl_beep);
 ...
 Fl_Button *btn2 = new Fl_Button(10, 40, 100, 20, "Hello");
 FL_FUNCTION_CALLBACK(btn2,
   fl_message,
   const char *, text, "Hello\n%d %d %d %d",
   int, a, 1,  int, b, 2,  int, c, 3,  int, d, 4
 );
 ...
 Fl_Button *btn = new Fl_Button(10, 10, 100, 20, "Test");
 FL_METHOD_CALLBACK(btn, Fl_Button, btn, color, Fl_Color, c, FL_GREEN);
 ...
 Fl_Button *btn = new Fl_Button(10, 10, 100, 20, "Test");
 FL_INLINE_CALLBACK(btn,
   const char*, name, btn->label(),
   { fl_message("Greetings for the %s button", name); }
 );
```

Everything is just in a header file: `FL/FL_CALLBACK.H` (the caps are there to make clear it's all macros, but I can certainly change that). Here is an excerpt from the docs (search in the Doxygen output under Files). A working example is in `examples/callbacks.cxx`. If you guys like the macros, I would link the macros from the regular `callback` documentation.

---

Inline a function as a callback with custom parameters.

 This macro makes it possible to write a callback function right where the
 widget and callback is declared, somewhat similar to a Lambda function. The
 function body is limited to a syntacx that the macro preprocessor tollerates.

 The first argument is the widget that will handle this callback. The second
 argument is the function body itself. Up to five triplets can now follow,
 where each triplet is a comma separated list of a type, the parameter name,
 and the value, for example: `int, x, 3`. Passing two arguments is done in two
 triplets: `int, x, 3, int, y, 4`.

 Custom parameters are duplicated (shallow copy) whenever the macro is called
 and are marked to be automatically free'd by the callback widget
 when destroyed.

 There is a little demo program for FL_*_*CALLBACK in `examples/callbacks.cxx`.

 \param WIDGET the widget that will call the callback
 \param FUNC the function body within the limits of the C macro preprocessor
 \param ... a list of zero to five triples declaring type, name, and value

 \see FL_METHOD_CALLBACK, FL_INLINE_CALLBACK


Albrecht-S:
I'm impressed! I read the code and docs, all looks very good. However, the macros are - as you wrote in fltk.coredev - "unreadable". Do I need to fully understand them? Probably not at this time.

Some first thoughts, I'd like to test it to get a feeling for the new features. This is not yet a review:

1. (Why) do we need this `do { ... } while (0);` construct? Wouldn't simple curly braces `{ ... };` be sufficient? The `do` seems to imply some kind of loop - until you see `while (0);`. Confusing.
2. The `"user_data auto-free"` feature looks good by itself. Would it be possible to strip this off of this PR and commit it separately before the PR gets merged ? Note: if you commit the auto-free feature to master, a following rebase of your branch should "just work" (maybe with minimal fixes).
3. Nothing more for now, except the usual typos ;-) ... which I may report in another proof reading. Don't worry yourself.

MatthiasWM:
At the bottom of the header file are unrolled macros and the line that creates them. It's very similar to how lambdas are implemented in C++. No need to understand them. I am sure we will see issues over time that need to be fixed. Biggest issue for me is that they do compile for everyone.

1: the `do while` is a common trick to make the trailing semicolon in macros language compatible. That's all, nothing else. 
2: yeah, possible, sure
3: sigh, even though I proof read my stuff :-/

ManoloFLTK:
Macro FL_FUNCTION_CALLBACK seems to fill a need that occurs often where the user has to define a structure type to transmit several values to a callback function. I would believe that what happens exactly when a shallow copy of argument values is performed needs be documented.

I'm puzzled by the 2nd parameter of macro FL_INLINE_CALLBACK about which the user is told "The function body is limited to a syntax that the macro preprocessor can handle." To me, it would seem necessary to define clearly what code is admissible in such function body. Without this, the macro becomes unusable.

MatthiasWM:
@ManoloFLTK Shallow copy means that pointers are copied verbatim, but the data that they are pointing to is not copied. So if you use a string as a parameter, it must be retained by the caller. I figure out how to improve on that. Anyway, I will document what I mean by "shallow copy".

The second parameter of `FL_INLINE_CALLBACK` is simply a function body including the leading '{' and trailing '}'. I will rephrase this part of the docs. All the macro handling seems to work just fine.

A common case where macros fail is with templates. For example `test<a,b>` as an argument for macros is interpreted as two parameters, `test<a` and `b>` thanks to the comma. But this does not apply here, so I can remove the cautious documentation line.

ManoloFLTK:
Would this description of what is admissible as 2nd argument of macro `FL_INLINE_CALLBACK` be correct?

A function body including the leading '{' and trailing '}'. It may contain local variable declarations, use global variables and functions, and use also the variables listed and initialized in further argument triples of the macro. Very large function bodies should be avoided because they may exceed the admissible size of a macro argument. Template-using statements should be avoided/are admitted (_select what applies_).

MatthiasWM:
I modified the addition quite a bit to make it more usable and more readable. 

Instead of a chunk of memory, user parameters are now maged in a class derived from a new class `Fl_Callback_User_Data`. This makes it possible to have more complex types as parameters, for example `Fl_String`, and have their destructors called automatically as well. The Auto-Free flag is now and Auto-Delete flag and must be set to true of false if user data is derived from `Fl_Callback_User_Data`.

The last thing to add now before this can be merged is some more documentation.

MatthiasWM:
I tried to find all typos, and I also flipped the order of FL_INLINE_CALLBACK's arguments to make them feel more like a c++lambda function. So the inline callback is now:
```
FL_INLINE_CALLBACK(my_button.
                   int, x, 3,  int, y, 4,
                     {
                       fl_message("Click with %d and %d", x, y); 
                     }
                   );
```

MatthiasWM:
I overlooked that I did not test Microsofts VisualC compiler. It took MS until C++11 to implement variadic arg macros correctly. This is the same time when lambdas were introduced which would also fix our missing feature. So what's left now is to skip the whole variadic macros and use FL_FUNCTION_CALLBACK_1 for 1 argument, FL_FUNCTION_CALLBACK_2 for 2 arguments, and so on. 

This is *very* frustrating. I may have to abandon this whole idea and deal with the fact that 1.4 will always be like 1998.

Albrecht-S:
> It took MS until C++11 to implement variadic arg macros correctly. This is very frustrating. I may have to abandon this whole idea and deal with the fact that 1.4 will always be like 1998.

This stuff is optional anyway and nobody needs to use it. IMHO it would be fine if we could check the compiler (`MSVC`) and maybe C++ version (`C++11`) and disable the whole stuff on such platforms like `MSVC` before it is usable.

My take on this:
1. either check compiler/C++ version, disable it entirely on older ones that are not compliant (and document this incompatibility)
2. or dismiss this new feature entirely if (1.) is not possible

Maybe I can help with (1.) , at least for CMake (less for configure).

**_Reasoning:_** new (and optional) features may need new compilers and C++11 is officially more than 10 years old (I don't know when exactly it has been officially published). As long as everybody can build FLTK it would be OK for me.

MatthiasWM:
> 1. either check compiler/C++ version, disable it entirely on older ones that are not compliant (and document this incompatibility)
> 2. or dismiss this new feature entirely if (1.) is not possible
> 
> Maybe I can help with (1.) , at least for CMake (less for configure).

So it's only a header file anyway and completely optional. I removed all the VARARG stuff, making it less elegant, but still very usable. It's not even that much more code in the end, so I think it's not in out way.

We have already a very nice way to check compilers in fl_attr.h . If I still needed to check for working varargs, I would add a macro there. A feature that runs on all platforms but MSWindows would be a bit silly.
 
> **_Reasoning:_** new (and optional) features may need new compilers and C++11 is officially more than 10 years old (I don't know when exactly it has been officially published). As long as everybody can build FLTK it would be OK for me.

Well, if we require at least C++11, we could simply replace the function pointer in callbacks with std::function and use lambdas with captures, and none of this is needed at all. I find it amazing that FLTK is simple enough so that we have a plain C wrapper for it out there, and complex enough, so that Python can offer a much better FLTK callback wrapper than FLTK itself.

Albrecht-S:
> A feature that runs on all platforms but MSWindows would be a bit silly.

Just to clarify: That's not what I said at all.
- `MSWindows !== MSVC`: There are other toolchains that can be used.
- Besides that it would only be `MSVC` compiling in **C++ mode < `C++11`**.

I guess that would be a very small minority these days. But if you changed it anyway this is moot, never mind.

MatthiasWM:
> > A feature that runs on all platforms but MSWindows would be a bit silly.
> Just to clarify: That's not what I said at all.

Sorry, I did not mean to put words in your mouth.
 
> * `MSWindows !== MSVC`: There are other toolchains that can be used.
> * Besides that it would only be `MSVC` compiling in **C++ mode < `C++11`**.

They kept the buggy preprocessor as their default. Fixing it requires additional compile command line arguments ;-) . No intention to mess with that.
 
Anyway, now finally everything is tested and compiles on all platforms. I'll sleep a night over it, and if there is no thumbs down and I feel confident, I will merge this,

Issue #733, opened 2023-06-06
FLTK multi-touch gesture support


jinjin-Helen:
**Is your feature request related to a problem? Please describe.**

On current FLTK version (1.3.8 or 1.4.0), I didn't see it support multi-touch for touchscreen. So, I have a question will it plan to support it? @ManoloFLTK

Albrecht-S:
@jinjin-Helen The answer is "yes", but with some restrictions and only for selected multi-touch gestures. There is already work in progress (investigation of basic functionality) that should make it into FLTK 1.4.0 as a working solution. However, 1.3.x will not be updated.

(Multi-)Touch support will be implemented as far as possible across platforms:
- Windows, macOS: very likely "full" support (as far as possible and feasible)
- Linux/Unix + X11: there are currently no plans because we don't know how to do it
- Linux/Unix + Wayland: maybe (needs investigation)

It is possible that FLTK 1.4.0 will have only partial support that will be extended in a later version (1.4.x).

----
Note: I renamed the subject of this issue. Please ask questions concerning FLTK development in one of our Google Groups [fltk.general](https://groups.google.com/g/fltkgeneral) or [fltk.coredev](https://groups.google.com/g/fltkcoredev). Please see also "[Reporting Bugs](https://www.fltk.org/bugs.php)". Thanks.

jinjin-Helen:
@Albrecht-S, thank you for answering my question. So, I think I should find another way to implement it.

Albrecht-S:
@jinjin-Helen:
> I think I should find another way to implement it.

1. On which platform(s) do you need multitouch gesture support?
2. If you find a way to implement it on X11 we'd be interested in patches, TIA.

Note: I reopened the issue because this is something we need to do in the near future (please don't close it).

jinjin-Helen:
@Albrecht-S ,  We are pre-researching to implement on `Linux + Wayland`.

Albrecht-S:
Sorry for the late reply. I don't know how to implement touch events properly under Wayland. Looking at the [Wayland touch event documentation](https://wayland-book.com/seat/touch.html) I get the implression that Wayland's touch support is only the "simple form" that sends single touch events - as opposed to sending "swipe", "zoom", "rotate", and other complex events. You would need to compose such complex events yourself. Disclaier: I'm not a Wayland expert.

Regarding x11 touch events, the only point I can remember is that you probably need (or **should** use?) libinput, but I don't know how to use it.

Good luck!

[...]

PR #740, opened 2023-06-29
wayland: basic handling of touch inputs

via:
When trying to use fltk's wayland backend with phosh on a pinephone, the touchscreen inputs aren't registered.  It seems that no pointer events are sent for the touchscreen, and the only way to get them is to use the touch event api.  There is a lot more that could be done (supporting gestures, multitouch, etc), but this at least brings basic UI functionality for devices like that (and allow fltk to be used for phone apps!)

I'm not very familiar with FLTK code, or at all with wayland, so this probably needs a lot of work, but I can work through the demos/tests without crashing.

### DIFF
--- a/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
@@ -359,6 +359,51 @@ static struct wl_pointer_listener pointer_listener = {
   pointer_axis
 };
 
+
+static void touch_down(void *data,
+         struct wl_touch *wl_touch,
+         uint32_t serial,
+         uint32_t time,
+         wl_surface *surface,
+         int id,
+         wl_fixed_t x,
+         wl_fixed_t y)
+{
+  if (id != 0) return;
+  Fl_Window *win = event_coords_from_surface(surface, x, y);
+  if (!win) return;
+  struct Fl_Wayland_Screen_Driver::seat *seat = (struct Fl_Wayland_Screen_Driver::seat*)data;
+  seat->pointer_focus = surface;
+  pointer_button(data, NULL, serial, time, BTN_LEFT, WL_POINTER_BUTTON_STATE_PRESSED);
+}
+
+
+void touch_up(void* data, wl_touch* wl_touch, uint serial, uint time, int id) {
+  if (id != 0) return;
+  pointer_button(data, NULL, serial, time, BTN_LEFT, WL_POINTER_BUTTON_STATE_RELEASED);
+}
+
+
+void touch_motion(void* data, wl_touch* wl_touch, uint time, int id,
+                  wl_fixed_t x, wl_fixed_t y) {
+  if (id != 0) return;
+  pointer_motion(data, NULL, time, x, y);
+}
+
+
+void touch_frame(void* data, wl_touch* wl_touch) { }
+void touch_cancel(void* data, wl_touch* wl_touch) { }
+
+
+static struct wl_touch_listener touch_listener = {
+  touch_down,
+  touch_up,
+  touch_motion,
+  touch_frame,
+  touch_cancel,
+};
+
+
 static const char *proxy_tag = "FLTK for Wayland";
 
 bool Fl_Wayland_Screen_Driver::own_output(struct wl_output *output)
@@ -903,6 +948,14 @@ static void seat_capabilities(void *data, struct wl_seat *wl_seat, uint32_t capa
     wl_pointer_release(seat->wl_pointer);
     seat->wl_pointer = NULL;
   }
+  
+  if ((capabilities & WL_SEAT_CAPABILITY_TOUCH) && !seat->wl_touch) {
+      seat->wl_touch = wl_seat_get_touch(wl_seat);
+      wl_touch_add_listener(seat->wl_touch, &touch_listener, seat);
+    } else if (!(capabilities & WL_SEAT_CAPABILITY_TOUCH) && seat->wl_touch) {
+      wl_touch_release(seat->wl_touch);
+      seat->wl_touch = NULL;
+    }
 
   bool have_keyboard = seat->xkb_context && (capabilities & WL_SEAT_CAPABILITY_KEYBOARD);
   if (have_keyboard && seat->wl_keyboard == NULL) {
--- a/src/drivers/Wayland/Fl_Wayland_Screen_Driver.H
+++ b/src/drivers/Wayland/Fl_Wayland_Screen_Driver.H
@@ -43,6 +43,7 @@ public:
   struct seat {
     struct wl_seat *wl_seat;
     struct wl_pointer *wl_pointer;
+    struct wl_touch *wl_touch;
     struct wl_keyboard *wl_keyboard;
     uint32_t keyboard_enter_serial;
     struct wl_surface *keyboard_surface;


Albrecht-S:
Thanks for the PR, it looks simple and straightforward. Our Wayland specialist @ManoloFLTK should take a look at this and verify if it's correct and complete (enough) to accept this.

Just guessing: maybe we need to check the number of touch points to distinguish "primitive" touch actions from multi-touch gestures (which could be ignored for now), and maybe to do this we'd also have to interpret touch "frame" and "cancel" messages?

src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
```diff
@@ -359,6 +359,68 @@ static struct wl_pointer_listener pointer_listener = {
...
+         int id,
+         wl_fixed_t x,
+         wl_fixed_t y)
+{
+  struct Fl_Wayland_Screen_Driver::seat *seat = (struct Fl_Wayland_Screen_Driver::seat*)data;
+  seat->serial = serial;
+  wld_event_time = time;
+  Fl_Window *win = event_coords_from_surface(surface, x, y);
+  if (!win) return;
+  win = win->top_window();
```
ManoloFLTK, line 376: This is superfluous because it's already di$one in event_coords_from_surface() called before.

ManoloFLTK:
Apart from the one superfluous statement flagged in the code, I believe the proposed PR is correct. However, I can't test it because I don't have access to a system with a touch screen, therefore I never see a wl_touch event generated by a Wayland compositor.

About frame events: yes, the documentation of the wl_touch object suggests these events should be taken care of. However, without access to a system that produces them, I can't say if the current PR that ignores them has potential failures.

Frame events exists also for the wl_pointer object but only starting with its version 5, and the Mutter compositor has version 2, so no frame events are generated by the wl_pointer object under Gnome.

I would be tempted to shorten significantly the new code required by this PR reusing extant code for support of wl_pointer events. That is in the attached patch file. But, again, I can't test this proposed code.
[touch.patch.txt](https://github.com/fltk/fltk/files/11935858/touch.patch.txt)

Albrecht-S:
I can try both the PR and @ManoloFLTK 's patch on my notebook with a touch screen. I'll try to do this later today.

I'm interpreting the documentation such that you get one or more touch events for single fingers (down, up, move, ...) and the `frame` event closes the sequence of touch events and you can create the correct FLTK events. As long as the `frame` event didn't occur, a `cancel` touch event could still dismiss everything. That would happen if/when the WM (composer) decides to use the touch event on its own, e.g. for a swipe or window move event. Even without that (canceling a series of touch events) using the first finger touch would probably be ambiguous and when the user puts more fingers on the touch device for a swipe or zoom gesture, having "used" the first touch would likely be wrong. I'll check this theory and let you know what I'll find out...

Albrecht-S:
Here are some preliminary test results but this is only a first snapshot. I believe the PR is a good start, but in this basic version we have too many issues with multiple finger touch gestures. My theory about a `frame` touch event being the "last" event of a series of touch events doesn't hold, at least not reliably. Yes, `frame` events are sent after other touch events, but there's no clear distinction of different basic touch events. In the log below you can see an excerpt of a two-finger drag/swipe gesture with several `motion` and `frame` events.

I'm not sure how to resolve these issues yet, maybe there is a simple solution (e.g. counting the touch points and ignoring everything after the first touch point etc.) but this needs more investigation. So far I only logged touch events and FLTK events that would be sent but I'm investigating further.

As the OP @via noted, single clicks work, and (touch point) move events are also visible I'm not sure what to send for finger move events. Shouldn't moving a finger that touches the screen be a "drag" event? Anyway, basics are working, so far, so good.

However, as soon as I have more than one touch point (which will certainly happen in real life, even if we announce that we don't "support" multi-touch gestures in the beginning), then weird things can happen. A two-finger drag (swipe) event would send FL_MOVE events of both fingers intermixed. An excerpt from my logs:
```
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=440, y=184)
[ 429:touch_frame]
[ 430:touch_frame] -- Fl::handle(IGNORED)

[ 410:touch_motion] time = 4187075, id =     0
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=313, y=250)
[ 410:touch_motion] time = 4187075, id =     1
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=439, y=183)
[ 429:touch_frame]
[ 430:touch_frame] -- Fl::handle(IGNORED)

[ 410:touch_motion] time = 4187081, id =     0
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=314, y=251)
[ 429:touch_frame]
[ 430:touch_frame] -- Fl::handle(IGNORED)

[ 410:touch_motion] time = 4187081, id =     1
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=438, y=182)
[ 429:touch_frame]
[ 430:touch_frame] -- Fl::handle(IGNORED)

[ 410:touch_motion] time = 4187092, id =     0
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=314, y=252)
[ 410:touch_motion] time = 4187092, id =     1
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=438, y=182)
[ 429:touch_frame]
[ 430:touch_frame] -- Fl::handle(IGNORED)

[ 410:touch_motion] time = 4187098, id =     0
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=315, y=253)
```

Take a look at the `FL_MOVE` events that would currently be sent to the application. `id = 0` results from the first touch point (finger), `id = 1` from the 2nd finger. The x/y coordinates of these `FL_MOVE` events would "jump back and forth". Only showing the FL_MOVE events:
```
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=440, y=184)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=313, y=250)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=439, y=183)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=314, y=251)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=438, y=182)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=314, y=252)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=438, y=182)
[ 424:touch_motion] -- Fl::handle(FL_MOVE) (x=315, y=253)
```

Sending these events would certainly be confusing. Note that this log results from testing the original PR with added log statements.

Almost all of my tests have been done with `test/unittests`. I could repeatedly "hang" the application, i.e. it didn't react on any events at some times, and I had to kill it (I don't know how to reproduce this though, I assume it happened when I tried a three-finger gesture).

I very quick test with `test/editor` showed that I could select text with touch events but dragging the text inside the editor didn't work. I didn't investigate this further yet.

That's it for now to keep you informed. I'm still working on this...

Albrecht-S:
One more (benign) issue I noticed: touch events don't work on the window title bar. This is very likely caused by missing touch event handling in libdecor. This could probably be documented and ignored - but I'm not sure what happens on pure touch devices like smart phones in this case (as the OP used). @via ?

via:
Thank you @ManoloFLTK and @Albrecht-S for your time looking at this.
I tried @ManoloFLTK's patch and everything works for my use case.

@Albrecht-S I see what you're talking about and agree that that is particularly confusing behavior.  I added gates to all the touch functions to only take action when id is 0, and that seems to disregard all additional pointers.  In the interest of keeping the primitive "mouse replacement" functionality working (since there is no mouse, only touch events, on this platform), would that be a good enough solution for now?

I'm still really unclear on what to do with frame/cancel.  I never get cancel events, I do get frame events like you do after the touch events/motion.

via:
@Albrecht-S regarding the title bar, thats an independent source of confusion for me. No other phosh windows have window decorations, and I think thats intentional.  It feels like the detection for whether to do clientside or not is not working, but I don't know enough about phosh or wayland to know whats wrong.  Either way, I can't do anything with the title bar either, but phosh doesn't support moving windows anyway.

Albrecht-S:
> I'm still really unclear on what to do with frame/cancel. I never get cancel events, ...

I'm pretty sure I saw at least once a cancel event, but I don't know how to (re)produce it. It's very likely related to move/drag outside the window but this is maybe not possible on your phone. Can you confirm this? Are all apps on your phone full-screen?

Albrecht-S:
Hmm, a more serious issue with the "PR code" is that I can't interact with popup windows. They either don't get the focus or something else doesn't work. Try `test/ask` to see this. @via Can you interact with popup windows created by `test/ask`? Note that using Fl_Choice widgets (and maybe other menu windows) seems to work, but separate popup windows have issues.

via:
I pushed to my branch exactly @ManoloFLTK's patch with gating for id == 0 in each of the functions.

Apps are all full screen, and notably if the fltk window is not resizable, the UI is not really usable. Phosh doesn't do well with windows it can't resize to full screen.  There is a bar at the top, and I can make the on-screen keyboard appear.  Trying to do gestures either to or from these non-app areas of the screen still did not produce any cancel events for me.

@Albrecht-S re: your last question.  When I use test/ask, like many of the example apps, it uses windows that aren't resizable, and thus are almost unusable on the phone.  However, even though they don't resize correctly, I still appear able to run test/ask, press "Test text", type in the window, press ok, and also dismiss all the warning/timeout messages.  All with touch presses.

via:
Here are some screenshots for fun and context.  You can see that the test text field works (I'm pressing the button) but its obnoxious to use.  I was just not going to use any windows like this in my app :)

Other than that I am happy with how it looks.

20230703_11h45m40s_grim
20230703_11h45m45s_grim
20230703_11h45m58s_grim
20230703_11h46m02s_grim
20230703_11h46m10s_grim

Albrecht-S:
> I still appear able to run test/ask, press "Test text", type in the window, press ok, and also dismiss all the warning/timeout messages. All with touch presses.

OK, then this seems to be different. Presumably because `phosh` does "something different" with the Wayland windows. Note that `phosh` is currently not a supported target of FLTK but it's interesting to see that it works somehow. Everything we can do to make it work better is worthwhile.

via:
@Albrecht-S what compositor/environment are you using? Especially if its a blocker for getting phosh touch functionality to work, I'd be happy to try to debug it.

Albrecht-S:
> what compositor/environment are you using? ... I'd be happy to try to debug it.

@via I'm using Linux Mint 20 (based on Ubuntu 20.04 LTS) with a `Gnome/Wayland` user session on my older notebook with touch display (the only touch display available to me). All contributions to a better Wayland experience would be appreciated by the FLTK Team. TIA.

Albrecht-S:
@via I added similar guards (`id != 0`) to your original version with my additional debug statements. At a first glance it seems to work better but this still needs more investigation. In practice I don't think that `id == 0` is not guaranteed by the Wayland protocol to be the first touch id but for now it seems to work. A simple replacement would be to store the id, count the touch points, and ignore all but this stored id. Or something like that...

In the future I'd like to support multi-touch gestures as well, but this needs to be done on other platforms (Windows, macOS) first where I do already have working prototypes. But that would be the second step, I'd be glad if we could at least support single and double clicks and move (drag) events now.

Albrecht-S:
@ManoloFLTK Thanks for the pointer. I read this chapter before and now I read it again, but unfortunately it doesn't **define** the typical sequence of touch events. What it does in code is something like:
- accumulate all touch events until ...
- a `touch frame` event arrives, and then
- print results for each touch point and **invalidate all touch points**.

This does not strictly correlate with the sequence of events I'm seeing because even after a `frame` event there are still (IMHO) **valid** touch points (which previously received `touch down` events) and which will later receive `touch up` events, before the final `frame` event is received. It can still be useful to **process** touch events (only) when a `frame` event is received, as you can see in my example log above (multiple `motion` events), but there's no strict encapsulation of valid touch events and exactly one final `frame` event. At least that's what I see with my older version of Gnome/Wayland (i.e. `mutter`). I still don't know what exactly causes a `frame` event to be sent. Is it caused by a timer, i.e. lack of further movement, or what else?

@ManoloFLTK Can you tell me in what unit the `time` argument is sent? Is it milliseconds, or anything else? This might help to analyze the flow of events. Sorry if this is obvious, but I don't know.

I'll continue testing now. My earlier tests were all done with the **_original PR code_** (modified only by adding test statements), I'll try your (@ManoloFLTK's) patch now (i.e. the **_current PR code_**). Maybe its pointer handling is better and it will allow to handle popup windows in `test/ask.cxx` etc. ? That's just a guess but worth a try. Maybe...

Albrecht-S:
@via, @ManoloFLTK I used the updated branch at commit 1964416246b4048d2d881497b6c380e979fc1af2 ("Use @ManoloFLTK's patch and add gating for touch id") but unfortunately this doesn't have any effect on the inability to interact with popup windows. Even if I **_click with the mouse_** into the popup window to make it the active window I still can't use the touch screen to click any button in the popup. The button shows the "down" state after clicking (and it stays so) but it doesn't seem to trigger the callback.

Albrecht-S:
Hmm, strange. Whenever a new window appears, clicking twice on a button seems to make it work. This seems to be true for "normal" and for popup windows. Once this has been done, single clicks seem to trigger the button as expected. Still testing...

ManoloFLTK:
> @ManoloFLTK Can you tell me in what unit the time argument is sent? Is it milliseconds, or anything else? This might help to analyze the flow of events. Sorry if this is obvious, but I don't know.

The documenatation says:
	`The time argument is a timestamp with millisecond granularity, with an undefined base.`
I would interpret that sentence to mean the time argument is given in milliseconds.

ManoloFLTK:
> No other phosh windows have window decorations, and I think thats intentional. It feels like the detection for whether to do clientside or not is not working, but I don't know enough about phosh or wayland to know whats wrong. 

Wayland uses the [XDG decoration](https://wayland.app/protocols/xdg-decoration-unstable-v1) protocol to determine whether the compositor runs in CSD (client-side decoration) or SSD (server-side decoration) mode. If the compositor does not support this protocol, FLTK falls back on no decoration drawn on the client side. If the compositor doesn't decorate its windows, they result undecorated.

ManoloFLTK:
@Albrecht-S 

>     print results for each touch point and invalidate all touch points.
> 
> This does not strictly correlate with the sequence of events I'm seeing because even after a frame event there are still (IMHO) valid touch points (which previously received touch down events) and which will later receive touch up events, before the final frame event is received.

That remark about touch points remaining valid after a first frame event makes sense considering the last sentence of the documentation of `wl_touch` frame events:
> Indicates the end of a set of events that logically belong together. A client is expected to accumulate the data in all events within the frame before proceeding.
> 
> A wl_touch.frame terminates at least one event but otherwise no guarantee is provided about the set of events within a frame. A client must assume that any state not updated in a frame is unchanged from the previously known state.

Albrecht-S:
@ManoloFLTK Thank you very much for your input on
- time in milliseconds, and
- details about the frame event which I must have overlooked.

This makes sense, given my observations, and a future implementation of multi-touch events would need to take care of the touch frame semantics, for instance to notice in the first frame that it's a multi-touch gesture rather than a single touch click.

This doesn't solve the popup window problem though. I need to investigate what a mouse click does different than a touch_down event. Maybe you can give me a pointer where to look, or maybe I can find it myself. This would certainly help me understand the Wayland protocol and our implementation better...

ManoloFLTK:
@Albrecht-S : The "classical" pointer events have listeners for entry, up/down, movement, leave, whereas touch events have listeners for down, up, movement (ignoring frame which is not of interest here). These must assign the Wayland state variable seat->pointer_focus to the wl_surface currently having the focus, assign the FLTK state variables Fl::e_x_root, Fl::e_y_root, Fl::e_x, Fl::e_y, and send FLTK events FL_ENTER, FL_PUSH, FL_RELEASE, FL_MOVE, FL_LEAVE. Currently, the touch listeners send FL_PUSH, FL_RELEASE and FL_MOVE events and assign Wayland and FLTK state variables but don't send FL_ENTER or FL_LEAVE. This lack my explain the problem with popup windows.

Albrecht-S:
**_For the record:_**

See also [this comment](https://github.com/fltk/fltk/issues/733#issuecomment-1646591162) regarding touch events on title bars (moving windows) not working.

As it appears we'd need a current version of `libdecor` together with a current version of `mutter` to make this work, at least with the `GTK plugin` of `libdecor`. With "current version of mutter" I mean one that includes the updates that have been merged on Jul 19, 2023 (see the linked comment above).

Unfortunately my current test system has a much older version of mutter...

PR #767, opened 2023-08-14
Set win_gravity hint for X11

six809:
First patch updates gravity hint to something other than StaticGravity if the "-geometry" arg has been processed.

It also removes (#if's out) positioning bodges that were apparently to work around the behaviour of some window managers.  So you might want to back that bit out, but IMHO it's the WM's responsibility to make its decorations visible if that's what the user expects...

Second patch then applies an offset for east and south gravity accounting for the difference between actual and scaled screen dimensions.

End result: "-geometry +0+0" puts window in top-left, no 20-odd pixel displacement, "-geometry -0-0" puts it flush in bottom-right corner.  "resize" test app grows and shrinks from said corner.

Apologies if I missed something, I'm really not much of a C++ programmer.

### DIFF
--- a/src/drivers/X11/Fl_X11_Screen_Driver.cxx
+++ b/src/drivers/X11/Fl_X11_Screen_Driver.cxx
@@ -26,6 +26,7 @@
 #include <FL/Fl_Image_Surface.H>
 #include <FL/Fl_Tooltip.H>
 #include <FL/filename.H>
+#include <math.h>
 #include <sys/time.h>
 
 #include "../../Fl_Timeout.h"
@@ -250,6 +251,22 @@ int Fl_X11_Screen_Driver::h() {
   / screens[0].scale
 #endif
   ;
+}
+
+// Returns a real (unscaled) X coordinate calculated relative to the right of
+// of the screen.  Used when positioning window with "east" gravity.
+
+int Fl_X11_Screen_Driver::x_from_right(int x) {
+	int dx = w() - x;
+	return (fl_workarea_xywh[0] + fl_workarea_xywh[2]) - rint(dx * screens[0].scale);
+}
+
+// Returns a real (unscaled) Y coordinate calculated relative to the bottom of
+// the screen.  Used when positioning window with "south" gravity.
+
+int Fl_X11_Screen_Driver::y_from_bottom(int y) {
+	int dy = h() - y;
+	return (fl_workarea_xywh[1] + fl_workarea_xywh[3]) - rint(dy * screens[0].scale);
 }
 
 #define USE_XRANDR (HAVE_DLSYM && HAVE_DLFCN_H) // means attempt to dynamically load libXrandr.so
--- a/src/Fl_Screen_Driver.H
+++ b/src/Fl_Screen_Driver.H
@@ -111,6 +111,8 @@ public:
   virtual int y() { return 0; }
   virtual int w() { return 800; } // default, FL_OVERRIDE in driver!
   virtual int h() { return 600; } // default, FL_OVERRIDE in driver!
+  virtual int x_from_right(int x) { return x * scale(0); }
+  virtual int y_from_bottom(int y) { return y * scale(0); }
   virtual int screen_count();
   void screen_xywh(int &X, int &Y, int &W, int &H, int mx, int my);
   virtual void screen_xywh(int &X, int &Y, int &W, int &H, int /*n*/) {
--- a/src/Fl_Window_Driver.H
+++ b/src/Fl_Window_Driver.H
@@ -58,6 +58,8 @@ private:
 protected:
   Fl_Window *pWindow;
   int screen_num_; // number of screen where window is mapped
+  int win_gravity_;
+
 public:
   Fl_Window_Driver(Fl_Window *);
   virtual ~Fl_Window_Driver();
@@ -69,6 +71,24 @@ public:
   int screen_num();
   void screen_num(int n) { screen_num_ = n; }
 
+  // Window gravity
+  enum WIN_GRAVITY {
+    WIN_GRAVITY_UNMAP = 0,
+    WIN_GRAVITY_NORTHWEST = 1,
+    WIN_GRAVITY_NORTH = 2,
+    WIN_GRAVITY_NORTHEAST = 3,
+    WIN_GRAVITY_WEST = 4,
+    WIN_GRAVITY_CENTER = 5,
+    WIN_GRAVITY_EAST = 6,
+    WIN_GRAVITY_SOUTHWEST = 7,
+    WIN_GRAVITY_SOUTH = 8,
+    WIN_GRAVITY_SOUTHEAST = 9,
+    WIN_GRAVITY_STATIC = 10,
+  };
+  /**
+    Sets the desired window gravity.  Almost certainly only useful in X11.
+  */
+  virtual void win_gravity(int g) { win_gravity_ = g; }
 
   // --- frequently used accessors to public window data
   /** returns the x coordinate of the window. */
--- a/src/Fl_Window_Driver.cxx
+++ b/src/Fl_Window_Driver.cxx
@@ -43,6 +43,7 @@ Fl_Window_Driver::Fl_Window_Driver(Fl_Window *win)
   wait_for_expose_value = 0;
   other_xid = 0;
   screen_num_ = 0;
+  win_gravity_ = WIN_GRAVITY_STATIC;
 }
 
 
--- a/src/Fl_arg.cxx
+++ b/src/Fl_arg.cxx
@@ -282,8 +282,8 @@ void Fl_Window::show(int argc, char **argv) {
     if (geometry) {
       int fl = 0, gx = x(), gy = y(); unsigned int gw = w(), gh = h();
       fl = Fl::screen_driver()->XParseGeometry(geometry, &gx, &gy, &gw, &gh);
-      if (fl & Fl_Screen_Driver::fl_XNegative) gx = Fl::w()-w()+gx;
-      if (fl & Fl_Screen_Driver::fl_YNegative) gy = Fl::h()-h()+gy;
+      if (fl & Fl_Screen_Driver::fl_XNegative) gx = Fl::w()-gw+gx;
+      if (fl & Fl_Screen_Driver::fl_YNegative) gy = Fl::h()-gh+gy;
 
       // Determine window gravity from geometry.  Note that negative X and Y
       // will only end up truly accurate if the screen scale is (or has been
--- a/src/Fl_arg.cxx
+++ b/src/Fl_arg.cxx
@@ -284,6 +284,36 @@ void Fl_Window::show(int argc, char **argv) {
       fl = Fl::screen_driver()->XParseGeometry(geometry, &gx, &gy, &gw, &gh);
       if (fl & Fl_Screen_Driver::fl_XNegative) gx = Fl::w()-w()+gx;
       if (fl & Fl_Screen_Driver::fl_YNegative) gy = Fl::h()-h()+gy;
+
+      // Determine window gravity from geometry.  Note that negative X and Y
+      // will only end up truly accurate if the screen scale is (or has been
+      // set to) 1.0.  This is because information about the actual screen
+      // dimensions has been lost scaling integers.
+      static const Fl_Window_Driver::WIN_GRAVITY gravity_table[16] = {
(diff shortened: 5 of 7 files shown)


src/drivers/X11/Fl_X11_Screen_Driver.cxx
```diff
@@ -252,6 +252,20 @@ int Fl_X11_Screen_Driver::h() {
   ;
 }
 
+// Returns the difference between the screen's actual width in pixels and the
+// scaled up width.  Used when positioning window with "east" gravity.
+
+int Fl_X11_Screen_Driver::w_dx() {
+	return fl_workarea_xywh[2] - (int)(w() * screens[0].scale);
```
Albrecht-S, line 259: The constant index 0 (zero) in `screens[0].scale` doesn't look right. Shouldn't this be the actual screen's index?

I'd assume that we'd need the screen index as an argument in ` int Fl_X11_Screen_Driver::w_dx(int idx)` and the caller would have to set it appropriately.

Note: the same applies to `int Fl_X11_Screen_Driver::h_dy()` below.

Disclaimer: I noticed this by reviewing the code (w/o deeper knowledge of what exactly is going on here).

Albrecht-S:
@six809 Thank you very much for this PR. I just skimmed through the code changes and found that this looks sensible. However, I don't know enough about X protocols (particularly window "gravity") to decide that it is correct. We should check (i.e. test) this. Maybe someone else who knows X better than me can chime in and verify this PR.

It would be helpful if you could provide some test commands (e.g. `test/hello -g 100x100+50+50`) that don't work correctly so we can test and see what your PR would change. Can you please also describe what is wrong in these cases?

Questions: I understand that `test/hello -g 100x100+0+0` would not position the window as it should. Is this correct? What should be the correct window position if it has a title bar in this case? Thanks in advance.

Albrecht-S:
OK, I tested an example on my screen resolution 1920x1200:
- `test/hello -g 300x200+1620+1000` is correctly positioned in the bottom right corner
- `xwininfo` reports: `-geometry 300x200-1-1`

So far, so good.
- `test/hello -g 300x200-1-1` is **_NOT_** positioned in the bottom right corner
- `xwininfo` reports: `-geometry 300x200-41-13`

Is this a valid example and would your patch fix this?
Sorry, I could have pulled your patch and tested it but I'm just asking for clarification.

six809:
So first off, I've found my own issues with the second patch - I'd best rejig and resend.  Basically the error between the largest scaled extent and the actual extent may be different to the error for the actual requested X or Y position.  Discovered as I moved to a machine with a different vertical resolution, but still high DPI.

Going through other question:

You're almost certainly right about the screen index.  But Fl_X11_Screen_Driver::w() and Fl_X11_Screen_Driver::h() both refer explicitly to screen 0 too, so the calculation will be relative to those dimensions.  Looks like some existing code might also need attention.

Your example of specifying 100x100+0+0 does the wrong thing currently: the bodge code ensures it's offset 20-something pixels vertically, and possibly even 1 pixel horizontally.

A window manager should place anything specified with +0+0 such that the top left of its decorations - including any border it adds - are at the top left of the screen.  Flagging NorthWestGravity (which really tends to be identical to the default) means windows resize right and down.

The example using -1-1 _should_ result in the bottom right corner of the window (including any decorations added by the WM) being 1 pixel inset up and left from the bottom-right corner.  -0-0 would specify it being right in the corner.  In either case flagging SouthEastGravity should then make windows resize left and up.

Now, what a WM _actually_ does is entirely up to it, I suppose - these behaviours are what always seemed logical to me, and it's what I implement in evilwm.

But yes, I'll get back to you with a fix for the second patch - should I close this issue and open another?  I'm not a big github user either!  ;)

Albrecht-S:
@six809 Thanks you for your comment and for answering my questions.

> should I close this issue and open another?

Not necessarily.

1. You can just add another commit that fixes the issue, as long as it doesn't become too confusing.

2. It is OK if you squash/merge/rebase your commits in your own branch.

3. OTOH I would recommend and prefer that you create your own branch on top of current master for your patch(es). If you did this you could obviously create new commits and thus produce a clean history. Then you'd also have to create a new GitHub PR, and if you do that please close the old (this) PR and refer to it in your new PR so we can keep the context.

I just noticed that you added another commit (see 1.) and this is fine for now.

six809:
Lovely.  I won't touch my fork for now, and in future I'll do it properly.

Albrecht-S:
> I won't touch my fork for now, and in future I'll do it properly.

That's OK, thank you for your contribution.

Just a note: it can take a while before we accept and merge a PR. Therefore it's best to create an own branch for every single issue you want to fix and eventually a PR that requests to merge your branch.

If it would be necessary/useful in the future to update your branch with current master please rebase your branch on top of the (then) current master branch (e.g. `git rebase -i master`). To push rebased branch back to an existing branch in your fork you'd have to use `git push --force`.

Albrecht-S:
> You're almost certainly right about the screen index. But Fl_X11_Screen_Driver::w() and Fl_X11_Screen_Driver::h() both refer explicitly to screen 0 too, so the calculation will be relative to those dimensions. Looks like some existing code might also need attention.

OK, we may need to check what's going on there. It's not my code, hence I can't tell off the top of my head. Thanks for the hint.

six809:
Albrecht Schlosser wrote:
> Note: I assume that Fl_X11_Screen_Driver::w() and Fl_X11_Screen_Driver::h() are called by Fl::w() and Fl::h() respectively which return _"the height/width in pixels of the main screen work area"_.

That sounds right.  But I wasn't joking when I said I was bad at C++.

Apart from the comment in FL/Fl.H, "via screen driver" (then it's a
virtual in Fl_Screen_Driver, implemented in Fl_X11_Screen_Driver),
I don't know how one could figure out that it ends up there.

..ciaran

Albrecht-S:
@six809 Please see my note on your latest commit regarding initialization of member variables. My proposed fix (builds fine with `-std=c++98`):
```
diff --git a/src/Fl_Window_Driver.H b/src/Fl_Window_Driver.H
index d3d548714..29c09bea8 100644
--- a/src/Fl_Window_Driver.H
+++ b/src/Fl_Window_Driver.H
@@ -58,7 +58,7 @@ private:
 protected:
   Fl_Window *pWindow;
   int screen_num_; // number of screen where window is mapped
-  int win_gravity_ = WIN_GRAVITY_STATIC;
+  int win_gravity_;
 
 public:
   Fl_Window_Driver(Fl_Window *);
diff --git a/src/Fl_Window_Driver.cxx b/src/Fl_Window_Driver.cxx
index 25d2554c6..92d1d7343 100644
--- a/src/Fl_Window_Driver.cxx
+++ b/src/Fl_Window_Driver.cxx
@@ -43,6 +43,7 @@ Fl_Window_Driver::Fl_Window_Driver(Fl_Window *win)
   wait_for_expose_value = 0;
   other_xid = 0;
   screen_num_ = 0;
+  win_gravity_ = WIN_GRAVITY_STATIC;
 }
 
 
```

With this fix I could compile and build your branch successfully. However, using `geometry -0-0` doesn't work on latest Debian 12 (Bookworm) with Gnome (I'm not sure if Gnome is the culprit?). I didn't test any other configs yet. Maybe you have an idea.

Screenshot (bottom right corner of desktop):

pr-767_test

Test 1:
```
$ bin/test/fullscreen -g -0-0 &
...
xwininfo: Window id: 0x2000006 "FLTK"

  Absolute upper-left X:  1460
  Absolute upper-left Y:  528
  Relative upper-left X:  18
  Relative upper-left Y:  64
  Width: 460
  Height: 640
  Depth: 24
  Visual: 0x3d
  Visual Class: TrueColor
  Border width: 0
  Class: InputOutput
  Colormap: 0x3c (installed)
  Bit Gravity State: ForgetGravity
  Window Gravity State: NorthWestGravity
  Backing Store State: NotUseful
  Save Under State: no
  Map State: IsViewable
  Override Redirect State: no
  Corners:  +1460+528  -0+528  -0-32  +1460-32
  -geometry 460x640--18-12
```

Test 2:
```
$ bin/test/fullscreen -g -1-1 &
...
xwininfo: Window id: 0x2000006 "FLTK"

  Absolute upper-left X:  1459
  Absolute upper-left Y:  527
  Relative upper-left X:  18
  Relative upper-left Y:  64
  Width: 460
  Height: 640
  Depth: 24
  Visual: 0x3d
  Visual Class: TrueColor
  Border width: 0
  Class: InputOutput
  Colormap: 0x3c (installed)
  Bit Gravity State: ForgetGravity
  Window Gravity State: NorthWestGravity
  Backing Store State: NotUseful
  Save Under State: no
  Map State: IsViewable
  Override Redirect State: no
  Corners:  +1459+527  -1+527  -1-33  +1459-33
  -geometry 460x640--17-13
```

I wonder why both tests report `NorthWestGravity`. Shouldn't this be `SouthEastGravity`?

Additional note to my system: I'm using a Gnome/Wayland user session, i.e. our X11 programs are using XWayland on the Wayland Compositor. This may cause some issues.

Any ideas?

six809:
Thanks for that fix - I've modified the commit.

Gnome is metacity yeah?  I've just installed and tried out metacity, and I see what you mean: a weird offset that patches 2 and 3 were supposed to sort out.  Ok that clearly needs more work!

However, the "grow" and "shrink" buttons in test/resize work as expected - they resize away from the correct corner.

I'm pretty sure the gravity reported by xwininfo will be for the parent window, ie not the window FLTK itself created, but the one the window manager created to reparent the application into.  As the WM controls the horizontal and the vertical, it doesn't need to actually set the window gravity property on the parent.

So basically the patches "Set window gravity hint..." and "... fix use of initialised variable" should be good, but the two that are supposed to fix FLTK scaled dimensions so that windows sit flush against edges/corners need more attention.  I'll try and look into that.  Annoyingly it all works in evilwm ;)

six809:
Got it!  Well, I think ;)

Testing in evilwm, there were no frame extents offered using EWMH hints.
Metacity defined some.  My calculation didn't account for that.

I've packed all the changes down into two functional patches.

Tested by moving my dock around and seeing FLTK take that into account when positioning!

Albrecht-S:
> Gnome is metacity yeah?

I don't think so. The Gnome/Wayland user session (as opposed to "Gnome Classic")  runs the **_Wayland compositor_** `mutter` which is - as far as I know - functionally equivalent to a classic (X11) window manager (WM). `XWayland` is a Wayland client that acts as an `X server`, thus enabling old X11 programs to connect to the Wayland display as X clients. I believe that there is no classic X11 WM involved (running under XWayland) to do all this.

I'll test your latest commits when time permits (need to do something else right now but wanted to let you know so far).

six809:
Albrecht Schlosser wrote:
> > Gnome is metacity yeah?
> 
> I don't think so. The Gnome/Wayland user session (as opposed to "Gnome
> Classic")  runs the **_Wayland compositor_** `mutter` [...]

Ah right, my mistake.  Seems it was in GNOME 2, but got superceded.
Still, I suspect if it's doing the right thing with these hints - and
it certainly seems to be - its replacement will too.

> I'll test your latest commits when time permits (need to do something else
> right now but wanted to let you know so far).

Thank you for all your help & insight - I'm at least semi-confident that
what's there now does what it's supposed to, and at least doesn't hurt
anything else.  And there's no rush :)

..ciaran

six809:
Just to add to your spam, I've added a third patch that interacts with the above, but that I think is correct even without it.

Previously FLTK used the original width/height in negative offset calculations even if it was about to change that to user-specified values.

Albrecht-S:
@six809 I assigned myself and I really want to pursue this PR but I decided not to apply this in FLTK 1.4.0 because of potential (unwanted) behavior changes depending on WM's (window managers) as discussed above.

Please leave this PR open, we'll address this in FLTK 1.5.0.

**_Reasoning_**: we want to keep FLTK 1.4 as compatible as possible to 1.3.x and FLTK 1.5.0 is scheduled to follow shortly (2-3 months) after the release of FLTK 1.4.0.

FTR: I created two branches in my own fork for reference:
- [pr-767_x-win-gravity_v2](https://github.com/Albrecht-S/fltk/tree/pr-767_x-win-gravity_v2): your latest commits from six809/master, rebased on current fltk/master (f8380476c29d6b9b91c8ef3a61d85899e6d0fd4a)
- [pr-767_x-win-gravity_v3](https://github.com/Albrecht-S/fltk/tree/pr-767_x-win-gravity_v3): same as above, squashed to one commit

Issue #674, opened 2023-02-08
FLUID: Shell Command additional features


erco77:
It would be best if fluid saved the `Shell -> Execute Command` in the .fl file so that the commands to build the .fl project are specific to that fluid project.

I'm not really sure where it stores it right now, but every time I change fluid files, I have to keep changing the commands around because it doesn't remember them in the .fl.

If hitting "save as default" saves the commands in a global place as the default for all new fluid projects, that's OK. But if the user changes the command, it should save that change in the .fl file, and each time that file is loaded, that should be the command that appears in the dialog.

From what I can tell the commands aren't saved in the .fl file.
There's controls in the dialog window I tried messing with, but I'm not sure exactly what they do:

- a "save as default" button
- a checkbox for "use settings in .fl design files"

These could really use some tooltip elaborations.

I'm left wondering what "save as default" does -- what's it saving and where is it saving it?

The "use settings in .fl files", I'm not sure which "settings" are being used, and why would those settings not be loaded from the .fl file, would they come from somewhere else?

I thought I understood the 'save as default' to save the commands I enter "somewhere", but they'd sometimes disappear within a fluid session, reverting to older/previous settings from older projects I haven't opened in a while. (Maybe ^Z undo operations affect this?)

And when I have the "use settings in .fl files" set, I was thinking it would save my commands to the .fl, but it doesn't seem to.

It would make more sense if _everything_ in the `Shell -> Execute Command` dialog were saved in the .fl file.

Albrecht-S:
@MatthiasWM You asked for comments, so here are my general thoughts:
- I agree with Greg that it would be useful to store the shell commands (script) somewhere
- this should be done with respect to the particular project (`.fl`) file
- I also agree with Matthias that "importing" shell scripts with `.fl` files sounds dangerous
- ISTR that Matthias' first implementation was platform related, IIRC you could store commands related to the current platform (Windows, Unix/Linux, macOS)
- I don't think that the relation to a particular OS (platform) is very helpful. The user might want to store **different** commands on different **hosts** although the platform is the same.

In the following text I'm using "project" as a synonym for "`.fl` file".

My proposal is that shell scripts should be stored separate from project (`.fl`) files. If they were stored in (FLTK) user preferences we could rely on the **user** not storing hazardous commands in their own user preferences. The fluid user preferences could be stored together with the name or the path of the `.fl` file (or in a fluid prefs subdirectory, but that's an implementation detail).

[...]

PR #774, opened 2023-09-02
#674: FLUID Shell Commands rework

MatthiasWM:
I am almost finished with a whole new concept for FLUID Shell Commands. To try them out, click on `Shell > Customise...` in the main menu. The rest should be pretty self explanatory. Unless the user keeps the default shell settings, everything will be saved together with the `.fl` file. I pretty much rewrote `fluid/shell_command.cxx` in case anyone wants to see the implementation.


MatthiasWM:
See https://github.com/fltk/fltk/issues/674

Albrecht-S:
FTR: I removed the confusing `#674: `" from the title of this PR. The reference to issue #674 has been given in the comment above (and now in this one as well).

I'm not a Fluid user and thus I don't have much to say about the usage of the "shell command" feature. I can try it though, maybe...

Since @erco77 was the author of the original issue: Greg, can you please test and report your findings?

Albrecht-S:
First test: build failure (platform Linux, gcc, built with CMake, commit 32ecd88f3e63e92e24cb345b43ec9ef5063be5d2):
```
/git/fltk/fluid_shell/fluid/alignment_panel.cxx:25:10: fatal error: ../src/flstring.h: No such file or directory
   25 | #include <../src/flstring.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Suggested fix (works for me):
```
diff --git a/fluid/alignment_panel.cxx b/fluid/alignment_panel.cxx
index a641535f9..9b784d123 100644
--- a/fluid/alignment_panel.cxx
+++ b/fluid/alignment_panel.cxx
@@ -22,7 +22,7 @@
 #include <FL/Fl_Tooltip.H>
 #include <FL/fl_ask.H>
 #include <string.h>
-#include <../src/flstring.h>
+#include "../src/flstring.h"
 void scheme_cb(Fl_Scheme_Choice *, void *);
 
 Fl_Double_Window *settings_window=(Fl_Double_Window *)0;
diff --git a/fluid/alignment_panel.fl b/fluid/alignment_panel.fl
index 1639ee7c0..929562b63 100644
--- a/fluid/alignment_panel.fl
+++ b/fluid/alignment_panel.fl
@@ -91,7 +91,7 @@ decl {\#include <FL/fl_ask.H>} {private global
 decl {\#include <string.h>} {private global
 }
 
-decl {\#include <../src/flstring.h>} {private global
+decl {\#include "../src/flstring.h"} {private global
 }
 
 decl {void init_scheme(void);} {
```

Going to test now...

Albrecht-S:
Test results while working ...
```
fltk-config --compile <<codefile>>
./<<basename>>
```
This seems to be the default shell command ("all: run"). However, items `<<codefile>>` and `<<basename>>` don't seem to have a meaning. Are they intended to be replaced by the user, or should they be replaced by fluid automatically? When I run these preset commands I get errors like these:
```
sh: 1: Syntax error: end of file unexpected
sh: 2: Syntax error: newline unexpected
sh: 2: Syntax error: newline unexpected
```

There's also a dropdown (Fl_Choice) with one item (`@BASENAME@`). Is this maybe intended to be used as a placeholder which will be replaced by fluid?

The settings window that includes the "Shell script" widget should really be resizable for better user experience. IMHO.

There's a new file written (`<fluid-file>.txt`) which is unexpected and may clobber an existing file (in my test case it's `tabs.txt`). Is this only for testing, or does it have a meaning for the user? In any case it should have a better name that's less likely to clobber an existing file.

I managed to edit a shell script like this:
```
pwd
ninja tabs
bin/test/tabs
```
It works fine for me.

Last but definitely not least: where (and when) is the shell script stored? I couldn't find it anywhere. Is this still missing? When I close and restart fluid I can't find my customized shell script anymore. :-(

> To try them out, click on Shell > Customise... in the main menu. The rest should be pretty self explanatory.

Hmm, not really. I see different commands but what would I do to store exactly one shell script with the fluid (.fl) file w/o "customizing" the entire fluid thing? Sorry if this sounds weird, maybe I'm too confused because I don't use fluid regularly.

MatthiasWM:
> I'm not a Fluid user and thus I don't have much to say about the usage of the "shell command" feature. I can try it though, maybe...

Thanks for trying it out anyway ;-)

> /git/fltk/fluid_shell/fluid/alignment_panel.cxx:25:10: fatal error: ../src/flstring.h: No such file or directory

Fixed. Thanks.


> Test results while working ...
> 
> ```
> fltk-config --compile <<codefile>>
> ./<<basename>>
> ```
> 
> This seems to be the default shell command ("all: run"). However, items `<<codefile>>` and `<<basename>>` don't seem to have a meaning. Are they intended to be replaced by the user, or should they be replaced by fluid automatically? When I run these preset commands I get errors like these:

> There's also a dropdown (Fl_Choice) with one item (`@BASENAME@`). Is this maybe intended to be used as a placeholder which will be replaced by fluid?

Yes, I changed the macro syntax to `@BASENAME@`, but I have not changed the defaults yet. This would insert `alignment_panel` or whatever in the script. There need to be a bunch more of these macros, but the original FLUDI code is not yet flexible enough. Still have to write that, which will disentangle some other code.

> The settings window that includes the "Shell script" widget should really be resizable for better user experience. IMHO.

Hmm, yes, I think I can implement that.

> There's a new file written (`<fluid-file>.txt`) which is unexpected and may clobber an existing file (in my test case it's `tabs.txt`). Is this only for testing, or does it have a meaning for the user? In any case it should have a better name that's less likely to clobber an existing file.

Ah, good finding. By default, the checkbox to write the internationalisation strings file is on, but if neither catgut nor get text is selected, it simply writes text file with all strings in the .fl file. I should probably disable that.

> Last but definitely not least: where (and when) is the shell script stored? I couldn't find it anywhere. Is this still missing? When I close and restart fluid I can't find my customized shell script anymore. :-(

Yeah, sorry, I should have mentioned that. The settings will be stored with the project file. I am really going back and forth about this, because there are som many other options to store that data. As a first version, I decided that this is sufficient. 

I give a lot more options with the layout presets, but I never received a single comment on that, so I assume that it is unused/undiscovered/unnecessary, or all of the above.
 
> > To try them out, click on Shell > Customise... in the main menu. The rest should be pretty self explanatory.
> 
> Hmm, not really. I see different commands but what would I do to store exactly one shell script with the fluid (.fl) file w/o "customizing" the entire fluid thing? Sorry if this sounds weird, maybe I'm too confused because I don't use fluid regularly.

No, I understand exactly what you mean. The current plan is the addition of `import` and `export` buttons, but the basic system is very simple: if you use the default commands that come with FLUID, nothing is saved. As soon as you change anything, the settings are no longer considered default a they are fully saved with the .fl project file. If you want your own set of scripts when starting a new project, either use (yet to be written) `import`, or simply load a project file that contains the shell commands, but Is otherwise empty, and start form there.

Again, thanks for checking things out and for your feedback.

Albrecht-S:
Thanks for the update (61ed8847df5626723644bf6b9e307bb48148dce7). Storing a shell command works now (almost). I could store a shell command and it's read back from the stored .fl file and usable. However, ...

When reading a .fl file with a stored command fluid issues an error message. Sometimes it looks "correct" (the offending line contains "}" but almost always the error message issues an arbitrary (incorrect) string.

Minimal test file:
[test1.fl](https://github.com/fltk/fltk/files/12512934/test1.fl.txt)

I made a new project, added a minimal function, and configured a new shell command with default values (except `flags`).
Open this file in fluid, or run `fluid -c test1.fl` multiple times and you will see different error messages like:
```
test1.fl:32: Unknown word "`"
test1.fl:32: Unknown word "�"
test1.fl:32: Unknown word "�"
test1.fl:32: Unknown word "+"
test1.fl:32: Unknown word "E#"
```

The problem is in [fluid/file.cxx, line 339](https://github.com/MatthiasWM/fltk/blob/61ed8847df5626723644bf6b9e307bb48148dce7/fluid/file.cxx#L339). Backtrace from gdb (may be off by one line or so due to testing):
```
#0  Fd_Project_Reader::read_children (this=0x7fffffffd5d0, p=0x0, paste=0, strategy=kAddAsLastChild, skip_options=0 '\000') at /git/fltk/fluid_shell/fluid/file.cxx:340
#1  0x000000000045a0f6 in Fd_Project_Reader::read_project (this=0x7fffffffd5d0, filename=0x7fffffffdbfa "/data/git/fltk/fluid_shell/test/tabs1.fl", merge=0, strategy=kAddAsLastChild) at /git/fltk/fluid_shell/fluid/file.cxx:401
#2  0x000000000045926d in read_file (filename=0x7fffffffdbfa "/data/git/fltk/fluid_shell/test/tabs1.fl", merge=0, strategy=kAddAsLastChild) at /git/fltk/fluid_shell/fluid/file.cxx:64
#3  0x00000000004784a9 in main (argc=2, argv=0x7fffffffd788) at /git/fltk/fluid_shell/fluid/fluid.cxx:1938
```

HTH.

src/filename_absolute.cxx
```diff
@@ -45,30 +45,44 @@ static inline int isdirsep(char c) {return c == '/';}
...
+ \param[in]  from relative filename
+ \param[in]  base relative to this absolute path
+ \return 0 if no change, non zero otherwise
+ */
+int                                     // O - 0 if no change, 1 if changed
+fl_filename_absolute(char       *to,    // O - Relative filename
+                     int        tolen,  // I - Size of "to" buffer
+                     const char *from,  // I - Absolute filename
+                     const char *base) { // I - Find path relative to this path
+  return Fl::system_driver()->filename_absolute(to, tolen, from, base);
```
Albrecht-S, line 69: Same here, how can a path be "absolute" to another path ...

More of such issues below. I'm **_absolute_**ly confused. Pun intended.

Albrecht-S: After reading again... Maybe the issue is only confusing wording and the comment(s)?

Probably better: "Returns an absolute filename from another filename that is relative to a base directory."

Should we also mention that the "base directory" must be an absolute path (if given), otherwise it would be the current directory, or something like that.

MatthiasWM:
> When reading a .fl file with a stored command fluid issues an error message. 

Great find! Thank you. I was missing a `goto` statement.

src/filename_absolute.cxx
MatthiasWM: Yes, it's confusing. I had to reread this myself a fe times, and even though the code worked, the docs are wrong. Also, documenting parameters twice, once in the Doxy comment, and a gain two lines below in code, is irritating, so I deleted it. Hoping that then phrasing is better.

Albrecht-S:
:+1 Works well now with my saved shell commands. Good job, thanks.

Minor nitpick, just as a reminder: resizable editor window for shell commands still missing.

MatthiasWM:
Thanks. The reliable editor is last when all the other widgets are in place.

Albrecht-S:
Another noteworthy point: I imagine two scenarios where it would be good or even necessary to remove the personal shell command settings and store the `project.fl` file without shell commands:

1. A developer working on fluid or any other fluid files in the FLTK distro (`src`, `test`, `examples`).
2. A programmer working on a (maybe public) project that will be deployed (e.g. stored on GitHub).

It would be useful to have some shell commands while **working** on such projects but we should have a way to store the files w/o shell commands. If I understood correctly that would be the case if I reset the shell command(s) to the default values. Well, although this is possible, users (and FLTK devs) would likely forget to remove the shell commands before committing the changes at some time.

What to do about this? I'm not sure. Some thoughts:

1. We could have an "export" feature that writes the `.fl` file with default settings.
2. We have already an "update" command that is intended to update the `.fl` file to the current fluid version (`fluid -u <file>`). In the future this command could strip the shell commands from the `.fl` file.
3. Just in case someone wants to update but **keep** the shell commands we could also add a `-k` (keep) switch which would not remove the shell commands when used together with `-u`.
4. Yet another option would be not to store the shell commands in the `.fl` file, i.e. either in user preferences or in a file besides the `.fl` file (same name, other extension). If we had `project.fl` we might want to store the shell settings in `project_shell.txt` or `project.something-else` - as discussed elsewhere in fluid related issues for security or other reasons. Maybe we could use `project.fl_shell` and use Git's `.gitignore` feature to exclude these files from committing and pushing them to the repo.

After thinking about this I might prefer option 4 because it has several advantages: it prevents accidentally storing shell commands in the fluid file in the first place but project files can still be moved and renamed together with the shell settings by the developer. Used together with the `@@...@@` variables for shell commands this could become a strong feature.

Storing shell command in preferences might make this more difficult. However, I imagine storing some "personal default shell commands" (templates) in user preferences would be useful. Start a new project, configure the shell commands from your preferences, and go.

MatthiasWM:
Thanks for your thoughts. Solution 4 is indeed what I am preparing.

Every shell command can now have a tool location. Currently supported are the user settings and inside the .fl file. I wanted to add two special files to the project folder if the user chooses so. `fluid.prefs` stores all tool settings (alignment, shell commands) that are useful for all .fl files in this folder. So whenever we read a .fl project file, we also read `fluid.prefs`. This file should be stored in the repository.

The second file is `fluid.user.prefs` and should *not* be stored in the repository, but should be tagged in `.gitignore`, allowing users to add their own shell commands as needed, but not to be published.

Apart from all that, there will be actions to import and export shell commands to and from the list.

I still need to find fltk-config, and I need a few sample commands to show the user what's possible and give a nice base to build upon. And a documentation page.

PR #796, opened 2023-10-15
Add virtual void Fl_Graphics_Driver::small_circle() to support fl_draw_circle()

ManoloFLTK:
I propose to limit the cases when function `fl_draw_circle()` replaces `fl_pie()` by small rectangles to platforms where it is necessary. Platforms that use modern drawing APIs such as Cairo, Quartz and PostScript don't need small circles to be emulated by small rectangles. The new graphics API FLTK uses under Windows, GDI+, doesn't need it either. Only X11 and GDI do. The introduction of a new virtual member function to class `Fl_Graphics_Driver`, called `small_circle()`, allows to implement that. It also removes the unsolved problem mentionned in the source code of `fl_draw_circle()` of supporting HighDPI screens such as retina displays.

### DIFF
--- a/src/Fl_Graphics_Driver.cxx
+++ b/src/Fl_Graphics_Driver.cxx
@@ -600,6 +600,14 @@ void Fl_Graphics_Driver::arc(int x, int y, int w, int h, double a1, double a2) {
 /** see fl_pie() */
 void Fl_Graphics_Driver::pie(int x, int y, int w, int h, double a1, double a2) {}
 
+/** see fl_draw_circle() */
+void Fl_Graphics_Driver::draw_circle(int x, int y, int d, Fl_Color c) {
+  Fl_Color current_c = color();
+  if (c != current_c) color(c);
+  pie(x, y, d, d, 0., 360.);
+  if (c != current_c) color(current_c);
+}
+
 /** see fl_line_style() */
 void Fl_Graphics_Driver::line_style(int style, int width, char* dashes) {}
 
@@ -931,6 +939,45 @@ void Fl_Scalable_Graphics_Driver::pie(int x,int y,int w,int h,double a1,double a
   h = floor(y+h) - yy;
   pie_unscaled(xx, yy, w, h, a1, a2);
 }
+
+
+void Fl_Scalable_Graphics_Driver::draw_circle(int x0, int y0, int d, Fl_Color c) {
+  Fl_Color saved = color();
+  color(c);
+
+  // make circles nice on scaled display
+  float s = scale();
+  int scaled_d = (s > 1.0) ? (int)(d * s) : d;
+  
+  // draw the circle
+  switch (scaled_d) {
+      // Larger circles draw fine...
+    default:
+      pie(x0, y0, d, d, 0.0, 360.0);
+      break;
+      
+      // Small circles don't draw well on many systems...
+    case 6:
+      rectf(x0 + 2, y0, d - 4, d);
+      rectf(x0 + 1, y0 + 1, d - 2, d - 2);
+      rectf(x0, y0 + 2, d, d - 4);
+      break;
+      
+    case 5:
+    case 4:
+    case 3:
+      rectf(x0 + 1, y0, d - 2, d);
+      rectf(x0, y0 + 1, d, d - 2);
+      break;
+      
+    case 2:
+    case 1:
+      rectf(x0, y0, d, d);
+      break;
+  }
+  color(saved);
+}
+
 
 void Fl_Scalable_Graphics_Driver::line_style(int style, int width, char* dashes) {
   if (width == 0) line_width_ = int(scale() < 2 ? 0 : scale());
--- a/src/drivers/GDI/Fl_GDI_Graphics_Driver.cxx
+++ b/src/drivers/GDI/Fl_GDI_Graphics_Driver.cxx
@@ -48,10 +48,9 @@ int Fl_GDIplus_Graphics_Driver::antialias() {
   return active;
 }
 
-#endif
-
-
-#if USE_GDIPLUS
+void Fl_GDIplus_Graphics_Driver::draw_circle(int x, int y, int d, Fl_Color c) {
+  Fl_Graphics_Driver::draw_circle(x, y, d, c);
+}
 
 int Fl_GDIplus_Graphics_Driver::gdiplus_state_ = Fl_GDIplus_Graphics_Driver::STATE_CLOSED;
 ULONG_PTR Fl_GDIplus_Graphics_Driver::gdiplus_token_ = 0;
--- a/src/fl_draw.cxx
+++ b/src/fl_draw.cxx
@@ -592,67 +592,27 @@ void fl_draw_check(Fl_Rect bb, Fl_Color col) {
 } // fl_draw_check()
 
 /**
-  Draw a potentially small, filled circle as a GUI element.
+  Draw a potentially small, filled circle using a given color.
 
-  This method draws a filled circle, using fl_pie() if the given diameter
-  \p d is larger than 6 pixels after scaling with the current screen
-  scaling factor (i.e. "physical" pixels).
+  This function draws using \p color a filled circle bounded by rectangle <tt>(x, y, d, d)</tt>.
+ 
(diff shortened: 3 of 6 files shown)


MatthiasWM:
Thanks for fixing this more globally! I would change the method name from `small_circle` to `circle`, simply because it draws any size circle, and it is not really important for the caller to know if there is any special handling for a specific driver, right?

MatthiasWM:
So this code does not implement the different driver specific calls yet, right? This would be s second PR. If I understand the code right, you have the pie-based drawing as the fallback in Fl_Graphics_Driver and the fl_rect based one in Fl_Scaled_Graphics_Driver. I suggest you make pie the deaf for everything and only implement the fl_rect based drawing for non-antialiasing drivers (X11, GDI?), maybe even call true circle functions if they exist for a driver?

Albrecht-S:
Thanks, @ManoloFLTK from me as well. I agree with @MatthiasWM to suggest to rename the new graphics driver method just `Fl_Graphics_Driver::circle(...)`.

Another option would be to overload `fl_pie(x, y, d, color)` instead, i.e. one without the parameters `a1` and `a2` but with the additional `color` argument that would be the same as calling `fl_draw_circle()` and would therefore obsolete the entire `fl_draw_circle()` function because we could call `fl_pie(x, y, d, color)` in the widget drawing code directly.

`fl_pie(x, y, d, color)` looks like a natural extension of `fl_pie(x, y, w, h, a1, a2)`.

The latter is just a thought...

src/fl_draw.cxx
```diff
@@ -592,67 +592,21 @@ void fl_draw_check(Fl_Rect bb, Fl_Color col) {
...
-void fl_draw_circle(int x0, int y0, int d, Fl_Color color) {
-
-#define DEBUG_DRAW_CIRCLE (0) // bit 1 = draw bounding box (green)
-
-  Fl_Color saved_color = fl_color();
-
-#if (DEBUG_DRAW_CIRCLE & 1)
-  fl_rectf(x0, y0, d, d, FL_GREEN);
-#endif
-
```
Albrecht-S, line 620: Please leave the debugging code and save/restore the color code here as noted above.

src/Fl_Graphics_Driver.cxx
```diff
@@ -932,6 +940,50 @@ void Fl_Scalable_Graphics_Driver::pie(int x,int y,int w,int h,double a1,double a
...
 
+
+void Fl_Scalable_Graphics_Driver::small_circle(int x0, int y0, int d, Fl_Color c) {
+  Fl_Color saved = color();
+
+#define DEBUG_DRAW_CIRCLE (0) // bit 1 = draw bounding box (green)
+#if (DEBUG_DRAW_CIRCLE & 1)
+  fl_rectf(x0, y0, d, d, FL_GREEN);
+#endif
+  color(c);
```
Albrecht-S, line 951: The above debugging code **must** be in the main function `fl_draw_circle()` because it is intended as a debugging aid for widget developers. It was a great help for me when I rewrote widgets to use fl_draw_circle() - not only for small circles.

Therefore it must not be only in the special driver method that uses rectangle drawing but in the main function.

Saving and restoring the color setting should also be done in `fl_draw_circle()` and could be removed here.

src/drivers/GDI/Fl_GDI_Graphics_Driver.cxx
```diff
@@ -48,10 +48,9 @@ int Fl_GDIplus_Graphics_Driver::antialias() {
   return active;
 }
 
-#endif
-
-
-#if USE_GDIPLUS
+void Fl_GDIplus_Graphics_Driver::small_circle(int x, int y, int d, Fl_Color c) {
+  Fl_Graphics_Driver::small_circle(x, y, d, c);
+}
```
Albrecht-S, line 54: Didn't you mention that the GDI+ driver doesn't need the special handling? If that is so, do we really need this overload that calls the (default) implementation back? Can't we just remove this one?

src/fl_draw.cxx
```diff
@@ -592,67 +592,21 @@ void fl_draw_check(Fl_Rect bb, Fl_Color col) {
 } // fl_draw_check()
 
 /**
-  Draw a potentially small, filled circle as a GUI element.
+  Draw a potentially small, filled circle using a given color.
```
Albrecht-S, line 596: As Matthias mentioned, the new function would draw circles of any size, hence this should read:
```
Draw a filled circle using a given color.
```

src/fl_draw.cxx
```diff
@@ -592,67 +592,21 @@ void fl_draw_check(Fl_Rect bb, Fl_Color col) {
...
+  filled rectangles if the converted diameter is ≤ 6 pixels.
 
-  If the diameter is 6 or smaller after scaling it approximates a filled circle
-  by drawing several filled rectangles, depending on the size because fl_pie()
-  might not draw well on many systems for smaller sizes.
-
-  \param[in]  x0,y0   coordinates of top left of the bounding box
-  \param[in]  d       diameter == width and height of the bounding box
+  \param[in]  x,y   coordinates of top left of the bounding box
+  \param[in]  d       diameter == width and height of the bounding box in FLTK units
```
Albrecht-S, line 605: Sorry, nitpicking: please align descriptions for better reading.

Albrecht-S (changes requested):
Thanks for the proposal, please check my comments and apply changes (or discuss why not).

ManoloFLTK:
> So this code does not implement the different driver specific calls yet, right? This would be s second PR.…

No. This PR is complete. The need for special processing for small circles exists only with X11 and GDI (GDI strictly, that is, not with GDI+). The hierarchy below class `Fl_Graphics_Driver` is such that X11 and GDI both derive from `Fl_Scalable_Graphics_Driver` and `Fl_GDIplus_Graphics_Driver` derives from `Fl_GDI_Graphics_Driver`. Implementing the normal circle drawing in `Fl_Graphics_Driver`, the special processing in `Fl_Scalable_Graphics_Driver`, and having `Fl_GDIplus_Graphics_Driver`'s member function call that in `Fl_Graphics_Driver` produces the desired overall processing. That is what the present PR does.

src/drivers/GDI/Fl_GDI_Graphics_Driver.cxx
ManoloFLTK: This should remain because of the hierarchy of classes below `Fl_Graphics_Driver`: `Fl_GDIplus_Graphics_Driver` derives from `Fl_GDI_Graphics_Driver` which derives from `Fl_Scalable_Graphics_Driver`. We have to undo in `Fl_GDIplus_Graphics_Driver` what is done in `Fl_Scalable_Graphics_Driver` to get the desired behavior at the GDI+ level.

src/Fl_Graphics_Driver.cxx
ManoloFLTK: Committed in 80d102d

src/fl_draw.cxx
ManoloFLTK: The "potentially small" expression is Albrecht's when introducing function `fl_draw_circle()` in 1.4. I think it makes sense because this function does draw all circle sizes but exists only because small circles may need special processing.

src/fl_draw.cxx
ManoloFLTK: Committed at c855d93.

src/drivers/GDI/Fl_GDI_Graphics_Driver.cxx
Albrecht-S: OK, I understand.

src/fl_draw.cxx
Albrecht-S: OK, at that time when I introduced it, that was something I wanted to emphasize. But anyway, I'm fine with leaving it.

ManoloFLTK:
@MatthiasWM @Albrecht-S Thanks for your comments about this PR. I believe to have now handled them all. Please, confirm or ask for more changes here.

Albrecht-S (approved):

Albrecht-S:
I formally "approved changes" via GitHub's function. I believe we're good to go now. Thanks for your great work!

PR #800, opened 2023-10-18
Fl_Terminal widget

erco77:
This PR introduces the Fl_Terminal widget with full support for Unicode and as many ANSI/xterm features as I thought I could support without going too far into the weeds.

The widget is API compatible with Fl_Simple_Terminal, and expands the API quite a bit for more full on terminal capabilities.

In this branch, all the examples/test programs have been refitted to use it.. **Except fluid**.

### Fluid Implications
My original plan was to remove all traces of Fl_Simple_Terminal and replace it with Fl_Terminal. But this might have implications in fluid that should perhaps involve Matt and others for design considerations. For instance perhaps the .fl file loader should attempt to refit references to Fl_Simple_Terminal on the fly to try to make the transition as smooth as possible for any 1.4.x users out there that are using Fl_Simple_Terminal in their fluid files.

For now I've left Fl_Simple_Terminal in place so fluid can continue to use it, and let the two widgets coexist until we figure out how to refit fluid in the least disruptive way for users possible.

So with this, fluid is really the only part of FLTK still using Fl_Simple_Terminal.

### Special Technical Documentation
There's full docs as usual, plus a special "Technical Document" that describes the internal design of the widget, useful for any devs needing to pick it up. I wanted to be sure there was a decent design document, so no one has to take guesses as to the design intentions.

This technical document currently shows up in the top level of the html index. I wasn't quite sure where to weave it in such that it appears with other, hopefully forthcoming widget technical documents so devs can get up to speed quickly on complicated widgets (like Fl_Text_Editor, Fl_Help_View, etc).


Albrecht-S:
Hi Greg, thanks for making Fl_Terminal and this PR.

I found some formal patches (besides your fix in 124e96514b2248945c27b8d186bba8459d37922e) which would need to be applied for the final merge:
- [gitignore.diff](https://github.com/fltk/fltk/files/13029563/gitignore.diff.txt)
- [makedepend.diff.txt](https://github.com/fltk/fltk/files/13029577/makedepend.diff.txt)

That's so far only from building your branch. I'm going to test as soon as possible.

erco77:
> I can add code to reinterpret an Fl_Simple_Terminal into an Fl_Terminal, if that is useful for you.

Ya, I wasn't sure how to handle the situation where a user, perhaps months ago, created a .fl file
that used fluid's New -> Text -> Simple Terminal.

I figure besides trivially changing fluid to just replace all instances of Fl_Simple_Terminal -> Fl_Terminal as well as rewriting the names that appear in the .fl file, it might be good to **have the new fluid's .fl file parser recognize the old "simple terminal" names, and automatically re-write them to just "terminal".**

I wasn't sure how to do that parsing trick properly, and wasn't sure if any of those alternate names (ID/type_name/alt_type_name) might somehow appear in the .fl file, and if so, "translate" those too.

So if a user had previously created an .fl file using "New -> Text -> Simple Terminal", the new fluid would be clever enough to translate the old Fl_Simple_Terminal names in the file to the new Fl_Terminal names, automagically "fixing" the user's .fl file.

fluid wouldn't have to do anything to user code (e.g. `code0, code1`..), as Fl_Terminal's API should be "similar enough" that user code like: `code0 {o->ansi(true);}` would still work fine. The only way user code could trip up is if they referred to Fl_Simple_Terminal by name, they'd have to change that code manually.

erco77:
@MatthiasWM Yes, the API is much larger now, as it supports a lot of features the old one could not. But the API is backwards compatible.

Keep in mind Fl_Simple_Terminal was added in 1.4.x and will probably be /removed/ in favor of Fl_Terminal.
I say probably; the only reason to keep Fl_Simple_Terminal is if there's some weird need.

But since Fl_Simple_Terminal was introduced in 1.4.0 (it's not in 1.3.x), and 1.4.0 has not released yet, we don't need to keep it if it's being superseded.  AFAIK we can break API within a release that hasn't released yet. If we really need to, we can either keep Fl_Simple_Terminal and mark it obsolete with a \#warning, or make a wrapper around Fl_Terminal. But I think we can just pull that tooth safely. Up to Albrecht really.

Albrecht-S:
> since Fl_Simple_Terminal was introduced in 1.4.0 (it's not in 1.3.x), and 1.4.0 has not released yet, we don't need to keep it if it's being superseded. AFAIK we can break API within a release that hasn't released yet.

I suggest to drop `Fl_Simple_Terminal` if/when `Fl_Terminal` will be pushed to the main repo. We have the same issue with `Fl_String` and `Fl_Int_Vector` which **will be removed** from the public API soon. They have been in 1.4 for a while but never in 1.3.

erco77:
Some low priority items I noticed while revisiting Fl_Terminal:
```
// FIXME: While dragging a selection, hitting shift stops the selection.

// FIXME: In test applications (e.g. test/table), filling terminal with text and then enlarging width
          of window cause an extra blank line appear at bottom and scrollbar tab shrinks in size.

// FIXME: in test/terminal, some input fields/buttons are lacking tooltips

// FIXME: Extending selections to the left/up (using shift-click) has some weird selection behavior.
         Seems to loose its sense of the selection end point.

// TODO: horizontal scrollbar option.
         Probably not as much of a hassle as one might think;
         It just needs to interface between the ring buffer's width <-> display.
         
// TODO: double clicking text should make word selection,
         and double-click-and-drag should word-enlarge selection
```
I'll edit+add to this comment if I find other things.

Albrecht-S:
> There should probably be an option that allows for a popup menu with "Copy/Paste" options.

Wouldn't the usual ctrl/cmd + C/V be enough? If we can select text already, wouldn't that be more common?

erco77:
@Albrecht-S I suggested it as more of a standard thing people come to expect from user interfaces, but yeah, entirely optional.

I'm just not used to seeing apps "do nothing" when you right click a text field; usually there's at minimum a copy/paste option, with paste grayed out for read-only contexts.

During testing I was just pretending the FLTK terminal was a real one, and made note of behaviors that seemed absent, and a right click copy menu was one. Come to think of it, a way to access 'Settings' is another, since the terminal has so many settings. But I suppose then it gets tricky of what options a user should be allowed to change.. probably better to not provide anything and let the app do that work itself.

So in the name of keeping things "light", I should leave it out; apps needing right click popups have come to expect they'll need to overlay an invisible Fl_Menu_Button (or derive a class and overload handle()) to get popup menus.

MatthiasWM:
The right-click menu is on the todo list for all text based widgets. If we do it, we need to decide what goes into that menu. Cut/copy/paste obviously, but Undo/redo? Delete? What if the user wants to add a spell checker menu item? And different items per input field? We also need to provide a way to replace the item text (and shortcut?) for a different languages.

It's of course doable, but it takes some thought for a good API.

Albrecht-S:
@erco77 The entire terminal app in a giant fluid file seems to be much too large, at least for my personal use. Finding (and rewiewing) code is extremely hard. Would you mind writing the code once and dropping the fluid file, i.e. committing the .cxx and .h files in instead?

RFC: what do others think?

erco77:
@Albrecht-S It's big, but I guess I'm just used to large fluid managed apps. When the entire app is just a big layout of widgets, I just use fluid like an IDE for the whole thing. I assume when someone picks it up, they can just look at the generated source code, then use fluid to make the mods. I guess it's just something I'm used to.

But I could definitely split the UI and back end test code into separate files if that'd help, similar to test/keyboard and test/mandelbrot, IIRC.

If I split it up, it might take me a while to do that. I'd also need some help doing the proper cmake/configure mods for that; I imagine that a mix of committed .fl and .cxx makes the build/clean instructions more.. interesting, especially when 'clean' has to remove source files.

But the UI layout has too many fields to NOT use fluid, so an .fl file needs to be involved at least for that.

Albrecht-S:
> I assume when someone picks it up, they can just look at the generated source code, then use fluid to make the mods.

That's exactly my problem because I'm not used to find the right place in the Fluid UI for a particular piece of code I'm interested in. See issue #804 but maybe someone (you?) can teach me how to work w/o help from a `find` function. How do you do that if it's not your project which you know well? The sheer size of almost 2000 lines of fluid project code makes me feel uncomfortable.

> I'd also need some help doing the proper cmake/configure mods for that...

With CMake in the test folder it's easy to do that. You can add either `.fl` or `.cxx` files to the sources parameter if you put them in quotes and separate them by `';'`. No issues with `make clean` etc., everything else works as expected by CMake magic. `fractals` and CubeView are good examples that uses both the sources and the libraries argument for more than one "object":
```
  CREATE_EXAMPLE (fractals "fractals.cxx;fracviewer.cxx" "fltk_gl;fltk")
  CREATE_EXAMPLE (CubeView "CubeMain.cxx;CubeView.cxx;CubeViewUI.fl" "fltk_gl;fltk")
```

Makefiles are certainly more difficult (not my strong part) but there are examples as well.

My first thought was simply to "write code" and put `terminal.cxx` and `terminal.h` under version control instead of `terminal.fl`. Are you also using icons or images? Such binary blobs would better be stored as files rather than in fluid code but maybe that's not the case?

Sorry, I've been to busy with other tasks so that I couldn't further look into your Fl_Terminal project.

erco77:
Yes, fluid has long needed a search (or find) dialog, so the few times I'd need to find something,  I'd open the generated source code in an editor, search for the line to find what function it's in, then open that function in fluid.

Either that, or there are rare times I'd simply hand edit the code in the .fl file, and reload it in fluid, being aware of the certain characters that have to be escaped when inside .fl files.

In the case of the test/terminal program, the bottom 20% of the source is the GUI layout generated by fluid, and that is best managed inside the GUI, and can be ignored in the .cxx, and the main_window() function 'collapsed' inside fluid, so one doesn't have to see all the UI widgets.

The middle 40% are all the ansi/xterm escape codes for the tests, each test page is a separate function called test_xxx(), so it should be easy to find the test page and match it to the function name that shows up in fluid.

The test functions chained together as described in the "DEVELOPER README" you see at the very top of the generated .cxx file:

image

Beyond that there's not much to it, but there's no short way around all that inline escape sequence stuff, short of moving them to another file. When editing the terminal test program, I found myself mostly just tweaking the escape sequences and surrounding text.

Albrecht-S:
I don't think that the new Fl_Terminal has wrap mode, does it?

If not this might be called a regression if it will replace Fl_Simple_Terminal everywhere. I know Fl_Terminal has its benefits, but maybe we need at least one or more of the following features (if they don't exist yet):
- horizontal scrollbars
- a way to scroll the text left and right, maybe with cursor keys
- wrap mode

Test case (just one I noticed but there may be more):
- run test/demo
- right-click somewhere at the bottom of the window
- click on "Show debug terminal"
- shrink the terminal width

The old Fl_Simple_Terminal would show all contents, whether or not the window is wide enough whereas the new Fl_Terminal suppresses the "wrapping part". In this specific case you could resize the window, but this is not the point.

Albrecht-S:
@erco77 Thanks for your explanation of the Fluid project of Fl_Terminal. Sure, I can find my way through all of this, but couldn't we simplify it for the general user (developer)?
OTOH, if nobody else thinks this should be changed, never mind.

erco77:
_Just to be clear, test/terminal is a QC checker, a stress test and to check for common regressions, not for the "general user", or even a developer to use to learn how to do things. examples/simple_terminal.cxx is the kind of thing for that. _

Because it's a stress tester for xterm sequences, the "walls of text" riddled with ugly escape sequences are gonna be large and ugly; I tried but could not come up with a way to make that pretty; escape sequences can't be padded with whitespace to make things line up.

As mentioned I could move all the escape sequence tests into a separate .cxx/.h file, which will shorten the .fl file; those escape sequence tests don't really need to be within fluid. If I do this, can I do it after the merge? I just don't know how long that will take me to do; I'd like to avoid getting out of sync for these PR's.

erco77:
@Albrecht-S Ya, no wrap mode. It can be added though; I'd need to (a) record where CRLFs occur during normal streams of line output so lines can be broken and wrapped properly throughout resizes, and (b) keep those CRLF locations updated when overstrike happens (due to cursor repositioning) and cleared by clear-to-end-of-line operations.

That won't be easy, but it can be done. I remember thinking about how that works, and it gets tricky due to cursor repositioning. I recall early versions of OSX's terminal did not support that either, and it took em a few years to get around to it. It's one of those things you 'add later', lol.

Albrecht-S:
Greg, please forget what I wrote about the fluid project, I withdraw my request. Let's better concentrate on other things that are more (or really) important. I understand what test/terminal is good for and it's hard to test or review anything if you need to work through the test environment. That was my point. Again, let's concentrate on other things...

erco77:
Ya, I think necessary features is more important for sure.

If the horiz scrollbar is a showstopper I can add it. I just wish I wouldn't have to be 'punished' by the repeated PR resyncing process; the longer it takes me to do something, the worse the "punishment" lol.

Albrecht-S:
OK, just for communication, let's call the issue the "horizontal scrolling" problem. The only thing I want to make sure is that we don't replace `Fl_Simple_Terminal` with `Fl_Terminal` and users can't see full the contents of warnings and other text appearing in the terminal window.

For comparison: A similar issue is in browser windows when text doesn't wrap (e.g. `<pre>`) and the horizontal scrollbar is way too far down to reach it w/o shifting the text you want to see up out of the window. In Firefox, for instance, the left and right keys can help to "scroll horizontally".

Whatever we do, users must be able to see the full contents of lines that don't fit in the display. That's all, I guess. How we do that doesn't matter. Wrap mode, scrollbar, arrow key "navigation", whatever.

Is this a showstopper for merging `Fl_Terminal` to avoid the "resync punishment"? I don't know but I'd say as long as we can **_guarantee_** that this will be fixed before the release of 1.4.0 it would be OK for me.

erco77:
@Albrecht-S They won't miss any text; super long lines of text will wrap around the screen as one expects. For instance I created a file /var/tmp/foo, pasted your message text into it, in VI held down Shift-J to bring it all on one line, saved, then in test/terminal put /var/tmp/foo in the "Show File" prompt and hit Enter, and it shows the long line of text this way:

image

If you then shrink the window, it clips off the text in the display, but the buffer is still there, so they can get it back by resizing the window larger again.

And here, when the window is sized smaller, I should add the horiz scrollbar to allow shifting the view around, so one doesn't have to resize the window larger again to see the clipped text.

What /doesn't/ work currently, and is unlikely to be added without a lot of effort is if you resize the window larger or smaller than it was, and have the long lines of text rewrap, as it does in modern terminals; it doesn't "rewrap the text" because it doesn't keep a record of that it was all one line.

That's kinda tricky to keep track of when your buffer is block oriented. And indeed, we're currently spoiled by the fancy modern terminals we have that do all kinds of fancy rewrapping; before 2005 or so, none of the GUI terminals did that. IIRC, OSX was the first, and then everyone was shamed into supporting it because it was so cool. I think I know how that works, but it's a bit tricky. I think it involves an extra per-character bit to indicate where the invisible line endings are, so that line wrapping can be calculated based on that.

erco77:
(I heavily edit the above; took me a few tries to get the text right)

I should add; a case where it gets ugly is if a line is printed to the terminal, say:

      Speak softly and carry a big stick.

..and a crlf follows the period. The end of line (EOL) marker bit would be set on the period character, indicating it was the last character on the line, and is a wrap point.

But then someone sends a few escape characters that plots an 'X' character floating off the end of the line, e.g.:

      Speak softly and carry a big stick.       X

What this should do is clear the EOL bit and move it to where the 'X' is, so it becomes the new end of line for that line.

Then someone uses more escape sequences to move the cursor just past the 'a' in 'carry' and sends a Clear-to-End-Of-Line (CLEOL, or `\033[K`) , so you end up with:

    Speak softly and ca

This should clear the EOL bit where the 'X' was and move it to the 'a'.
This logic has to be extended to all the escape sequences, and involves a bit of careful thinking to avoid problems.

So in short, the devil's in the details on that one. Not to mention the logic in the resize code to calculate line wrapping.

erco77:
Anyway, all that was just to be clear about what is easy to add for the scrollbar behavior, and what is hard to add, and why it's hard.

MatthiasWM:
> > There should probably be an option that allows for a popup menu with "Copy/Paste" options.
> 
> Wouldn't the usual ctrl/cmd + C/V be enough? If we can select text already, wouldn't that be more common?

Ah, I found the link: [749](https://www.fltk.org/str.php?L749+P0+S-2+C0+I0+O0+E0+V1.+Q). The first STR for a rich mouse button pulldown menu for text widgets is form March 2005. ;-)

erco77:
@MatthiasWM Ha, almost 20 years coming up. One can have grandchildren waiting for features, lol. Looks like I'm quoted in the OP, and 4 years later I followed up with a cheat sheet implementation.. ah, sweet memories.

erco77:
Just curious -- is this on hold for the horiz scrollbar?

I just want to point out word wrapping does work in that long lines printed will wrap at the edge of the screen.
Horiz scroll is only needed if you shrink the window after a long line is printed, the long line displays clipped but is still there and can be selected even though it's offscreen, etc. So no data loss, if that's the concern.

But if the PR's on hold for that feature, let me know and I'll try to add it when I can find some available cycles; it's tough, I'm super busy and a bit behind..

erco77:
@Albrecht-S I should be able to guarantee the horiz scrollbar can be ready by the 1.4.0 release.

Albrecht-S:
@erco77 Greg, I'm currently trying to review the code. So far I did some experiments with the horizontal scrolling/wrapping issue and came to the conclusion that it might not be as big a problem as I thought before (thanks for your comments on this above, too). But anyway, having a horizontal scrollbar might be useful.

I'll try to add more tomorrow...

erco77:
Ya, if I remember correctly a horiz scrollbar's position should be able to be included in the draw code to offset the display horizontally, the same way the vertical scrollbar offsets vertically. That part should be trivial. The calculation for the scrollbar tab size should be easy too; simply the buffer's width vs. the display width. Don't have to worry about how wide the widest line in the display is.
What might be weird tho, I'm anticipating, is if the user widens the window, widening the buffer, then makes it smaller again, the horizontal scrollbar will appear and remain on, until the window is returned to the wide (or wider) width. That might be disconcerting, as one is then stuck with the horizontal bar eating up vertical screen space from then on, unless one resizes to the widest setting to make it disappear again. In which case I might have to keep track of the widest line, just so the horiz scrollbar can be hidden if nothing is worth seeing offscreen. It can be done, but then a calculation for the widest line would have to be maintained, which is annoying.

Albrecht-S:
> It seems best that a terminal never have a horizontal scrollbar (I've never seen one I don't think).

Yes, that's what I see with real terminal programs, but these programs usually have wrap mode **and** reformat the lines when the window width is enlarged or shrunk. Anyway, let's keep this horizontal scrolling issue as-is and wait if we get complaints. And to get complaints we should IMHO merge your `Fl_Terminal` branch as soon as possible. But see below for questions and suggestions.

Albrecht-S:
I tried to review the code but this is an overwhelming amount of code. Awesome! Thanks!

Some personal comments and questions follow. These _should not hold the merge_ but I'd appreciate if you answered the questions and, if useful, modified code and/or docs.

Some methods and functions are undocumented (maybe even public ones, but for sure some protected ones) which makes it even harder to see what's intended and what's going on. I didn't review the documentation systematically though, this could be done later (i.e. after the merge).

1. I don't grok the real usage of the `CharStyle` class. Do we have per-character styles or not, i.e. is the `CharStyle` stored with each character? If yes, there's a comment that says:
>  Will likely never implement as part of this widget:
> ...
>     - per-character font faces + sizes (font face/size is global only)

2. Another comment in `Fl_Terminal.H` says:
> // TODO: CLEAN UP WHAT'S PUBLIC, AND WHAT SHOULD BE 'PROTECTED' AND 'PRIVATE'
> //       Some of the public stuff should, quite simply, "not be".

If this comment was true it should be fixed before the release (otherwise the comment could be removed).

3. From `FL/Fl_Terminal.H`:
> protected:
>  int   scrollbar_w(void) const;

This method is protected and has no documentation. I believe it could be **public**, should be **renamed** to `scrollbar_width()`, and **needs documentation**.
Rationale: `scrollbar_w` is too cryptic and we have precedence methods that use `scrollbar_width()` in other widgets.
**_Note:_** some of these other methods are deprecated and should be removed or re-implemented differently but I think `[public] int scrollbar_width(void) const;` should return the **actual scrollbar width** (which it does), as opposed to `scrollbar_size()`. I'll look into the deprecated methods myself and I'll try to fix them if possible. Or maybe remove them as already documented.

4. Please check the copyright notes. I suggest (in this order):
```
// Copyright 2022 by Greg Ercolano.
// Copyright 2023 by Bill Spitzak and others.
```
rather than the opposite order with `1998-2011 by Bill Spitzak and others` - if that was what you intended, of course. It doesn't make sense to have a copyright `1998-2011` if the widget didn't exist yet.

5. There are lots of `NOTUSED` comments. Can you remove these comments before the merge? Unless there are intentions to implement what's now commented out.

6. Constructor documentation and code:
```C++
/**
  The constructor for Fl_Terminal.
  This creates an empty terminal with defaults:
  - white on black text; see textfgcolor(Fl_Color), textbgcolor(Fl_Color)
  - rows/cols based on the \p W and \p H values, see display_rows(), display_columns()
  - scrollback history of 100 lines, see history_rows()
  - redraw_style() set to RATE_LIMITED, redraw_rate() set to 0.10 seconds
*/
Fl_Terminal::Fl_Terminal(int X,int Y,int W,int H,const char*L) : Fl_Group(X,Y,W,H,L) {
  // ...
  // End group
  end();
}
```
The docs of the c'tor should document its parameters and make clear that `end()` is called although `Fl_Terminal` is derived from `Fl_Group` which is not the "usual" behavior but IMHO sensible here. It should only be documented.

7. Last but not least: I believe that `Fl_Simple_Terminal` should be dropped entirely by the merge. There's no need to keep it.

8. Other than that I did some tests with `test/terminal` and I believe it's best to merge it now and let other users test it as well.

**_Thanks again for this awesome new widget!_**

erco77:
One nit I have with the widget: too many color methods. Not sure how to solve that one.
Fluid will need a special color menu for this;

- current text fg/bg
- default text fg/bg for text (used during a 'reset')
- cursor fg/bg
- selection fg/bg

There is probably a better and more consistent way to do the color api.. didn't like having so many methods, but I think it's probably unavoidable, given how text can be so many different colors, affecting how selection and cursors work, it's I don't think possible to automate contrast colors for those properly (so that one only needs a selection color (for the bg), and the fg color inside the selection is automatically chosen based on an algorithm). I know we improved our contrast color stuff, but I /think/ users of the widget would want full control of this.. not sure.

Also there currently isn't a textcolor(), which is what Fl_Simple_Terminal used. Since Fl_Terminal has full color control, I felt 'textcolor()' wasn't good enough, one needed to be clear there was a fg and bg color. I didn't think textcolor() and textbgcolor() was a good idea, but maybe it is for consistency with the rest of the fltk api?

erco77:
I'll check over your list, see if I can fix anything before release. Gotta have breakfast first..

erco77:
> 1. I don't grok the real usage of the CharStyle class.
Ah, OK, so this is an argument to the public methods `text_utf8(..style)` + `text_ascii(..style)` to allow per-char control of everything in style. The comment about never implementing "per-character font face/size (font face/size is global only)"  is regarding not supporting /different/ font face/sizes. You can set one font face/size and use it throughout, but can't make it "different", because assumptions are made about the font face/size being consistent throughout the terminal. The other attributes of CharStyle can be different per char, just not those two.
I can probably be clearer about that somehow..

> 2. TODO: Clean up what's public/protected

Yeah, I think I did do that pass, but it seems like it needs really careful analysis in ways that API designers are better at than myself. I'm always a little too liberal letting people access everything, and realizing years later I shouldn't have done that. I still find it hard to decide what to hide and what not. Like scrollbars seem important to expose, so that users can control the scroll position, or see where we're scrolled to, change its colors or such things.

I did I think go as far as I could doing that TODO item, but I think I left it in because I think it needed an extra check by others. So I do seek feedback on that.

> 3. From FL/Fl_Terminal.H: protected: int scrollbar_w(void) const;
> This should be documented/public and the name too cryptic.

Hmm, yes agreed the name is too cryptic, and I guess it could be public if a proper name were available. I think I hid it because I wasn't happy with the name, lol.

It should really be "size" because "width" doesn't work if you've got a horizontal scroller -- the size would be 'height' for that. And since the size is supposed to be the same for all scrollers, I needed to avoid the w/h concept. But size would compete with the public size api.

Perhaps the right name might be `scrollbar_actual_size()`, since that's what it's returning, and could make it public -- might that be a good solution?

> 5. A lot of NOTUSED comments

Ya, IIRC I left those in because they were used at one point, and after some optimizations they became unused, but I thought were good to leave in place in case they need to be resurrected in the future, as it might not be obvious how to do that -- didn't want to loose the work, esp. since I had to flatten out my commits..!

I'll give it a once over though to see if there's ones that are "obvious" and no record is needed.

> 6. (paraphrasing) Constructor documentation should mention end() is called.

OK

> 7. Drop Fl_Simple_Terminal

OK -- the reason it's there in the fork is to keep fluid working, because I wasn't sure how to replace it with Fl_Terminal; fluid's changed a bit lately, and it was too hard to figure out how to reimplement without matt's help. The colors API is one such issue. I might also need to provide an extra color method `textcolor()` as a convenience wrapper for `textfgcolor()` just to be compatible with old Fl_Simple_Terminal's way of setting the text color and apps that might use that method.

* * *

The Fl_Terminal API is vast, I know -- most of the code is probably comment documentation, which probably doesn't help readability. I've always not liked that about doxygen, mixing the docs in with the code makes the code hard to see. At the time we switched, I went so far as to make VI macros that "collapse" and "expand" to hide/show the doxygen comments while in the editor. I think I posted those VI macros way back when on the forum. Admittedly I don't use the macros I made much.. only when I /really/ need them. I've generally gotten used to doxygen comments, and the syntax coloring does a good job of making it easy to not see them.

erco77:
Regarding Fl_Simple_Terminal, I thought it best to leave it in there in 1.4.x (prerelease) so the pull could be done sooner rather than later and be as simple as possible, and then rework that later when me and matt had time to rework fluid, then we can remove it. It might also be good to have the fluid rework and Fl_Simple_Terminal removal from it and the code base be its own commit to fltk master, so the Fl_Terminal pull isn't too vast.

Albrecht-S:
> Regarding Fl_Simple_Terminal, I thought it best to leave it in there in 1.4.x (prerelease) so the pull could be done sooner rather than later and be as simple as possible, and then rework that later when me and matt had time to rework fluid, ...

I absolutely agree with this. Let's merge first and remove `Fl_Simple_Terminal` later if this is easier to achieve (and I think it is). As long as we don't release 1.4 `Fl_Simple_Terminal` doesn't really matter.

erco77:
If we're ready to merge, I'll do the above mentioned mods after.

To do the merge, I'll try to resync everything again, and if you want, you can pull the trigger. Currently it says there's no conflicts with the github green button above all our comments -- does that mean it's safe to use that?

Albrecht-S:
> It should really be "size" because "width" doesn't work if you've got a horizontal scroller

I don't think this is really an important point: the scrollbar is usually a **long object** and its "width" is the smaller of the two dimensions, no matter whether it "stands upright" (vertical) or it "lies on the ground" (horizontal). And since we do already have a precedent I think we **can** use `scrollbar_width()` as long as it is documented properly.

> Perhaps the right name might be scrollbar_actual_size(), since that's what it's returning, and could make it public -- might that be a good solution?

Well, if we can't agree about using `scrollbar_width()` then this would be my second choice. However, then we should also change the "other" widgets accordingly.

Albrecht-S:
>     TODO: Clean up what's public/protected
>  Yeah, I think I did do that pass, but it seems like it needs really careful analysis in ways that API designers are better at than myself. I'm always a little too liberal letting people access everything, and realizing years later I shouldn't have done that. I still find it hard to decide what to hide and what not.

If you're in doubt it's better not to expose a method. We can always add methods, even during a patch release (e.g. 1.4.3) because adding methods doesn't break the ABI.

> Like scrollbars seem important to expose, so that users can control the scroll position, or see where we're scrolled to, change its colors or such things.

It's good that you did not embed the scrollbar but allocated it with 'new'. This way someone could even replace the scrollbar with their own derived scrollbar class. Unfortunately this is not the case in our existing widgets like `Fl_Scroll` and we're even exposing the embedded widget. In such cases it would be much better to use methods that forward the call to the object or expose only a pointer to the object. IIRC this looked good in `Fl_Terminal`.

However I can't tell if `Fl_Terminal` exposes too many methods in its public API. I really didn't have the time to check it but we can certainly clean this up after the merge (before release) if need be.

Albrecht-S:
> If we're ready to merge, I'll do the above mentioned mods after.

I would prefer to have the small "cleanup" mods done before the merge. Doing it after the merge would add noise to our commit history for no useful reason.

> To do the merge, I'll try to resync everything again, and if you want, you can pull the trigger. Currently it says there's no conflicts with the github green button above all our comments -- does that mean it's safe to use that?

Basically I think it is (you should obviously use the "Squash and merge" option). If you only changed comments etc. in yet another commit (in `Fl_Terminal` header and source files) before the merge that would not affect the mergeability.

erco77:
OK -- I'll try to get to this tonight after work. Since the actor's strike just ended, the industry I work in just kicked into gear, so I'm suddenly even more busy than I was..

Albrecht-S:
About the color API: I have no opinion because I didn't study it in full detail. What you wrote looks fine but the question about all the colors persists. Sorry, I don't think I can help with this question.

erco77:
> 1. I don't grok the real usage of the CharStyle class.
OK, clarified a bit below the comment about different fonts/sizes in 06a6c45fe.
> 2. TODO: Clean up what's public/protected
Not sure what to do about this; not sure I want to remove it until users have had a go at things, and other devs have gotten to play with this widget.
> 3. scrollbar_w(void) should be documented/public and the name too cryptic.
I saw your comments regarding naming (width vs size), but honestly it'd be too inconsistent to use anything but 'size' -- the method does return `Fl::scrollbar_size()`, so if width would have been a better choice, I think that ship has sailed. So what I've changed is in b40e3c7bd.
> 4. copyright mods
Fixed in 2322641f3
> 5. A lot of NOTUSED comments
Fixed in 532d9a4fe. 
> 6. Constructor documentation should mention end() is called, and parameters doc'ed.
Fixed in 9bd2d798b.
> 7. Last but not least: I believe that Fl_Simple_Terminal should be dropped
I believe we agreed we're going to hold off on that until after the merge for the reasons described above.

OK, I think that's all the items. I'll hold for a final check before I do the squash+merge.

Albrecht-S:
> I'll hold for a final check before I do the squash+merge.

I reviewed the commits you pushed recently and it all looks fine to me. Good to go from my side.

Albrecht-S:
@erco77 Great, I just noticed the merge. Thank you very much.

@MatthiasWM Note, regarding Fluid integration: we need to **replace** `Fl_Simple_Terminal` with `Fl_Terminal` so we can **remove** `Fl_Simple_Terminal` entirely after this step is done. Thanks in advance.

erco77:
Yep, merged.

I'm assuming the gitlab docs will rebuild tonight, so that users can read the Fl_Terminal documentation, as they wouldn't really have a way to read it unless they have doxygen locally.

Albrecht-S:
I uploaded current docs as you suggested in fltk.general and the gitlab docs will update at ~6am UTC IIRC.

MatthiasWM:
I should have time today to pull Fl_Simple_Terminal form Fluid and Unites, if that's not gone already

MatthiasWM:
Do you have a list of parameters that you would like to have added to the FLUID UI?

MatthiasWM:
Ok, so dd8ac4d55755175ead0fa10a958a219d299948ae should replace incoming Fl_Simple_Terminals and otherwise only use Fl_Terminal. Shall I do all other occurrences of Fl_Terminal in `test`?

BTW: nice pitfall: since Fl_Terminal is now derived from Fl_Group, calling `Fl_Terminal::clear();` will now remove all children instead of clearing the text on screen and need to be replaced with appending the esc sequence to clear the screen. Maybe override Fl_Group::clear()? Or is it better to make Fl_Terminal::clear_screen() available? Glad I tested the shell terminal in FLUID ;-)

MatthiasWM:
Last comment: wow! You implemented an entire true full on terminal. And it's *fast*. Amazing.

Albrecht-S:
> ... since Fl_Terminal is now derived from Fl_Group, calling Fl_Terminal::clear(); will now remove all children instead of clearing the text on screen ... Maybe override Fl_Group::clear()?

Yes, we **_really should_** override `Fl_Group::clear()` in `Fl_Terminal`. `Fl_Simple_Terminal` had (and currently still has) its own [clear()](https://www.fltk.org/doc-1.4/classFl__Simple__Terminal.html#a7e48189c468920a9c3fdfcd121e0ecba) method [1] and so should `Fl_Terminal`. Removing all children wouldn't make sense if someone called `clear()` by accident or because they replaced **_(beta)_** `Fl_Simple_Terminal` with `Fl_Terminal` in their apps - same as you did now for Fluid.

This method would do something different than just adding an escape sequence and the escape sequence would (AFAICT) not work if '[Fl_Terminal::ansi()](https://www.fltk.org/doc-1.4/classFl__Simple__Terminal.html#a6fb412e11ecbb8d3524acced68cce627)' mode was disabled.

-----
[1] Note: the link to the docs will no longer work after `Fl_Simple_Terminal` has been removed. Docs state:
> Clears the terminal's screen and history.
> Cursor moves to top of window.

Albrecht-S:
> Shall I do all other occurrences of Fl_Terminal in test?

@MatthiasWM There is no Fl_**Simple_**Terminal in `test/` anymore (if that's what you wanted to ask?).

However, the following lines may need your attention:
```
fluid/README_fl.txt:343:   |    |    +-- Fl_Simple_Terminal
fluid/factory.cxx:856:  const char *alt_type_name() FL_OVERRIDE { return "Fl_Simple_Terminal"; }
```

@erco77 We have this comment and code in `Fl_Terminal.H`:
```C++
FL/Fl_Terminal.H:1008:  // Fl_Simple_Terminal API compatibility
FL/Fl_Terminal.H-1009-  int   history_lines(void) const;
FL/Fl_Terminal.H-1010-  void  history_lines(int val);
```

I think we can remove this "API compatibility" comment when we remove `Fl_Simple_Terminal`.

Are we ready to remove Fl_Simple_Terminal now?

Albrecht-S:
Hmm, the log of commit 69189c8f070c9e3347dae6cc835e5f510eca989f says:
> Fl_Terminal opens the X11 Display before shown, so we need to generate a proxy in batch mode. [This will no longer  work when we add more Fl_Terminal attributes to FLUID.]

@MatthiasWM @erco77 Without checking **why** this is so, I'm wondering: can we change `Fl_Terminal` not to open the display? ... Thinking about it ... Is this because Fl_Terminal does some font calculation for display width -> number of columns. If this is all we need, can this be postponed until the widget is shown() ... ?

MatthiasWM:
Yes, that's probably what it is. Right now I try to fix the color generation in FLUID, but I will try to look into that later tonight.

Albrecht-S:
Regarding my other question whether we're ready to remove `Fl_Simple_Terminal`:

Should we wait a few days or a week to give users of 1.4 (git) a chance to change their code?

@erco77, @MatthiasWM What do you think?

MatthiasWM:
@erco77 also, it seems that the boxtype for Fl_Terminal must be FL_DOWN_BOX, yet the default is FL_DOWN_FRAME and no background is drawn? I have not looked in the code yet, just trying to leave a stable version for the next few hours.

erco77:
Wow, you guys have been busy today.. I just woke up and haven't had my coffee yet, but:

> Do you have a list of parameters that you would like to have added to the FLUID UI?

I think the color() for background and textcolor() for text should work now (as of a commit I added yesterday).

Things unique to add I think would be:

- `selectionfgcolor() / selectionbgcolor()`
- `cursorfgcolor() / cursorbgcolor()`
- `history_rows(int)` and just call it 'history' as the label in fluid with a tooltip 'Number of lines of screen history for terminal'
- `ansi()` on/off

> Yes, we really should override Fl_Group::clear() in Fl_Terminal

Yes, I'll look into adding that; there should probably be a clear() and clear(Fl_Color).

It's kinda confusing since there's so many clear() methods in Fl_Terminal.H that are methods for the internal classes.
Would it be good programming practice to move all the internal class definitions into a separate file, e.g. `Fl_Terminal_Classes.H` and just `#include` them within the `class Fl_Terminal` definition, or some such?

Albrecht-S:
> I'd offer we put a #pragma warning (or whatever) in the Fl_Simple_Terminal.H file indicating this widget is being removed before 1.4.0 releases and to migrate to Fl_Terminal instead.

Although it's maybe not the perfect solution `FL_DEPRECATED` comes to mind. I works on all platforms that support it and doesn't do any harm otherwise.

Albrecht-S:
> It's my thinking that it would be good to wait a bit and not rush its full removal to allow people still using Fl_Simple_Terminal time to transition.

I agree. At some time in the near future we'll announce a feature freeze. That would be the time to finally remove Fl_Simple_Terminal.

erco77:
> [matt] it seems that the boxtype for Fl_Terminal must be FL_DOWN_BOX, yet the default is FL_DOWN_FRAME and no background is drawn?

Hmm, you should be able to change the box type and it should respond; the test/terminal program lets you interactively change the box() to test.

Can you supply replication for the problem you're seeing using that test program? The FLTK default up box decorations look weird on a terminal, which is why the default is a down frame instead. I assume fluid inquires from the widget what its defaults are, but perhaps there's a reason it doesn't (e.g. the headless command line support?) There were also some considerations for when different schemes are set, which affected the terminal drawing strangely (gradients), so I think I had to bypass that stuff so it would draw correctly, esp. since individual characters can have their own bg color.

erco77:
@MatthiasWM So long story short, it's best the default box() be Fl_Terminal's default of an FL_DOWN_FRAME. I'd though of making it FL_THIN_DOWN_FRAME, which I think looks nicer, but decided to keep as close to the FLTK default look as possible. (shrug)

MatthiasWM:
@erco77 Fl_Terminal() calls Fl_Terminal::CharStyle::CharStyle(void) which calls Fl_Terminal::CharStyle::update(void) which in turn calls fl_font(), and that one calls fl_open_display(). I know it kinda sucks, but can you defer opening a connection to the DISPLAY until you actually draw the terminal? Fl_Text_Display did that too, and it means that apps that construct but never show Fl_Terminal can not run headless.

This is related to the boxtype issue that I described earlier, but I managed to write a workaround. Textfont, color, and size are now selectable. Rest will follow later.

Albrecht-S:
@erco77 Please see also this bug report from "Tigercat" in `fltk.general`: https://groups.google.com/g/fltkgeneral/c/MocGlCiBOLU/m/YEQ2CIFeAwAJ

erco77:
Hmm, this might be tricky; Fl_Terminal automatically sets its character buffer size (the number of characters across and up and down) based on the widget's initial XYWH values, and uses the font size to do that.

So if the window width is 800 pixels and the characters are 10 in width, it makes an 80 character wide buffer (10*80=800). Ditto for the buffer height.

I guess I could just guess based on the fontsize, but guessing wrong makes for problems.

If this is just so fluid can load in -c mode, maybe there's some way to globally hint to the widget to prefer "guessing" to actual font use, just so fluid can do its thing.

Or is there some other way to get the font width without involving the display? (I'm guessing not)

erco77:
Perhaps a better way to do this is to provide some extra optional arguments past the usual XYWHL to the constructor that set the initial text buffer width/height, e.g.
```
// CTOR
Fl_Terminal(int X, int Y, int W, int H, const char*L=0,
            int display_cols=-1,   // force terminal's initial #cols for the text buffer display
            int display_rows=-1    // force terminal's initial #rows for the text buffer display
            );
```
This way fluid could set those extra values, preventing the widget from having to make the fl_font calculations.
I'm assuming I can retool things to do that.

We don't often have widget constructors that have extra arguments in the ctor, but I'm not sure I see an alternative, as the buffer needs to exist once the widget is created, so that people /can/ add text to it even before the widget is displayed.

This might give tigercat a solution as well.

Albrecht-S:
_Thinking ..._
Another option **_might be_**:
1. set an initial sensible buffer size when the widget is created, e.g. 80x100 and
2. handle the `FL_SHOW` event to actually do the font calculation and
3. (as you likely already do) adjust the buffer size when `resize()` is called.

Point 1. would have to flag that the size was initially set (not calculated) so points 2. and maybe 3. could see this flag and do the font calculation as appropriate.

erco77:
@Albrecht-S Early in development I had it setting the buffer size to a "sensible size", but quickly found that to cause problems; making assumptions about a default buffer size is not a good way to go, because different applications need different things. And changing (shrinking) the buffer size after adding text means possible data loss. This is why /resizing/ a window smaller doesn't also shrink the buffer size.

Pretty sure the two options we have to provide users:

- On widget creation, buffer size is automatic based on widget's w()/h(). This is the current default behavior which needs the font calculation in advance
- Let the user _optionally_ specify the buffer's row/col size during widget creation as optional ctor arguments, overriding the above automatic behavior, and avoiding (I think) any pre-show() font calculations

erco77:
If the optional ctor arguments are a bad way to go, the only other thing I can think of are having some global defaults for the row/col size.

The default for the globals could be row/col=(-1,-1), indicating the buffer size is automatically calculated (as it is now, based on font size), and the user (or fluid) could set the default in advance with e.g.
```
int main() {
  ..
  Fl_Terminal::default_display_cols(80);
  Fl_Terminal::default_display_rows(25);
  ..
  tty = new Fl_Terminal(0,0,0,0); // uses 80x25 for the buffer size
  ..
}
```
I'm thinking the ctor optional arguments would be better than this.. opinions?

MatthiasWM:
As you saw, I overrode resize(). FLUID instantiated Fl_Terminal with a size of 100x100 and then resizes it to whatever it needs. As you described, this can lose data. If I create a Terminal at 100x100, write something into it, and the resize it, and finally show it, my text is gone, but the cursor is still at the new position. It's ok behavior, but it does require FLUID to rebuild the content after that particular resize.

IMHO, the best thing would be no additional parameters or calls at all, but to wait with allocating the buffer until you actually need it. But I am very aware that this is no trivial. So for the time being, I have a second proxy class, which is fine for now.

erco77:
@MatthiasWM  Some clarification is needed:
- if you resize() the widget smaller, it /clips/ data, it doesn't loose it.. the data is still there. (if you resize the window larger it reveals the text that was clipped off at the smaller widget size)
- if you resize() the widget larger, the buffer will begin enlarging if the window gets larger than the buffer


So in other words, shrinking the /widget/ is not the same as shrinking the /buffer/. The former doesn't cause data loss, the latter does.

When you shrink the widget, the buffer remains whatever size it was.
When you enlarge the widget, the buffer may enlarge if it can fit more columns/rows.

This perhaps confusing behavior is exactly what xterm(1) does, and mutty(1) IIRC, and other terminal emulators I've seen that haven't yet evolved to support the fancy "word wrapping during resize" behavior.

MatthiasWM:
> > IMHO, the best thing would be no additional parameters or calls at all, but to wait with allocating the buffer until you actually need it.
> 
> Gotta disagree there; there's a lot of things the user can do after the widget is created but before the widget is shown that can play around with the buffer (adding text to it, changing colors/font sizes, presetting the scrollbar, etc) that really can't be deferred to draw time.

Understood. I meant 'needed' in the sense of 'accessing it', just not in the constructor. @Albrecht-S , how important is it to not open the display?  I know that we discussed this for Fl_Text_Display, but I don't remember the outcome.

erco77:
Yes, here the widget is sized smaller vertically, which moves the text up into the history (to preserve it), and should be re-revealed if you resize the widget larger again.
The floating cursor is an interesting artifact because there was no crlf at the end of the string, but the cursor can't move into the history buffer, so it's forced onto the display in the last horiz position it was in.

erco77:
Matt, can we have a quick facetime/skype or whatever? I have a few questions about fluid too that's related.

erco77:
It'll probably make more sense if you try this modified version of your program, which lets you resize the window, showing how it reveals the text when you make the window larger, and how it moves the text back into the history when you make it smaller.

```
#include <FL/Fl.H>
#include <FL/Fl_Window.H>
#include <FL/Fl_Terminal.H>

int main(int argc, char **argv) {
  Fl_Window *window = new Fl_Window(120+40, 120+40);
  Fl_Terminal *t = new Fl_Terminal(20, 20, 200, 200);
  t->append("Hello...");
  t->resize(20, 20, 120, 120);
  window->end();
  window->resizable(t);
  window->show(argc, argv);
  return Fl::run();
}
```

Albrecht-S:
> @Albrecht-S , how important is it to not open the display?

@MatthiasWM It becomes really important if you don't **have** a display, e.g. on a server where you want to use `fluid -c ...`. This would (AFAICT) typically be the case on CI machines GitHub/GitLab and likely in user development environments. Building FLTK is **only one** example...

I wonder if we would experience build fails in our GitHub CI environment. I just checked a build on my Linux system after ensuring that no `{X11|Wayland}` display can be found and for now it works (as was to be expected since you worked around it in fluid).

However we can't be sure that users do something else with `Fl_Terminal` such that it would try to open the display before `show()`. Setting FLTK options with `fltk-options` in commandline mode would be another example, BTW.. That could be done in some automated install script. This is only an example why users may do things without a display, I'm aware that `fltk-options` doesn't include an `Fl_Terminal` widget.

MatthiasWM:
I sent you an email with the Zoom data.

Albrecht-S:
@erco77 That said, I would prefer a solution that
1. does not add extra arguments to the constructor and
2. does not open the display at widget construction time.

I mean, the default should not be to calculate font sizes in the constructor. Everything that can be done by the user before `show()` that might need a valid / sensible buffer size could call a method that initializes it when really needed, for instance:
```
  if (!valid()) initialize_buffer();
  // do whatever is to be done after that (in Fl_Terminal code, not user code)
```
Such a construction is IIRC recommended or even required in user code of `Fl_Gl_Window::draw()`.

If we set a pretty small size like 80x24 (VT100 size) by default data loss would be unlikely and if we documented this fact then it's the user's responsibility to set an appropriate (non-default) buffer size if need be. In real cases many users may want to set their own minimal buffer size and I'd say that this should be done shortly (or immediately) after construction before any data is written to the widget.

Would this work?

erco77:
OK, so I caught up with Matt, and:
- He currently has a workaround for the font/display issue using an Fl_Group in place of Fl_Terminal. It's not optimal because he has to hard code Fl_Terminal's defaults, but hopefully I can provide him something in the near future based on what we come up with here
- I talked to him about the deferred buffer thing, and we agreed there's no way to defer it; the buffer really has to exist after the ctor is called.

One can use the display_columns() and display_rows() to change the buffer size, but the default has to exist. It might be possible to use some default (like 80x25 or whatever), but it makes the average use of the widget difficult, because it will crop unexpectedly. The default behavior really should be based on the widget size.

The only way out it /seems/ is to specify the ctor with optional arguments, or perhaps a second constructor like we have for Fl_Box, e.g. (pseudo code cause I'm in a rush_:
```
   Fl_Terminal(x,y,w,h,l=null);  // normal ctor
   Fl_Terminal(x,y,w,h,l,row,col); // extra ctor with fixed arguments (no defaults)
```
This avoids having extra optional arguments in the ctor, while defining a fixed ctor type.
Then fluid could use that second ctor, and thus avoiding the display thing that is very unique to only fluid. (I don't think user programs would /ever/ have to worry about that issue)

Albrecht-S:
@erco77 I understand your points (no need to communicate further). I'm fine with whatever you decided together and if we have to go the c'tor route I'd vote for the `extra ctor with fixed arguments (no defaults)`.

> Then fluid could use that second ctor, and thus avoiding the display thing that is very unique to only fluid. (I don't think user programs would /ever/ have to worry about that issue)

OK, then let's do it and see if someone complains. Hopefully not.

The bug report mentioned above: I saw that you replied and you'll try to fix it by setting sensible values if the widget is too small so we can avoid zero division or whatever Tigercat's exact issue was. This would be useful anyway, now that we know that it can happen. Users sometime use widgets in an unexpected way. ;-)

erco77:
OK, sounds good.

Keep in mind when fluid uses this new ctor, it's /only/ for fluid's own internal use as a "widget database", the new ctor wouldn't end up in fluid's user generated code, it'd just be to avoid the headless display issue for `fluid -c`.

I'll work on that next after I push out the public clear() method.

erco77:
> Fl_Terminal() calls Fl_Terminal::CharStyle::CharStyle(void) which calls Fl_Terminal::CharStyle::update(void) which in turn calls fl_font(), and that one calls fl_open_display(). I know it kinda sucks, but can you defer opening a connection to the DISPLAY until you actually draw the terminal?

I've opened issue #837 to explore this.

_You're right it is a pain, lol. I'm sure this has been explored before, but I just have to turn the tables: Can't fluid defer instancing widget(s) until it has to bring them up in the UI? I would like to think the .fl file alone is all the source code generator needs to generate code. I'm assuming the .fl file only has widget values that are NOT the default from the widgets (since the widgets know their own defaults), and so only non-default values would end up in the generated .cxx/.h code._

Regardless, I'll explore this.. it might be possible though certainly a bit painful.

MatthiasWM:
> _I would like to think the .fl file alone is all the source code generator needs to generate code. I'm assuming the .fl file only has widget values that are NOT the default (the widgets know their own defaults), and so only non-default values would end up in the .fl and in the generated .cxx/.h code._

Yes, it can and it currently does. FLUID works by instantiating every supported widget. It does so to get the default values for every type (box types, colors, etc.) and to store widget specific parameters (text size for text based widgets etc.). I am currently instantiating a custom widget instead that stores some of the Fl_Terminal parameters, and that works fine, but it will run out of sync if we should ever change default parameters in FLTK and not in FLUID. Not a huge deal, but something we will surely forget.

I still think it is unexpected for users that simply instantiating a widget and not showing it already opens a connection to the display, but I am sure you discussed that in the other thread (#837).

erco77:
OK, so I added the fix for issue #837 in 6252131 that provides that extra constructor that prevents opening the display by providing the row/col/hist values, so it doesn't have to be calculated via fl_width/fl_height/etc.

So with that you should be able to replace the fluid custom stand-in widget with an actual Fl_Terminal instance using the new ctor.

I closed issue #837, but feel free to respond there if there's trouble.

erco77:
> I still think it is unexpected for users that simply instantiating a widget and not showing it already opens a connection to the display..

Well, I guess it's unexpected that doing anything with fonts accesses the display too, which is the reason for all this, but that's the way X11 works, and it makes sense in the X11 universe where fonts might be resolved on a remote DISPLAY.

I'm not sure if that's the case with Wayland, or for that matter any of the other platforms, so it may become a non-issue.

I'm also not sure users should expect instancing widgets wouldn't access the display. It seems to cause confusion that they don't, actually -- people are often surprised by the "delay" between when a widget is instanced, and when it is 'mapped' to the display.

Issue #828, opened 2023-11-07
RFE: Add "combine mask" to Fl_Image_Surface for masking effects similar to using Fl_Window::shape()


pdhahn:
**Is your feature request related to a problem? Please describe.**
I would like to specify a mask for drawing effects on surface images. This would be a "combining" mask that allows drawing over a "shape" defined for the drawing surface with effects similar to what can be achieved in the case of windows using the Fl_Window::shape() method. This means controlled transparency or (effectively) full clipping.

**I am using FLTK**
 - Version 1.4.0

**Describe the solution you'd like**
The mask would be an instance of Fl_RGB_Image. The new public methods (accessors) in Fl_Image_Surface would be
```c++
const Fl_RGB_Image* mask(void);  // get mask
int mask(const Fl_RGB_Image* mask);  // set mask; makes internal copy of 'mask' (?)
```

**Describe alternatives you've considered**
Employ Fl_Window::shape() and then draw the window into the surface image using the inherited method `Fl_Widget_Surface::draw (Fl_Widget *widget, int delta_x, int delta_y)`.

Assuming that would work, it still would be more intuitive as well as convenient if the new mask feature were implemented in Fl_Image_Surface directly.

**Additional context**
It should be possible to change the mask more that once between drawing actions while preserving results in a cumulative fashion. For example, data already drawn into the surface image is preserved wherever the new mask clips new drawing actions that would otherwise overwrite that data.

pdhahn:
Thanks for correcting the cross-wired class name. When typing, I sometimes fall victim to this kind of transposition problem without even noticing.

ManoloFLTK:
I've got it working quick and dirty for the macOS and hybrid Wayland/X11 (both draw with Cairo) platforms. For these, it's enough to re-use the window shaping system of these platforms.

But, for the pure X11 platform, the window shaping system uses the Shape extension of X11 which is apparently devoted to windows, not to pixmaps.
For the Windows platform, the window shaping system uses 
[SetWindowRgn()](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowrgn) that's also specific of windows.

How to support this feature under X11 and Windows?

pdhahn:
For X11, maybe utilize [XShapeCombineMask](http://www.massapi.com/method/xs/XShapeCombineMask.html)()?

pdhahn:
Offscreen _pixmap_ or _offscreen_ window (with window manage intereference disabled)?

[...]

PR #851, opened 2023-11-29
New member function Fl_Image_Surface::mask(Fl_RGB_Image*)

ManoloFLTK:
This is a proposal to solve issue #828. It consists in a new member function added to class `Fl_Image_Surface` documented as follows:
image

The test app `device` is modified to exercize the new member function: see its new `Fl_Image_Surface::mask()` radio button which gives examples of construction of masked images.


Albrecht-S:
@ManoloFLTK You asked for my review and I read the documentation provided by you above. I believe it's OK to have this new method but I'm too busy and not familiar with this `Fl_Image_Surface` stuff, hence I can't help with a real review. Sorry.

MatthiasWM (approved):
I don't have a specific use, but I can see how this helps for certain drawing tasks. I tested the code on an M1 MacBook Pro and it works as expected. The API seems fine to me. Thank you for the addition.

pdhahn:
Thanks. At one point, I had planned to try to implement my own version and submit as a pull request (per #828), but I had to divert to higher priorities (current). I made progress but didn't get to a finish point before diverting. Like a lot of you probably, I have too many irons in the fire.

I look forward to seeing your implementation code when available for download.  After that, if I might still be able to contribute something of related value, time permitting, I will submit the pull request or communicate my work via **fltk.general** for consideration by the FLTK developer team.

ManoloFLTK:
@erco77 Many thanks for your input on this topic.

> In the description, I wasn't sure what this meant exactly:
> ```
> On some platforms, other colors produce intermediate masking scores.
> ```
> Does that mean non-white, i.e. gray scales, might produce a soft matte? Similar to how an alpha channel that varies between 0 - 255 would cause a partial composite?

On some platforms, drawing is antialiased so the mask image constructed by drawing with white color on a solid black background contains either black or white at the vast majority of pixels, but other colors at some border pixels. This statement means these pixels get masked with intermediate intensity, yes, similarly to a partial composite. This is a very small detail. I would appreciate if you could propose a clearer, while short, explanation of this detail.

erco77:
For this bit:

> On some platforms, other colors produce intermediate masking scores.

..I might say instead, assuming this is correct:

        On some platforms, shades of gray in the mask() image will have  the effect of transparency;
        values closer to white will be more transparent, closer to black will be less transparent.

> Overall, the image obtained calling Fl_Image_Surface::image() contains all drawings made
> until the image() call in the white mask areas, and only drawings made before the mask() call in black areas.

One question: is it the case that once a mask() is defined, does it only affect what's drawn _from then on_, and not what's been drawn before the mask() was created? And if so, then it sounds like the mask is applied only during new drawing operations after the mask() has been created.

I imagine the other way a mask() might work is that it doesn't affect drawing at all, but is applied at the time of displaying the image.

I did clone your fork, checked out the masked-image branch, built the code and tested the test/device program, checked the mask() box, hit GO and I see this image:

image

It's hard to tell from that image if the mask, which I take it is the "FLTK" text and circular cutout, is being applied during drawing, or is just masking out the image surface.

Another question: I didn't dig into the code, but I'm guessing the FLTK and circular cutout were created on the fly by creating an image() that was all black, and then white text and a thick white circle were "drawn" into it to create the mask. And then a separate image() surface is created, perhaps a yellow fill, and the mask() image applied to it, and then the test program's widgets were then drawn over the mask(), causing it to only draw through the white areas of the mask(), leaving the yellow background untouched in the parts of the mask() that were black. **Is that correct?**

If that's so, then it might be good to include a small sequence of images showing how everything is applied, e.g. something like:

image

erco77:
Yes, I agree a small commented code example would probably be best, and a more minimal diagram than mine, just to show visually what the mask() and final result would look like, as someone reading the docs might not be able to run the demo at the time they're reading it.

PR #898, opened 2024-01-29
Fix fluid target added twice when crosscompiling

LilyWangLL:
When `CMAKE_CROSSCOMPILING` is true, `fluid` will be added twice by `add_executable,` it will cause the following error:
```
  _add_executable cannot create imported target "fluid" because another
  target with the same name already exists.
```
The following codes are from `FLTK-Targets.cmake` which is generated by `CMake`:
Line 58:
```
# Create imported target fluid
add_executable(fluid IMPORTED)
```

### DIFF
--- a/CMake/FLTKConfig.cmake.in
+++ b/CMake/FLTKConfig.cmake.in
@@ -38,7 +38,9 @@ if (CMAKE_CROSSCOMPILING)
     PATHS ENV PATH
     NO_CMAKE_FIND_ROOT_PATH
   )
-  add_executable(fluid IMPORTED)
+  if(NOT TARGET fluid)
+    add_executable(fluid IMPORTED)
+  endif()
   set_target_properties(fluid
     PROPERTIES IMPORTED_LOCATION ${FLUID_PATH}
   )


Albrecht-S:
Hi @LilyWangLL, thank you very much for your bug report and this PR, please see comments below.

1. I confirm that you found a bug
2. This will be fixed in FLTK 1.4.0
3. I'm not sure if this can (will) be fixed in FLTK 1.3 (maybe we'll release 1.3.10 soon)
4. Unfortunately your proposed solution (and this PR) is not correct (see below, why)

**_Explanation:_** When cross-compiling there are two distinct fluid targets involved:
- (a) the host fluid application that's used for building test and example programs, which this particular part of the code wants to "import"
- (b) the fluid target that's being built for the target platform which is the target that's already defined.

**_History:_** In previous versions we didn't build fluid when cross-compiling and the "imported" fluid program was used for building the test and example programs, hence there was no conflict. Later we fixed this bug (now we're building fluid even if cross-compiling) but this introduced the new bug you are seeing.

**_Solution:_** We need to differentiate better between the host and the target fluid program - and this is what I'm going to do.

That's why I set the label `invalid` but this doesn't mean that your **_report_** is invalid, it means only that **_the proposed fix is not correct_**.

However, the real fix may take some time because I'm just in the process of the transition to "modern CMake" in our 1.4 development branch. **_Please leave this PR open, I'll close it when the bug is fixed._**

Albrecht-S:
Note: I looked at the [vcpkg PR](https://github.com/microsoft/vcpkg/pull/36442) and noticed that this PR concerns the build of FLTK 1.3.9. I don't know anything about vcpkg but IMHO this is also the wrong solution for this build (it hides the warning but doesn't solve the bug).

I'll let you know about my progress ASAP ...

LilyWangLL:
> Note: I looked at the [vcpkg PR](https://github.com/microsoft/vcpkg/pull/36442) and noticed that this PR concerns the build of FLTK 1.3.9. I don't know anything about vcpkg but IMHO this is also the wrong solution for this build (it hides the warning but doesn't solve the bug).
> 
> I'll let you know about my progress ASAP ...

Thanks for your confirm, I will keep this PR and my vcpkg PR open until the correct solution.

dg0yt:
It correct and necessary to not create the target again when it already exists.
What is missing from the PR is also not modifying the target's properties when it already exists.

The minimal test case for such problems is a repeated `find_package`:
~~~
find_package(FLTK CONFIG REQUIRED)
find_package(FLTK CONFIG REQUIRED)
~~~
Looks artificial, but this is the net effect of when pulled in multiple times for transitive usage requirements.

Albrecht-S:
@dg0yt Thanks for your comment, you are correct about not trying to add a target twice and not to change its properties. In general. I also used your minimal test case for my internal tests.

However, as I wrote before, this PR is not correct regarding the `fluid` target because it's way more complicated and we need to distinguish cross compiling and normal builds. When cross-compiling we need two `fluid` targets (with different names, obviously): one from the host for building the test/demo programs and one that we actually build.

I'll consider this in my fix which will definitely be different, but I'm leaving this PR open as a reminder until I can find the time to commit a proper fix.

Albrecht-S:
FTR, final note (hopefully): commit 63d7c71e1a926f487f22aa26042a2582624b3b17 (Git `branch-1.3`) fixes the issue for FLTK 1.3 as well. This is a partial backport of changes done in 1.4 (Git `master`).

This fix will be included in FLTK 1.3.10, the next and supposedly last release in the 1.3.x series.

@LilyWangLL Please confirm that this fixes the issue for you.

LilyWangLL:
> FTR, final note (hopefully): commit [63d7c71](https://github.com/fltk/fltk/commit/63d7c71e1a926f487f22aa26042a2582624b3b17) (Git `branch-1.3`) fixes the issue for FLTK 1.3 as well. This is a partial backport of changes done in 1.4 (Git `master`).
> 
> This fix will be included in FLTK 1.3.10, the next and supposedly last release in the 1.3.x series.
> 
> @LilyWangLL Please confirm that this fixes the issue for you. This PR will be closed when confirmed or in about 7-10 days, whichever comes first. Feel free to close it yourself. Thanks.

Thanks for your fix and reply, I update my patch on vcpkg with your new change.

PR #905, opened 2024-02-13
make FLTK_SCALING_FACTOR override the auto-detected value

mikeandmore:
Currently, FLTK_SCALING_FACTOR doesn't override the existing scaling factor detected from xrdb. This patch fixes that. If the environment variable is detected, it'll override the scaling factor.

### DIFF
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -1254,10 +1254,10 @@ int fl_handle(const XEvent& thisevent)
     d->init_workarea();
 #if USE_XFT
     float old_scales[Fl::screen_count()];
-    for (auto i = 0; i < Fl::screen_count(); i++) old_scales[i] = d->scale(i);
+    for (int i = 0; i < Fl::screen_count(); i++) old_scales[i] = d->scale(i);
     d->use_startup_scale_factor();
-    for (auto i = 0; i < Fl::screen_count(); i++) {
-      auto new_scale = d->scale(i);
+    for (int i = 0; i < Fl::screen_count(); i++) {
+      float new_scale = d->scale(i);
       if (old_scales[i] - new_scale < 0.01) continue;
       d->scale(i, old_scales[i]);
       d->rescale_all_windows_from_screen(i, new_scale);
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -38,6 +38,7 @@
 #  include <FL/filename.H>
 #  include <stdio.h>
 #  include <stdlib.h>
+#  include <cmath>
 #  include "flstring.h"
 #  include "drivers/X11/Fl_X11_Screen_Driver.H"
 #  include "drivers/X11/Fl_X11_Window_Driver.H"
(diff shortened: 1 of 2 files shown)


Albrecht-S:
@mikeandmore Thanks for the PR, but I don't think it's acceptable in the current form:

1. It looks like you removed the existing xrdb query altogether. Why did you do that, is there a reason why it is not needed (if the env var is not set)?
2. The new code is not compliant with FLTK 1.4.x and earlier coding standard. We allow to use C++98 and 'auto' has been added later. I didn't try to patch + build yet though, maybe there's more...

Of course patches like yours are always appreciated if they are correct and their form is acceptable. In the next minor FLTK version (1.5.0) we'll allow C++11 (or maybe higher) coding standards.

@ManoloFLTK It looks like this PR is for you, would you like to assign it to you?

mikeandmore:
Yes, I can explain. :)

1. `desktop_scale_factor()` (and `use_startup_scale_factor()`) uses `XGetDefault()` to query xrdb already. We don't need to spawn an external process for that.
2. I can fix that. I wasn't aware that supporting C++98 was a goal.

mikeandmore:
Removed `auto`

src/Fl_x.cxx
```diff
@@ -1276,7 +1253,15 @@ int fl_handle(const XEvent& thisevent)
     Fl_X11_Screen_Driver *d = (Fl_X11_Screen_Driver*)Fl::screen_driver();
     d->init_workarea();
 #if USE_XFT
-    after_display_rescale(&(d->current_xft_dpi));
+    float old_scales[Fl::screen_count()];
+    for (int i = 0; i < Fl::screen_count(); i++) old_scales[i] = d->scale(i);
+    d->use_startup_scale_factor();
+    for (int i = 0; i < Fl::screen_count(); i++) {
+      float new_scale = d->scale(i);
+      if (old_scales[i] - new_scale < 0.01) continue;
```
Albrecht-S, line 1261: Hmm, shouldn't this compare the absolute value of the difference? The removed code seemed to use `fabs()` in a similar case...
Sorry, if this is wrong, I'm just having a quick look at the code.

mikeandmore: good catch. fixed

ManoloFLTK:
I am not favorable to apply this PR for these reasons:

1) Function `after_display_rescale() `remains necessary. It does not run when the app first opens the X11 display. It runs while the app runs, if one screen (there can be several) is resized, or has its DPI changed, possibly also when the relative locations of diverse screens change. 

2) The `popen()` call therein remains necessary because function `XGetDefault()` always returns the value valid at application startup, missing all future changes.

3) Environment variable FLTK_SCALING_FACTOR is on purpose processed downstream from the reading of the Xft.dpi resource. That's in analogy to how Ctrl/+/-/0/ keystrokes work: they operate as multipliers of the initial scaling value which was set dividing the value of Xft.dpi by 96.
FLTK_SCALING_FACTOR has the same multiplier effect. Notice that FLTK_SCALING_FACTOR operates identically for all platforms, X11 or not.

@mikeandmore I am willing to consider arguments whether FLTK_SCALING_FACTOR should operate differently.

src/Fl_x.cxx
```diff
@@ -38,6 +38,7 @@
 #  include <FL/filename.H>
 #  include <stdio.h>
 #  include <stdlib.h>
+#  include <cmath>
```
Albrecht-S, line 41: Please use "math.h" instead. Sorry for "nitpicking", but that's **old** FLTK style.

PR #906, opened 2024-02-13
Fix Terminal character position and interactive selection

MatthiasWM:
This patch fixes the vertical character position, so that characters are rendered starting in the top left corner. The current code uses the character baseline and character Y position interchangeably, which offsets the selection box vertically. This can be tested by selecting text by clicking into the bottom half of a character, which selects the character one row too low.

This patch fixes the horizontal selection offset which did not take the frame width of the widget into account.

This patch adds code that expands character selection when the mouse crosses the middle of a character horizontally., not the left or right edge. This mimics Fl_Input and other common GUIs.

This patch adds code to select a word or a line on double or tripple click respectively. It's not perfect as it does not handle dragging after multi-clicks very well, but it's better than before. I may add the missing code later in another patch.

This patch fixes the "overshoot" in calculating the vertical number of rows depending on the window height because the vertical position is now more precise.

### DIFF
--- a/FL/Fl_Terminal.H
+++ b/FL/Fl_Terminal.H
@@ -658,30 +658,34 @@ private:
   // Class to manage mouse selection
   //
   class FL_EXPORT Selection {
+    Fl_Terminal *terminal_;
     int srow_, scol_, erow_, ecol_;     // selection start/end. NOTE: start *might* be > end
     int push_row_, push_col_;           // global row/col for last FL_PUSH
+    bool push_char_right_;
     Fl_Color selectionbgcolor_;
     Fl_Color selectionfgcolor_;
     int state_ ;                        // 0=none, 1=started, 2=extended, 3=done
     bool is_selection_;                 // false: no selection
   public:
-    Selection(void);
+    Selection(Fl_Terminal *terminal);
     int srow(void) const { return srow_; }
     int scol(void) const { return scol_; }
     int erow(void) const { return erow_; }
     int ecol(void) const { return ecol_; }
-    void push_clear() { push_row_ = push_col_ = -1; }
-    void push_rowcol(int row,int col) { push_row_ = row; push_col_ = col; }
-    void start_push() { start(push_row_, push_col_); }
-    bool dragged_off(int row,int col) { return (push_row_ != row) || (push_col_ != col); }
+    void push_clear() { push_row_ = push_col_ = -1; push_char_right_ = false; }
+    void push_rowcol(int row,int col,bool char_right) {
+      push_row_ = row; push_col_ = col; push_char_right_ = char_right; }
+    void start_push() { start(push_row_, push_col_, push_char_right_); }
+    bool dragged_off(int row,int col,bool char_right) {
+      return (push_row_ != row) || (push_col_+push_char_right_ != col+char_right); }
     void selectionfgcolor(Fl_Color val) { selectionfgcolor_ = val; }
     void selectionbgcolor(Fl_Color val) { selectionbgcolor_ = val; }
     Fl_Color selectionfgcolor(void) const { return selectionfgcolor_; }
     Fl_Color selectionbgcolor(void) const { return selectionbgcolor_; }
     bool is_selection(void) const { return is_selection_; }
     bool get_selection(int &srow,int &scol,int &erow,int &ecol) const; // guarantees return (start < end)
-    bool start(int row, int col);
-    bool extend(int row, int col);
+    bool start(int row, int col, bool char_right);
+    bool extend(int row, int col, bool char_right);
     void end(void);
     void select(int srow, int scol, int erow, int ecol);
     bool clear(void);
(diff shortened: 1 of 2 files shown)


erco77:
OK, I'll take a look..

erco77:
Cool that you were able to get the font drawing more accurate!

Some small things while testing on Linux (Ubuntu 20.04) and comparing old to new:

- Seems like the underline is too low now, making it seem more like overline for the line below:

image

- Also seems like the strikeout is too low now; I was kinda happy how strikeout went through the center line of uppercase text, e.g. the center line of "B", but now it's even below the center of a lowercase 'o'.
-

MatthiasWM:
Yeah, I moved them both, because on macOS, they look much more like on the left *after* the change. At some point, we have to make our font attributes similar across platforms as well... . Anyway, I will revert to the old code.

erco77:
I see the new double-click to highlight a word; nice.
Is it a future plan to select words if you drag after double-clicking a word? (Whenever a new feature is added, someone's always gonna notice the micropscopic missing sub-features, lol)

erco77:
Drawing fonts relative to the new 'baseline' variable seems clearer and self-documents well; I like that change.

MatthiasWM:
I kept the name change. I just reverted the calculation to the original values a minute ago.

erco77:
And glad that you got rid of the "+0.5 overshoot" stuff; that was definitely a "make it work for now, figure it out later" hack, lol

MatthiasWM:
> * I notice a new Fl_Terminal* member was added to the Selection class, and as an argument to the constructor. Is that a leftover from debugging? It didn't seem to be used.

No. Unfortunately it is the only way to get the number of columns in the buffer, so the selected column value won't overflow in `Fl_Terminal::Selection::extend`, see `int maxCol = terminal_->ring_cols()-1;` in line 210.

erco77:
Glad you fixed the shift-click selection; there was something really wrong with how I was doing that. Making a highlight and then holding down shift and randomly clicking around above and below the start of selection did not extend the selection correctly, and I think it does now.

erco77:
> No. Unfortunately it is the only way to get the number of columns in the buffer, so the selected column value won't overflow in Fl_Terminal::Selection::extend, see int maxCol = terminal_->ring_cols()-1; in line 210.

Oh, I was searching for "terminal->" to see if you had used that anywhere, but I didn't include the trailing underbar in my search, I see, OK.

MatthiasWM:
I really like the new Fl_Terminal and I use it as a builtin debug in a project. I even added minimal code to enter command line text. Not production ready by any means, but good enough for a non-utf8 "gdb" style input. Code follows just for fun.

```
class TFLTerminal : public Fl_Terminal
{
	std::vector<std::string> history;
	std::vector<std::string> future;
	std::string cmd;

public:
	TFLTerminal(int x, int y, int w, int h, const char* l = nullptr) :
			Fl_Terminal(x, y, w, h, l) { }
	int
	handle(int e) FL_OVERRIDE
	{
		char c;
		switch (e)
		{
			case FL_KEYBOARD:
				c = Fl::event_text()[0];
				if (c < 127 && c > 31 && isprint(c) && !Fl::event_command())
				{
					cmd += Fl::event_text();
					this->printf("%s", Fl::event_text());
				} else if (Fl::event_key() == FL_Enter)
				{
					do_callback();
					history.push_back(cmd);
					future.clear();
					cmd.clear();
					this->printf("\n");
				} else if (Fl::event_key() == FL_Escape)
				{
					cursor_left((int) cmd.length());
					this->printf("\033[K");
					cmd.clear();
				} else if (Fl::event_key() == FL_Up)
				{
					if (!history.empty())
					{
						if (!cmd.empty())
							future.push_back(cmd);
						cursor_left((int) cmd.length());
						this->printf("\033[K");
						cmd = history.back();
						history.pop_back();
						this->printf("%s", cmd.c_str());
					}
				} else if (Fl::event_key() == FL_Down)
				{
					if (!cmd.empty())
						history.push_back(cmd);
					cursor_left((int) cmd.length());
					this->printf("\033[K");
					if (future.empty())
					{
						cmd.clear();
					} else
					{
						cmd = future.back();
						future.pop_back();
					}
					this->printf("%s", cmd.c_str());
				} else if (Fl::event_key() == FL_BackSpace)
				{
					if (!cmd.empty())
					{
						cmd.pop_back();
						this->printf("\033[1D\033[K");
					}
				} else if (Fl::event_key() == 'v' && Fl::event_command())
				{
					Fl::paste(*this);
				} else
				{
					return Fl_Terminal::handle(e);
				}
				return 1;
			case FL_PASTE:
				cmd += Fl::event_text();
				this->printf("%s", Fl::event_text());
				return 1;
		}
		return Fl_Terminal::handle(e);
	}
	const char*
	value()
	{
		return cmd.c_str();
	}
};
```

erco77:
Oh, you added triple-click line selection too.. cool!
I think dragging while in a triple-click selection would extend the selection in line mode _(bet you're working on that :^D)_

This all looks good to me; I'll try diffing the old and new code later tonight, as right now I'm just looking at your commit diffs, when I should really look at the overall between your branch HEAD and the master HEAD to get a proper idea of what's changing.

MatthiasWM:
My branch is a very current copy of master. The complete diff is in the GitHub web interface under the "Files Changed" tab:
https://github.com/fltk/fltk/pull/906/files

MatthiasWM:
@erco77 Greg, would you like me to push the changes, or do you need some more time? I'll be happy to fix whatever you find.

erco77:
Before doing a merge, can you check the two Pending code comments and weigh in on those?

I don't know how to reference them directly, but if you view this issue on github and scroll down, you should be able to see two pending code comments awaiting review:

image

MatthiasWM:
Ah, ok, the typos is fixed.

The second bug was caused by seeing `xy_to_glob_rowcol()` returning an `int`, and not reading the line where you convert it to a `bool`. Thanks for finding this.

erco77:
I'd say go for it.
If there's anything to fix, we can do it as an update.

PR #910, opened 2024-02-16
using xlib to query the new DPI

mikeandmore:
Always open up a new display when current DPI isn't available.

In the main loop, we reuse `desktop_scale_factor()` function.

### DIFF
--- a/src/drivers/X11/Fl_X11_Screen_Driver.cxx
+++ b/src/drivers/X11/Fl_X11_Screen_Driver.cxx
@@ -429,7 +429,7 @@ extern void fl_fix_focus(); // in Fl.cxx
 void Fl_X11_Screen_Driver::grab(Fl_Window* win)
 {
   const char *p;
-  static bool using_kde = 
+  static bool using_kde =
     ( p = getenv("XDG_CURRENT_DESKTOP") , (p && (strcmp(p, "KDE") == 0)) );
   Fl_Window *fullscreen_win = NULL;
   for (Fl_Window *W = Fl::first_window(); W; W = Fl::next_window(W)) {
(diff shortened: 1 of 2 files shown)


src/drivers/X11/Fl_X11_Screen_Driver.cxx
```diff
@@ -1369,14 +1369,17 @@ static void* value_of_key_in_schema(const char **known, const char *schema, cons
 void Fl_X11_Screen_Driver::desktop_scale_factor()
 {
   if (this->current_xft_dpi == 0.) { // Try getting the Xft.dpi resource value
-    char *s = XGetDefault(fl_display, "Xft", "dpi");
+    // Always get the up to date value
+    Display *new_dpy = XOpenDisplay(0);
```
Albrecht-S, line 1373: I wonder if this would always open the "right" display. FLTK has commandline arguments that can be used to open any possible display, not only the "default" display (the one set by the DISPLAY env var). The above code might therefore open another display with different settings than the one that's currently in use.  Theoretically. Maybe.

Albrecht-S:
@mikeandmore Thanks for the new PR. I'm leaving this to @ManoloFLTK but wanted to add my 2 ct just that this aspect is not forgotten.

ManoloFLTK:
For the reason explained by Albrecht, the proposal to call XOpenDisplay(0) is not adequate.

src/drivers/X11/Fl_X11_Screen_Driver.cxx
mikeandmore: Fixed. Now it's using `XDisplayString()` to get the display name from `fl_display`

mikeandmore:
> For the reason explained by Albrecht, the proposal to call XOpenDisplay(0) is not adequate.

This can be easily fixed. Can you reopen the PR?

mikeandmore:
Hmm... It looks like since you close the PR, this PR will not pickup my further fix any more. I'll open a new PR

PR #913, opened 2024-02-17
using xlib to query the new DPI (old #910)

mikeandmore:
Always open up a new display when current DPI isn't available.

In the main loop, we reuse desktop_scale_factor() function.

This PR now makes sure we open the same display as `fl_display`.

### DIFF
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -1215,24 +1215,15 @@ static void react_to_screen_reconfiguration() {
 
 #if USE_XFT
 static void after_display_rescale(float *p_current_xft_dpi) {
-  FILE *pipe = popen("xrdb -query", "r");
-  if (!pipe) return;
-  char line[100];
-  while (fgets(line, sizeof(line), pipe) != NULL) {
-    if (memcmp(line, "Xft.dpi:", 8)) continue;
-    float dpi;
-    if (sscanf(line+8, "%f", &dpi) == 1) {
-      //fprintf(stderr," previous=%g dpi=%g \n", *p_current_xft_dpi, dpi);
-      if (fabs(dpi - *p_current_xft_dpi) > 0.01) {
-        *p_current_xft_dpi = dpi;
-        float f = dpi/96.;
-        for (int i = 0; i < Fl::screen_count(); i++)
-          Fl::screen_driver()->rescale_all_windows_from_screen(i, f);
-      }
-    }
-    break;
+  float old_dpi = *p_current_xft_dpi;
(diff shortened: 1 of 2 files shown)


ManoloFLTK:
@mikeandmore I am sorry to write this, but I'm not going to change the present source code which correctly handles changes in the DPI values of displays unless a PR gives evidence that a change is necessary. This PR adds calls to XDisplayString(), XOpenDisplay(), XCloseDisplay(). Are we sure that it's better than a call to popen()? Even so, changes of the display DPI values are rare events, which can't occur at high rate. Is it useful to process them a few hundredths of a second faster?

mikeandmore:
@ManoloFLTK Performance is a very minor argument.

The major reason is:
1. We rely on an external dependency. We don't know if that exists or not. Even if it exists, it might not be the binary we are looking for, for example, when another binary happens to name `xrdb` in the `$PATH`.
2. Executing an external binary may be troublesome for sandboxing, making sandboxing rules much more complicated.
3. As you and @Albrecht-S pointed out, if application sets up a custom `fl_display`, then executing `xrdb` won't work because that always open the `$DISPLAY`.

Also, now reading XFT DPI settings are all done in the same function, making the code easier to follow in the long run.

Albrecht-S:
@mikeandmore Thanks, your arguments sound plausible.

I feel inclined to add to my previous arguments: FLTK sets the `DISPLAY` environment variable at startup in the environment, particularly if it has been overridden with the commandline argument `-di[splay] host:n.n`. This means that any program we run in this environment can use the DISPLAY variable to open the same display. To be sure I tested this with a VM and `ssh -X` using both the forwarded X connection and the host's X display. This works as intended.

This would enable us to use the pipe as in the original code and read from the correct display but this doesn't invalidate @mikeandmore's concerns about running another program like `xrdb` and about sandboxing.

OTOH `Display *new_dpy = XOpenDisplay(0);` as in the **_previous patch_** would very likely also open the correct X display given that the `DISPLAY` environment variable should be set as needed.

That said, I tend to agree with Mike that his patch looks like an improvement. Of course we need to verify that the right display would indeed be used...

@ManoloFLTK What do you think about these arguments?

mikeandmore:
@Albrecht-S Application can also declare `extern Display *fl_display` to set a custom display. I believe global variable usage are pretty popular in FLTK applications.

ManoloFLTK:
@mikeandmore Many thanks for finding a pure Xlib way to get a possibly changed DPI value of an X11 display while an FLTK app keeps running. This solution is now committed to the master repository. I close this PR.

PR #914, opened 2024-02-19
Fix two mis-use of cairo

mikeandmore:
The first patch fixes a use-after-free of cairo context, which can cause clipping fails.

The second patch makes sure we call `cairo_surface_flush()`, which is required by cairo. Although currently, it does not seem to affect anything, but in the future, this might change.

### DIFF
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -1215,24 +1215,15 @@ static void react_to_screen_reconfiguration() {
 
 #if USE_XFT
 static void after_display_rescale(float *p_current_xft_dpi) {
-  FILE *pipe = popen("xrdb -query", "r");
-  if (!pipe) return;
-  char line[100];
-  while (fgets(line, sizeof(line), pipe) != NULL) {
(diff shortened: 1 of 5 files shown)


mikeandmore:
Ouch, I'm should have rebase with the main branch. Maybe instead of looking at my two patches in parallel, we can look at first one instead?

Albrecht-S:
Mike, thank you very much for your activities and suggestions in PR's. However, it's difficult to see what you mean with "first patch" etc.. What we need is:

- please create one branch per issue in your repository (you might need three or four branches for all your proposals so far)
- give each branch a useful name
- then create one PR per issue, i.e. per branch, and explain what you want to do and why

This way we can select one or more of your PR's and different developers can work on different PR's which is important since e.g. Cairo issues and other X11 issues (PR's) will very likely be worked on by different devs. Thanks for your understanding.

Note: I'm not closing this PR for now but I would strongly suggest that you do as described above. Thanks.

mikeandmore:
@Albrecht-S Thanks. That's a good idea. Let me do it.

There will be more PRs to come, fortunately, most of them are independent.

PR #918, opened 2024-02-24
Added horizontal scrollbar to Terminal widget

jgriffitts:
I have added a horizontal scrollber to Fl_Terminal, as I offered to do [here](https://groups.google.com/g/fltkgeneral/c/3b0XH8f1rnA/m/tBxoC5khAAAJ). Since this is the first time I've had my fingers in fltk source, I hope you'll review what I did and let me know what I screwed up. 
I added the following functions to the API:
```
  int   hscrollbar_size(void) const;
  void  hscrollbar_size(int val);
  int   hscrollbar_actual_size(void) const;
  void  hscrollbar_style(HScrollbarStyle val);
  HScrollbarStyle hscrollbar_style(void) const;
  int   hscroll_col(void) const;
  void  hscroll_col(int col);
enum HScrollbarStyle {
    HS_OFF        = 0x00,     ///< scrollbar is always invisible
    HS_AUTO       = 0x01,     ///< scrollbar is visible if widget has been resized in a way that hides some columns (default)
    HS_ON         = 0x02      ///< scrollbar is always visible
  };
```
I have tested a fair amount on Ubuntu and Windows 10, and built in both places.

My test hack program is pasted below. You'll see lots of commented-out function calls that can be uncommented to test different things.
```
#include <FL/Fl_Double_Window.H>
#include <FL/Fl_Terminal.H>

Fl_Double_Window* win = 0;
Fl_Terminal* term = 0;


int main(int argc, char** argv) {
    win = new Fl_Double_Window(900, 320, "App");
    win->begin();
    term = new Fl_Terminal(0, 0, win->w(), win->h());
    term->label("term");
    win->end(); 
    win->resizable(win);
    win->show(argc, argv); 

    term->textbgcolor_xterm(0);
    term->textattrib(Fl_Terminal::BOLD);
    term->append("Bold text that will immediately be erased.\n");
    // term->margin_left(99);
    // term->margin_top(59);
    // term->margin_right(19);
    // term->margin_bottom(29);

    win->resize(win->x(), win->y(), win->w()/2, win->h());
    // term->hscrollbar_style(Fl_Terminal::HScrollbarStyle::HS_ON);
    // term->hscrollbar_style(Fl_Terminal::HScrollbarStyle::HS_OFF);
    //  term->hscrollbar_style(Fl_Terminal::HScrollbarStyle::HS_AUTO);

    term->clear();
    term->textattrib(Fl_Terminal::NORMAL);
    term->append("Non-bold text");
    for (int i=0; i<30; ++i)
        term->printf("%d\n", i);
    term->append("\n0         1         2         3         4         5\n");
    for (int i=0; i<15; ++i)
        term->append("0123456789");

    // term->printf("\n<<%d>>\n", term->hscroll_col());
    // term->hscroll_col(1);
    // term->printf("<<%d>>\n", term->hscroll_col());
    // term->hscroll_col(2);
    // term->printf("<<%d>>\n", term->hscroll_col());
    // term->hscroll_col(3);
    // term->printf("<<%d>>\n", term->hscroll_col());
    // term->hscroll_col(4);
    // term->printf("<<%d>>\n", term->hscroll_col());
    // term->hscroll_col(5);
    // term->printf("<<%d>>\n", term->hscroll_col());
    term->hscroll_col(16);
    term->printf("<<%d>>\n", term->hscroll_col());
    // // term->hscroll_col(44);
    // // term->printf("<<%d>>\n", term->hscroll_col());
    // // term->hscroll_col(1000);
    // // term->printf("<<%d>>\n", term->hscroll_col());
    return Fl::run();
}
```
I'll be doing the Rust bindings for this over the weekend, and then do some end-to-end testing.


jgriffitts:
In case anyone is looking at this already: I discovered bugs today and commited a fix

erco77:
Giving it a look-see now.

We can probably change `enum HScrollbarStyle` to `enum ScrollbarStyle` and the API adjusted accordingly so it can be used for both scrollbars in the future.

Regarding HSCROLLBAR_RANGE=20, should that maybe just be `display_columns()` (perhaps with a minimum enforced), so that each step is a single char? I tried it and it seemed to work well, e.g. commenting out the const int in the .H file, and declaring this at the top: `#define HSCROLLBAR_RANGE display_columns()`. If that sounds OK, I'd probably replace the macro with just `display_columns()`.

Still working thru the code, will add notes as I go, and will probably propose a patch.

erco77:
Seems stable by the way -- I like it! Will try the options in your test program too. I'd probably end up adding those commented out tests to the already huge terminal test program as UI options. I might have to break up that UI into tabs, lol.

erco77:
One note:

The horizontal-specific methods/variable related to the trough size can be removed, e.g. `hscrollbar_size()` and `hscrollbar_size_` -- the `scrollbar_size()` applies to ALL scrollbars (the size of the "troughs"), serving both scrollbars equally. 

The default inherits the global "scrollbar size" when zero, which affects all scrollbars, and we allow the widget to have different sized scrollbars if needed, but the size applies to both. We shouldn't offer separate controls for horizontal+vertical trough size; one should be the same as the other. (This is how all other widgets with h/v scrollbars do things as well; one size applies to both).

I'll propose all my suggested changes in a patch forthcoming.

erco77:
I did notice a very small artifact above the horiz scrollbar, probably just a small math issue; I'll see if I can identify it:

scrollbar-artifact

erco77:
I need to hit pause on looking this over, as I've got some guests coming.

Dropping a _work in progress_ patch here, just to remember where I am so far, as I haven't fully tested my changes yet:
[erco-terminal-mods_v1.patch.txt](https://github.com/fltk/fltk/files/14474588/erco-terminal-mods_v1.patch.txt)

Liking your horiz scroll code, and with some mods would like to get it into master. I can handle the merge.
Will try to pick up on this later tonight

jgriffitts:
Thanks for your thorough review. For `hscrollbar_size` I was trying for maximum flexibility, but I agree completely with being consistent with other widgets.  I know of a downside for using smaller resolution for `HSCROLLBAR_RANGE`: there's a side-scroll button on my trackball, and this will make it move in inconveniently small staps. That's not a big problem as far as I'm concerned -- I 'm not in the habit of using that button personally and it works inconsistently in different programs. There may be people out there who care.

erco77:
(My guests are running late so I have a smidge of time)
Regarding the scroll range, yes, we could adjust the range so it takes maybe four ticks per char or some such. Pretty sure the underlying valuator widget allows floating point so that the range can still be in chars, but the UI scrolling can be done at a different step rate of e.g. 0.25 for more fine resolution that scales with the column size. We can investigate that after as a fine-tuning step.

erco77:
Found the cause of the artifact; addressing that in this follow up patch, along with a few other small things: [erco-terminal-mods_v2.patch.txt](https://github.com/fltk/fltk/files/14478369/erco-terminal-mods_v2.patch.txt)

Also, added interactive adjustments to your test program to flesh out a few of the above addressed issue: [test-hscroll.cxx.gz](https://github.com/fltk/fltk/files/14478429/test-hscroll.cxx.gz)

jgriffitts:
I have applied your v2 patch here locally and tested in Linux and Windows. I see nothing to complain about. I verified the artifact that you found and that it disappeared with your patch, and can see where my height calculation was defective.

Let me know if you want me to push the patched sources up into this PR.

erco77:
New [erco-terminal-mods_v3.patch.txt](https://github.com/fltk/fltk/files/14485682/erco-terminal-mods_v3.patch.txt) and new test program [test-hscroll.cxx.gz](https://github.com/fltk/fltk/files/14485690/test-hscroll.cxx.gz) which includes interactive adjusters to stress test for bugs.

Some notes on the new patch:

- I didn't realize the horiz scrollbar wasn't itself using columns as its units, as the vertical scroll does. So I reworked the horiz initialization to match the vertical scrollbar's so the range of values represents actual column numbers. This allows the `hscrollbar->value()` to reflect the actual column# offset, and simplifies the code in draw().

- Doing the above also removes the need for the new `hscroll_col()` methods, simplifying the API, so the app can simply set column numbers using `hscrollbar->value()` directly, the hscrollbar's `minimum()/maximum()` being the range of values allowed for `value()`.

- I modified the horiz scrollbar's resolution to 0.25 so there's at least 4 ticks of the scrollbar per column changed. This makes scrolling smoother when there's not many columns to scroll. Let me know if that helps with your trackball setup.

- I noticed the background colors weren't scrolling, so applied your start_col/end_col logic to the `draw_bg()` code as well. This can be seen in the new test program with the red/grn/blu background colors on the last line.

This is starting to shape up pretty well; was able to shorten the code a bit, which is always a good thing.

Gonna take another break, but will revisit with more stress testing and run a comb thru the code to make sure to remove unused/commented out code, and double check the docs. I should probably also add some docs describing how an app can set scrollbar row/column positions using the scrollbar's range.

jgriffitts:
Regarding the trackball horizontal scroll: some of these devices have a function where pushing the scrollwheel to the left or right causes a click that moves the horizontal scroll one tick to the left or right. If horiz scroll resoution is set too fine, then it takes a large number of these clicks to move the scroll significantly. The click is awkward to do (at least for me) so moving a large number of ticks can take a while. There is a repeat function, but the rate is slow (possibly configurable, I haven't looked). 

Different apps handle this function differently. I just tried it in VS Code in Windows, and each click moved about 6.5 characters to the side. LibreOffice Writer seems to move 3/4" for each click. MS Visual Studio moves one column per click.

I personally don't use this function much, but if one were to optimize a GUI for it, the step size should probably be multiple columns. I am not advocating paying much attention to this.

erco77:
Ya, I'll let you know when I think one of my patches is worthy, then you can apply and I'll do the merge. Still doing stress tests.

I should probably sync my patches with fltk:master (I added a text() method recently), so that the squash+merge will go easily.

erco77:
OK, this v4 patch is what I'd suggest you apply to your fork: 
[erco-terminal-mods_v4_final.patch.txt](https://github.com/fltk/fltk/files/14502428/erco-terminal-mods_v4_final.patch.txt)

In this patch:
- Mainly massaged the docs
- Removed my old, unnecessary protected `scrollbar_width(void)` method in favor of your technique, which is more accurate.

erco77:
Ew, OK, I noticed your fork didn't do our gitlab checks (not sure why), so I pushed it to my fork to check, and found a few errors on the linux build, which are just a few one line changes I've addressed in this followup v5 patch: [erco-terminal-mods_v5_final.patch.txt](https://github.com/fltk/fltk/files/14502750/erco-terminal-mods_v5_final.patch.txt)

Here's the errors this follow up v5 patch solves:

image

jgriffitts:
I have applied the v5 patch and this should be ready for you!

I have been compiling this on Linux all along with no warnings or errors. No idea why the checks were throwing errors, but I guess we could start comparing versions if it's important. It also compiles fine for me _after_ the patch. I fiddled with my github repository, went into "Actions" and clicked the box to initialize workflows, and it looks like the checks ran this time.

erco77:
@Albrecht-S  - requesting your assistance on this PR.

PR's scare me because it's sometimes not clear to me what I'll end up with after doing it.

To test, I tried github's recommended command line equivalents on my local copy of fltk:master, and it made a complicated "merge commit". So I abandoned that, recloned fltk:master and tried again using `git pull --rebase https://github.com/jgriffitts/fltk.git master` but now I'm getting weird conflicts.

It would be very easy for me to simply copy the two changed files from my local copy of jgriffitts master and add them to my local fltk:master and do a simple commit like I always do, but I'd like to understand the proper PR technique.

erco77:
@jgriffitts I just noticed your original code reclaims the screenspace when the horiz scrollbar disappears, whereas my mods somehow disabled that. I'll track that down and fix while I await some info from @Albrecht-S about the PR.

Albrecht-S:
> @Albrecht-S - requesting your assistance on this PR.

@erco77 I'll look into it to find a proper solution. The current situation is ... well ... "suboptimal" for two reasons:
1. The entire PR was done in the master branch instead of its own branch
2. The upstream/master (fltk/master) branch has been "merged downstream" twice which makes it harder.

I'm sure I can find a clean solution but I'm currently busy with Windows stuff. Give me a few hours, I'll try to do it today (in CET).

jgriffitts:
> The entire PR was done in the master branch instead of its own branch

This is my fault and I knew better, but didn't notice the error for a while. If you like, I'll submit a _new_ PR for this, done in its own branch, and this one can be abandoned. (Or is there a way to modify this PR? I'm not a github guru.) I should be the one to do the extra work to fix my screw-up.

Albrecht-S:
@erco77 @jgriffitts Please see my rebased PR #928. I couldn't resist trying to "fix" the PR and it was pretty simple and done much quicker than I had anticipated. Basically I ...
- added a new branch on top of @jgriffitts' master branch
- rebased this branch on top of origin/master (current FLTK master)
- fixed trailing whitespace and copyright year in another commit
- pushed this to my repo (preserving all commits)

@jgriffitts Thanks for your offer. If you accept my branch, please confirm and @erco77 may then merge it. Squashing it would (as you probably know) make only one commit of all commits in this branch, that's what we prefer for such local changes.

@erco77 ISTR that there was a tiny difference I couldn't verify easily (one additional line, but I don't remember what it was). Please compare and maybe you'll need to remove that additional line. Feel free to add commits to my branch (devs are allowed to do that).

Finally you may use GitHub's "Squash and merge" button to merge it.

WRT feature freeze: I believe it's better to modify the terminal widget now rather than adding this feature later because it would certainly break the ABI, so please feel free to merge it before the release - if you find it's OK. Unfortunately I couldn't afford the time for a full review (yet). But it looks like Jonathan did a good job, thanks to @jgriffitts for this.

Albrecht-S:
@jgriffitts Before I forget: many thanks for your work on FLTK and fltk-rs. That's very much appreciated!

> is there a way to modify this PR? I'm not a github guru

Usually you can rebase and force-push your own branch that's used in a PR. We (the FLTK team) always prefer rebasing rather than big merges as seen above, particularly merging "downstream". You and other (FLTK) devs with write access can even add commits to your branch if you didn't explicitly forbid it (it's a PR option).

That all said, this works well if you started the PR with your own branch. I don't think that you can "change" a PR to use another branch of yours. Therefore a new PR would have been appropriate but I did this already.

erco77:
Thanks Albrecht!
I'm busy today, but I'll revisit tonight to figure out your solution.

jgriffitts:
Superseded by [PR #928](https://github.com/fltk/fltk/pull/928)

PR #948, opened 2024-04-08
Fix Fl_Terminal::handle_unknown_char()

Albrecht-S:
I'm opening this PR for myself and for review by @erco77. I intend to push this myself with necessary modifications after review, e.g. remove the mods of the demo program, and see question below.

**_Problem:_** see screenshot and details below:
Fl_Terminal_unknown

The red marks show what's wrong, the green mark shows the same after the patch (correct).

The old code called `Fl_Terminal::handle_unknown_char()` when `Fl_Terminal::plot_char()` encountered an illegal UTF-8 character which put the "unknown" char in the normal output stream rather than at the requested position (if `show_unknown()` was enabled). I fixed this issue in my patch by adding an overridden method with `(drow,dcol)` arguments:
```diff
   int handle_unknown_char(void);
+  int handle_unknown_char(int drow, int dcol);
```

I believe this approach is correct - so far (see screenshot).

The implementation is as follows:
```C++
int Fl_Terminal::handle_unknown_char(int drow, int dcol) {
  if (!show_unknown_) return 0;
#if (0)
  escseq.reset();               // disable any pending esc seq to prevent eating unknown char
#endif
  int len = (int)strlen(unknown_char);
  Utf8Char *u8c = u8c_disp_row(drow) + dcol;
  u8c->text_utf8(unknown_char, len, *current_style_);
  return 1;
}
```

@erco77 **_My question:_** do we need (want?) to call `escseq.reset();` here which is currently disabled in my patch?

The reason why I disabled it (as opposed to the "other" `handle_unknown_char()` method) is that `plot_char()` generates IMHO "out of band" data and should not interfere with the current output stream which may have left a partial escape sequence. However, this is debatable, and that's why I'm asking you, Greg. What do you think?

Thanks for reading and looking into it. If you agree I will remove the '#if (0) ... #endif' block entirely before rebasing and pushing.

Note 1: while I was at it I made related documentation changes, minor reformatting, and I added a protected static class variable `const char *Fl_Terminal::unknown_char = "¿";` (this is the initalization in the C++ file) - basically to avoid redundant code. In the future we might want to add public getter and setter methods so users can fine tune the "unknown character".

Note 2: this PR includes the modified demo program `test/handle_keys.cxx` which generates the above shown output. Run this with the old or new code...

### DIFF
--- a/FL/Fl_Terminal.H
+++ b/FL/Fl_Terminal.H
@@ -1027,6 +1027,7 @@ public:
   void append(const char *s, int len=-1);
 protected:
   int handle_unknown_char(void);
+  int handle_unknown_char(int drow, int dcol);
   // Drawing
   void draw_row_bg(int grow, int X, int Y) const;
   void draw_row(int grow, int Y) const;
@@ -1170,6 +1171,9 @@ public:
   // API: Show unknown/unprintable chars
   bool  show_unknown(void) const;
   void  show_unknown(bool val);
+protected:
+  static const char *unknown_char; ///< "unknown" replacement character
+public:
   // API: ANSI sequences
   bool  ansi(void) const;
   void  ansi(bool val);
--- a/src/Fl_Terminal.cxx
+++ b/src/Fl_Terminal.cxx
@@ -38,6 +38,12 @@
 #include "Fl_String.H"
 
 /////////////////////////////////
+////// Static Class Data ////////
+/////////////////////////////////
+
+const char *Fl_Terminal::unknown_char = "¿";
+
+/////////////////////////////////
 ////// Static Functions /////////
 /////////////////////////////////
 
@@ -2995,8 +3001,10 @@ const Fl_Terminal::Utf8Char* Fl_Terminal::utf8_char_at_glob(int grow, int gcol) 
 void Fl_Terminal::plot_char(const char *text, int len, int drow, int dcol) {
   Utf8Char *u8c = u8c_disp_row(drow) + dcol;
   // text_utf8() warns we must do invalid checks first
-  if (!text || len<1 || len>u8c->max_utf8() || len!=fl_utf8len(*text))
-    { handle_unknown_char(); return; }
+  if (!text || len<1 || len>u8c->max_utf8() || len!=fl_utf8len(*text)) {
+    handle_unknown_char(drow, dcol);
+    return;
+  }
   u8c->text_utf8(text, len, *current_style_);
 }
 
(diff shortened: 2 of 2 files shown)


Albrecht-S:
Note 3: Unrelated, however this is something I noticed while testing (code in the modified `handle_keys.cxx`):
```C++
  for (int i = 26; i < 29; i++) {
    char buf[40];
    int len = sprintf(buf, "   ASCII(%2d) = '%c'.\n", i, i);
    win->tty->append(buf, len);
  }
```

@erco The output is also shown in above screenshot. As you can see I wrote control chars in a loop with `Fl_Terminal::append()`. You may want to extend the loop from 0 to 31 but the only **_unexpected output_** (from my POV) was with the Escape character. It seems to consume the Esc char and the following `'` char (illegal Esc sequence) and doesn't output anything for this. Do you think that this should (and can) be "fixed"? Should I open an issue, or do you want to look into this independently from this PR?

erco77:
Ah, interesting.

OK, well the intention of the handle_unknown_char() was only to output a character, not change state. I don't think it should do anything to the ESC state; that's better done at the caller, as the ESC handling stuff is higher level.

Also, the plot_char() documents it doesn't handle ESC sequences, so in this case that function shouldn't do anything to the ESC state either.

Regarding having a version of `handle_unknown_char()` that accepts row,col, sure, that sounds good.

erco77:
Regarding the control character loop, is the request to have the sequence 'ESC' output a  "¿" instead of just ignoring the sequence? If so, sure, and yes, open a separate issue and I can handle that.

Of course if you want to be able to print control codes and not have ESC be parsed as an ANSI sequence, just turn `ansi()` mode off before calling append(). (I'm not really sure how fl_draw() will handle drawing an ESC character.. I suppose it depends on what utf8 fonts are accessible, so that it might show the symbol "␛" instead of something random and weird)

Albrecht-S:
@erco77 OK, I removed the `#if` block as written above and pushed my changes to master (05a95e146b2429154b360030d098f0dae5555706). Thanks for your comments.
You wrote:
> the intention of handle_unknown_char() was only to output a character, not change state. I don't think it should do anything to the ESC state ...

But the "original" `handle_unknown_char()` method does exactly that, it calls:
```C++
  escseq.reset(); // disable any pending esc seq to prevent eating unknown char
```
... and I believe this is OK _in this context_. This was the only reason why I copied - but disabled - this statement in my version, and why I asked you ...

I'll open another issue for control character handling, particularly the `Esc` character.

Issue #753, opened 2023-07-18
Give the option to disable the static library


rempas:
**Is your feature request related to a problem? Please describe.**
Having to build the static library that I never use.

**I am using FLTK**
 - 1.4.0

**Describe the solution you'd like**
I would like to have on option on CMAKE to disable the static library and only build the shared one. There is that option for the shared library (as it should) for people that would only use the static one. I just don't understand why there isn't one for the static one as well.

**Describe alternatives you've considered**
I suppose that I can remove it manually from the Make file but that's not a proper way of doing it and it doesn't make sense...

**Additional context**
I do understand and respect the idea of FLTK been small enough to be statically linked but some people would still want to use the shared library anyways. So I don't see why having to build it anyways. Thank you for your time reviewing that!

rageworx:
Dear @rempas ,
I like to ask you about why need to disable static library ?
As I know, CMAKE provides an option for SHARED_LIBRARY.
![image](https://github.com/fltk/fltk/assets/6948225/0b428236-e741-4673-9fb2-576290426502)

FLTK is builds static library as default, and shared is an option, then you can use only shared library.
Is it something different ?

rempas:
> Dear @rempas , I like to ask you about why need to disable static library ? As I know, CMAKE provides an option for SHARED_LIBRARY. ![image](https://user-images.githubusercontent.com/6948225/254182026-0b428236-e741-4673-9fb2-576290426502.png)
> 
> FLTK is builds static library as default, and shared is an option, then you can use only shared library. Is it something different ?

I think that it would be nice if it was more configurable and flexible. Like I said, not everyone (I mean, I don't think I'm not the only one in the whole world... right?) would want to use the static library so I think that there is no point and having to build it (hence wait for it to compile) and then install it and have it in your system.

Again, nothing to big, the compile times aren't so long (unless you have an older system) and the size of the library isn't so big but again, it's just for a better feel and image for the library.

Albrecht-S:
@rempas Thank you for the request. I can see why someone may want this, one reason is to enable only shared FLTK library builds of FLTK projects - besides the compilation time.

@rageworx OPTION_BUILD_SHARED_LIBS only **adds** the shared FLTK libs but does not disable building the static libs.

FWIW: CMake has an own variable named `BUILD_SHARED_LIBS` for users to configure **either** building shared **or** static libs, but this option is not used by the FLTK build system. FLTK currently builds always static libs and optionally shared libs.

[...]

PR #954, opened 2024-04-11
CMake: Use imported targets for PkgConfig packages

jwillikers:
This carries include directories and the required link flags. Several libraries brought in through `pkg_check_modules` aren't linked correctly using the `_LDFLAGS` variables. Instead, they link the libraries directly with `-l`. This is problematic because it doesn't properly pull in the actual directory of the library being linked when it is not in a system library directory. Likewise, the necessary include directories for several of these targets aren't properly set either. Linking against the imported targets automatically pulls in the necessary include directories.

This issue arose trying to build with Wayland support using Conan packages for the Wayland libraries. They provide the necessary pkg-config files, but since the libraries and includes are in non-standard locations, Wayland support can't be built.


Albrecht-S:
Thanks for the PR, I'm going to check it now...

Albrecht-S:
Patch looks good but my first tests show that it doesn't work if WAYLAND is configured to be used on my Linux host. Works well w/o Wayland though. I'll try to find out what's going on in about an hour from now.
FYI:
### With Wayland:
```
-- Configuration Summary for FLTK 1.4.0 generated by CMake 3.25.1 --
[...]
-- 
-- Build configuration      Debug

-- Image Libraries          JPEG = System: /usr/lib/x86_64-linux-gnu/libjpeg.so
--                          PNG  = System: /usr/lib/x86_64-linux-gnu/libpng.so
--                          ZLIB = System: /usr/lib/x86_64-linux-gnu/libz.so

-- Use Wayland              Yes (can also run as X11 client)
-- Use system libdecor      No
-- All drawing uses Cairo   Yes
-- Use Pango                Yes
-- Fl_Cairo_Window support  No
-- Use std::                No

-- End of Configuration Summary --

-- Configuring done
CMake Error at src/CMakeLists.txt:773 (target_link_libraries):
  Target "fltk" links to:

    PkgConfig::DBUS

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
...
```
### Reconfiguring w/o Wayland:
```
$ cmake -D FLTK_BACKEND_WAYLAND:BOOL=OFF .

-- Configuration Summary for FLTK 1.4.0 generated by CMake 3.25.1 --
[...]
-- 
-- Build configuration      Debug

-- Image Libraries          JPEG = System: /usr/lib/x86_64-linux-gnu/libjpeg.so
--                          PNG  = System: /usr/lib/x86_64-linux-gnu/libpng.so
--                          ZLIB = System: /usr/lib/x86_64-linux-gnu/libz.so

-- Use Wayland              No (X11 is used)
-- All drawing uses Cairo   Yes
-- Use Pango                Yes
-- Fl_Cairo_Window support  No
-- Use std::                No

-- End of Configuration Summary --

-- Configuring done
-- Generating done
```
... works.

Albrecht-S:
@jwillikers: **_Progress report:_** I found two instances of `pkg_check_modules` that didn't have the `IMPORTED_TARGET` argument but referred to the **target** later in the build: `DBUS` and `GTK`. After adding the missing argument the build proceeded. :-)

Still testing though...

jwillikers:
@Albrecht-S My bad there, I've added those the `IMPORTED_TARGET` keyword to those two `pkg_check_modules` calls.

Albrecht-S:
OK, thanks for adding these (and force-pushing your branch). I'm testing further, please give me some time and don't update your branch while I'm testing - this would confuse me. ;-)

Meanwhile I found another issue: creation of the `fltk-config` shell script "exports" the CMake targets which is obviously not correct, but I'll investigate this myself. This is something I need to work on anyway...

I'll let you know if I need help.

Albrecht-S:
@jwillikers Sorry for the delay, I was distracted by other stuff. Finally I decided to merge your PR as-is and to push another commit after that. The reason was that some variables used to generate the `fltk-config` script were updated erroneously in your PR but I changed a lot more to "revert" this and to refactor and document the usage of these variables in commit b4cf1a9824f2c4ba9596044962d3af36e3ca3d99.

I'm pretty sure that our two commits together do what you intended and that the generation of `fltk-config` was improved by these two commits. There's a gray zone though which I'm not sure about: if it was necessary to add some include directories to the generated `fltk-config` script, then this has not been done yet because we can't add CMake targets to a shell script. The effect is that the correct libraries were added but there may be missing include directories. This needs investigation but I'm leaving this open for now.

That said, thanks for your support to improve the FLTK build system.

jwillikers:
> @jwillikers Sorry for the delay, I was distracted by other stuff. Finally I decided to merge your PR as-is and to push another commit after that. The reason was that some variables used to generate the `fltk-config` script were updated erroneously in your PR but I changed a lot more to "revert" this and to refactor and document the usage of these variables in commit [b4cf1a9](https://github.com/fltk/fltk/commit/b4cf1a9824f2c4ba9596044962d3af36e3ca3d99).
> 
> I'm pretty sure that our two commits together do what you intended and that the generation of `fltk-config` was improved by these two commits. There's a gray zone though which I'm not sure about: if it was necessary to add some include directories to the generated `fltk-config` script, then this has not been done yet because we can't add CMake targets to a shell script. The effect is that the correct libraries were added but there may be missing include directories. This needs investigation but I'm leaving this open for now.
> 
> That said, thanks for your support to improve the FLTK build system.

No worries, I thought your response was incredibly fast. I noticed that there could be missing include directories using just `_LDFLAGS`, but testing on your latest commit doesn't have any issues with missing include directories for me on Linux via Conan. The Conan package doesn't use or ship the `fltk-config` script, which is probably why this wouldn't turn up any errors for me. If you wanted to, you could use the `_INCLUDE_DIRS` or `_CFLAGS` variables to obtain the necessary include directories, but I don't need that.

I still need to see if I run into any issues on other platforms, but it's pretty exciting to be able to drop those patches. I think the only remaining patch left for the Conan package just disables building the static `fltk` library at all when doing a shared library build.

Albrecht-S:
> No worries, I thought your response was incredibly fast.

Thanks.

> I noticed that there could be missing include directories using just _LDFLAGS, but testing on your latest commit doesn't have any issues with missing include directories for me on Linux via Conan. The Conan package doesn't use or ship the fltk-config script, which is probably why this wouldn't turn up any errors for me.

If `Conan` doesn't use the `fltk-config` script it shouldn't be affected by my commit which targets only the generation of this script. We didn't get any reports about missing include dirs but OTOH I know (and you wrote) that using the CMake target in a pure CMake build includes both the linker flags and the include directories. So what I wanted to say: If there's a need to use an additional include directory because pkg-config "finds" packages in non-standard dirs, then this works in a pure CMake build (like Conan seems to be) - however we are ignoring these potential include directories in `fltk-config` because I only picked the linker flags so far. There would be a risk that `fltk-config --compile` wouldn't work in such a scenario, but since I'm aware of this I can investigate this separately.

Albrecht-S:
> I think the only remaining patch left for the Conan package just disables building the static fltk library at all when doing a shared library build.

The FLTK "philosophy" has always been to build the static libs and opt-in to building shared libs, and that's why we ignore CMake's `BUILD_SHARED_LIBS` variable in general and always use `STATIC` as default. In fact someone opened an issue to build only shared libs (#753) but I dismissed it recently because it became too complex, given that we'd have to build our tools like `fluid`, `fltk-options`, and all test programs with shared libs if we accepted this request. Sure, if we only wanted to build the FLTK libs alone, no executables (fluid, fltk-options), and no test programs, then we could probably have done it (see [this comment](https://github.com/fltk/fltk/issues/753#issuecomment-1642262631)).

One problem with a shared-library-only build would be on Windows where we need to execute `fluid` during the build, but if it was built using the shared lib (DLL), then this DLL would have to be in a place where the system (i.e. fluid) would find it during build time, etc. etc..

Another "problem" was the definition of appropriate and backwards compatible CMake option names that would allow to disable the static libraries and build only the shared libs. And last but not least, what would the CMake target names be if either the static or the shared libs were built? Currently we have `fltk::fltk` and (optional) `fltk::fltk-shared`. If we built only the shared libs, what would their target names be, maybe `fltk::fltk`? How would consumers of FLTK know that?

That said, if you have a patch that considers all of these issues I'd be all ear and would be interested to see it. Can you point me at an existing patch in Conan?

jwillikers:
> > I think the only remaining patch left for the Conan package just disables building the static fltk library at all when doing a shared library build.
> 
> The FLTK "philosophy" has always been to build the static libs and opt-in to building shared libs, and that's why we ignore CMake's `BUILD_SHARED_LIBS` variable in general and always use `STATIC` as default. In fact someone opened an issue to build only shared libs (#753) but I dismissed it recently because it became too complex, given that we'd have to build our tools like `fluid`, `fltk-options`, and all test programs with shared libs if we accepted this request. Sure, if we only wanted to build the FLTK libs alone, no executables (fluid, fltk-options), and no test programs, then we could probably have done it (see [this comment](https://github.com/fltk/fltk/issues/753#issuecomment-1642262631)).
> 
> One problem with a shared-library-only build would be on Windows where we need to execute `fluid` during the build, but if it was built using the shared lib (DLL), then this DLL would have to be in a place where the system (i.e. fluid) would find it during build time, etc. etc..
> 
> Another "problem" was the definition of appropriate and backwards compatible CMake option names that would allow to disable the static libraries and build only the shared libs. And last but not least, what would the CMake target names be if either the static or the shared libs were built? Currently we have `fltk::fltk` and (optional) `fltk::fltk-shared`. If we built only the shared libs, what would their target names be, maybe `fltk::fltk`? How would consumers of FLTK know that?
> 
> That said, if you have a patch that considers all of these issues I'd be all ear and would be interested to see it. Can you point me at an existing patch in Conan?

Yeah, I could tell it wouldn't be very straightforward for fltk since you support building both shared and static libraries simultaneously. I don't really know a good way of handling that in CMake. I think most projects that do this end up rolling their own way of doing it unfortunately, and I'm not very familiar with it. I've seen that Meson projects have a much easier time doing this since they have a `both_libraries` function that can be used when creating a library.

The patch for Conan is [here](https://github.com/conan-io/conan-center-index/blob/master/recipes/fltk/all/patches/1.3.8-0003-build-static-only-on-static.patch) and is super simple because we don't need to support all of the use cases you do.

Albrecht-S:
> The patch for Conan is ... super simple because we don't need to support all of the use cases you do.

@jwillikers OK, I see. Thanks for the info anyway.

I conclude that you will have to live with this patch for now. I won't touch this in 1.4.0 (and maybe not in any 1.4.x version) but maybe later. 1.5.0 will be an opportunity to revisit this feature.

Albrecht-S:
@jwillikers Unfortunately I had to change (kinda revert) parts of your PR because it broke user projects. See commit a9d3d3e3fbefc80278392a88c0a81a5bbb97eb51 for the full list of changes. Note that I refactored the assignments (using a macro for more flexibility) which makes the diff appear larger than it actually is.

**_Summary of changes:_**
- use _new macro_ `append_optional_libs()` to update the "OPTIONAL_LIBS" variable which ...
- extracts `INTERFACE_LINK_LIBRARIES` from the previously used `PkgConfig::TARGETNAME` targets ...
- instead of using the target name literally, and finally ...
- remove duplicates from `OPTIONAL_LIBS` and ...
- use `target_link_libraries(<target> *PRIVATE* ${OPTIONAL_LIBS})` rather than `PUBLIC`.

I hope (and I'm confident) that this doesn't break your Conan build. Please test again and report here whether it still works for you. TIA.

jwillikers:
> @jwillikers Unfortunately I had to change (kinda revert) parts of your PR because it broke user projects. See commit [a9d3d3e](https://github.com/fltk/fltk/commit/a9d3d3e3fbefc80278392a88c0a81a5bbb97eb51) for the full list of changes. Note that I refactored the assignments (using a macro for more flexibility) which makes the diff appear larger than it actually is.
> 
> **_Summary of changes:_**
> 
>     * use _new macro_ `append_optional_libs()` to update the "OPTIONAL_LIBS" variable which ...
> 
>     * extracts `INTERFACE_LINK_LIBRARIES` from the previously used `PkgConfig::TARGETNAME` targets ...
> 
>     * instead of using the target name literally, and finally ...
> 
>     * remove duplicates from `OPTIONAL_LIBS` and ...
> 
>     * use `target_link_libraries(<target> *PRIVATE* ${OPTIONAL_LIBS})` rather than `PUBLIC`.
> 
> 
> I hope (and I'm confident) that this doesn't break your Conan build. Please test again and report here whether it still works for you. TIA.

@Albrecht-S The imported CMake targets also carry the necessary include directories, so you'll need to accommodate that somehow if your just linking libraries. I got this error with the latest commit no master to show this.

```
/home/jordan/.conan2/p/b/fltk85379015e07d2/b/src/src/drivers/Wayland/Fl_Wayland_Screen_Driver.H:27:10: fatal error: wayland-client.h: No such file or directory
   27 | #include <wayland-client.h>
      |          ^~~~~~~~~~~~~~~~~~
```

I'd be curious to know how project is being consumed where those imported targets aren't available, which must be from the corresponding `find_package` calls not being reached. Is it from an installation of FLTK where the CMake modules are missing those targets?

Albrecht-S:
> The imported CMake targets also carry the necessary include directories, so you'll need to accommodate that somehow if your just linking libraries. I got this error with the latest commit on master. [...]

@jwillikers Thanks for your comment. I'm aware that CMake targets provide include directories as well but the code in question (`OPTIONAL_LIBS`) didn't take care of that. I guess that your build is different because `wayland-client.h` is not in a system location. On my Linux system I see `/usr/include/wayland-client.h` which is obviously in a standard system location and found by the compiler anyway. Some questions:

- Can you tell me more about the situation on your system? For instance, where is "your" `wayland-client.h`?
- Just for my understanding: which package provides the appropriate include directory?

As I wrote, it's only for my understanding. Other users don't seem to suffer from this difference.

You wrote:
> I'd be curious to know how the project is being consumed where those imported targets aren't available, which must be from the corresponding find_package calls not being reached.

All these libraries and include directories in question (`OPTIONAL_LIBS`) are only `PRIVATE` dependencies of FLTK. User projects should not need to know the include directories of Wayland specific headers - unless they do Wayland specific stuff themselves, of course. But that's another story. Therefore user projects don't use - and **_must not_** (need to) use - `find_package()` calls for these (unknown) packages. These are implementation details of the FLTK library.

I wonder why and at what point in your build the wayland headers are not found. I guess you report it because the FLTK library can't be built, or are you seeing this in consumer projects?

An example CMake project can be found in our `README.CMake.txt` (slightly edited):
```
cmake_minimum_required(VERSION 3.15)
project(hello)
# optional: (see docs)
set(FLTK_DIR "/path/to/fltk" CACHE FILEPATH "FLTK installation or build directory")
find_package(FLTK CONFIG REQUIRED)
add_executable       (hello WIN32 MACOSX_BUNDLE hello.cxx)
target_link_libraries(hello PRIVATE fltk::fltk)
```

This demo project worked before your PR, failed after it, and was fixed again with my additional commit. This was reported in [fltk.coredev](https://groups.google.com/g/fltkcoredev/c/Akt9MBIKKxs/m/sRqgwr6sAgAJ) with the following CMake error message:
```
CMake Error at
/home/.../BUILD-Linux-amd64/Release/install/share/fltk/FLTK-Targets.cmake:60
(set_target_properties):
  The link interface of target "fltk::fltk" contains:

    PkgConfig::PANGOCAIRO

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

This bug report was from a user's project but I could also confirm it with the `hello world` project above. Note that the user project doesn't "know" about the usage of Pango or Cairo in the FLTK library and doesn't search for such packages.

This is all only to explain why I had to change your PR in the first place. I don't say that we don't need to add the include headers for the FLTK library build in your use case though. I'm just investigating and I want to understand what's going on.

jwillikers:
@Albrecht-S Yeah, you've got it right that the Wayland includes aren't in the system include directories.

> Can you tell me more about the situation on your system? For instance, where is "your" wayland-client.h?

It's from the Wayland Conan package, which resides in the Conan cache. Conan generates the appropriate pkg-config files for all of its packages in the cache required for fltk and all dependencies and points CMake to this directory.

> Just for my understanding: which package provides the appropriate include directory?

That would be the [Wayland package](https://github.com/conan-io/conan-center-index/blob/master/recipes/wayland/all/conanfile.py).

> I wonder why and at what point in your build the wayland headers are not found. I guess you report it because the FLTK library can't be built, or are you seeing this in consumer projects?

No, this is just building fltk.

> This bug report was from a user's project but I could also confirm it with the hello world project above. Note that the user project doesn't "know" about the usage of Pango or Cairo in the FLTK library and doesn't search for such packages.

Thanks, that tells me exactly what's going on. Consumers are using the `FLTK-Targets.cmake` module to pull in an installed fltk project. The exported targets still link against the imported CMake targets from the `pkg_check_modules` calls, which is why they aren't working. CMake exports are really tricky to get right, and I'm a bit rusty on that particular topic. Things always get really complicated around this. Like, that `FLTK-Targets.cmake` is from the `install(EXPORT` command I think and not the `export(TARGETS` command.

That said, I think things should be pretty trivial to patch up for my use case by just including a couple `target_include_directories` calls to include the include directories for those `PkgConfig::` targets. I think improving the CMake exports is the proper long-term fix, because this ought to allow the project to use imported targets without breaking things for consumers using the library in this way.

Albrecht-S:
> Yeah, you've got it right that the Wayland includes aren't in the system include directories.
> It's from the Wayland Conan package, which resides in the Conan cache.
> That would be the [Wayland package](https://github.com/conan-io/conan-center-index/blob/master/recipes/wayland/all/conanfile.py).

Thanks, I understand "not in the system include directories" and "Conan package, which resides in the Conan cache" but the given link doesn't help me because "I don't speak Conan"  ;-)
Anyway, that's what I wanted to know.

What we know now is:
- we can't use `PkgConfig::*` targets in include and link requirements because consumers don't know these targets
- my previous commit fixed that for linker requirements (libraries)
- we need a similar approach for include directories.

> I think things should be pretty trivial to patch up for my use case by just including a couple target_include_directories calls to include the include directories for those PkgConfig:: targets.

Yup, that's what I'm going to do tomorrow - or ... see below. The crux is that we also need to feed this info into our `fltk-config` script which only understands shell and compiler and linker switches and libraries but that's as trivial as doing it for the CMake part, just a few more lines of code.

> I think improving the CMake exports is the proper long-term fix, because this ought to allow the project to use imported targets without breaking things for consumers using the library in this way.

I'm not sure if I understand what you mean with this statement. It's clear that we can't simply `export` an `imported target` like `PkgConfig::*` (I tried that in a "naïve" first test just to be sure). I could try to create a separate `INTERFACE` library with all the `PkgConfig::*` details (either targets or explicit link libraries and include directories) instead of collecting them in variables like `OPTIONAL_LIBS`. This could simplify our CMake code quite a bit - I'll try this tomorrow as well.

jwillikers:
> I'm not sure if I understand what you mean with this statement. It's clear that we can't simply export an imported target like PkgConfig::* (I tried that in a "naïve" first test just to be sure). I could try to create a separate INTERFACE library with all the PkgConfig::* details (either targets or explicit link libraries and include directories) instead of collecting them in variables like OPTIONAL_LIBS.

I'd need to look into it more to find how other projects accomplish this, but I'd guess that they incorporate the necessary `find_package` / `pkg_check_modules` calls in the CMake config to make the imported targets available. I think this would end up being more portable than using hard-coded paths or assuming everything can be found in system include and library directories.

Albrecht-S:
Commits 89b957586e4ae2343c35284be00d6afa08dc37ca and yet another fix for shared libs in e39c96f419b598c2469efa1bd38c7e644f562df8 should fix all issues related to your Conan build.

@jwillikers Please test again and confirm that it works for you now.

Albrecht-S:
> I'd need to look into it more to find how other projects accomplish this, but I'd guess that they incorporate the necessary find_package / pkg_check_modules calls in the CMake config to make the imported targets available. I think this would end up being more portable than using hard-coded paths or assuming everything can be found in system include and library directories.

@jwillikers I'd be all ears if you found a workable solution other (better) than mine in above mentioned commits.

1. The missing headers (i.e. their search directories) to build FLTK itself are added in my commits. This concerns only building the FLTK library itself.
2. The link libraries found by `pkg-config` are provided by absolute paths to the libs in the FLTK targets in FLTKConfig.cmake or in the related FLTK-Targets file. This seems to be useful (and intended by CMake) because you might otherwise find different shared libs on a system in the consumer's CMake environment.

Regarding "portability": if the Conan build includes "special" shared libs in the Conan build environment then the application linking to FLTK **_must_** IMHO be linked to these specific shared libraries (which is done with the current version). I can see problems if the libs linked to by the FLTK library are moved (installed) and then deleted because absolute paths are used by the linker. I have no idea how this could be solved. Also, deploying built user programs to other systems might be an issue. This is fragile anyway (although I see users doing this) and I don't know how to solve this too. Or maybe this isn't a problem at all, I don't know.

jwillikers:
> Commits [89b9575](https://github.com/fltk/fltk/commit/89b957586e4ae2343c35284be00d6afa08dc37ca) and yet another fix for shared libs in [e39c96f](https://github.com/fltk/fltk/commit/e39c96f419b598c2469efa1bd38c7e644f562df8) should fix all issues related to your Conan build.
> 
> @jwillikers Please test again and confirm that it works for you now.

This works! Thank you!!

> @jwillikers I'd be all ears if you found a workable solution other (better) than mine in above mentioned commits.

I'll certainly let you know if I do!

> The link libraries found by pkg-config are provided as absolute paths to the libs in the FLTK targets in FLTKConfig.cmake or in the related FLTK-Targets file. This seems to be useful (and intended by CMake) because you might otherwise find different shared libs on a system in the consumer's CMake environment.

CMake's `pkg_check_modules` seems to predate its support for imported targets, which is probably why it provides the libraries in a way where you can add them directly to `target_link_libraries` type calls without having to use an additional variable containing the link directories. CMake seems to provide [find_dependency](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) for this purpose, as it's meant to be used in CMake package files, but it's meant to easily work for `find_package` calls and not `pkg_check_modules` calls.

Albrecht-S:
> This works! Thank you!!

Great, thanks for the feedback.

> CMake's pkg_check_modules seems to predate its support for imported targets...

Yup, [CMake docs](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules) say that `IMPORTED_TARGET` has been added in CMake 3.6. But I don't think that this is relevant for our problem.

Let's recap what happened:
1. Your PR added the `IMPORTED_TARGET` argument in an attempt to use the imported target directly in `target_link_libraries()`.
2. Although this worked for your Conan build of the FLTK library itself it broke a simple "Hello World" consumer project.
3. My solution to (2.) was to _extract_ the link libraries and include directories from the imported `PkgConfig::` target and to add them as `target_link_libraries()` and `target_include_directories()` to the target `fltk` (`fltk::fltk`) etc..

In retrospect we could have used the respective PkgConfig **_variables_** directly in the first place but I'm not going to turn this back.

> CMake seems to provide [find_dependency](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) for this purpose, as it's meant to be used in CMake package files, but it's meant to easily work for find_package calls and not pkg_check_modules calls.

Note: our goal is to create one target (`fltk::fltk`) that provides all its dependencies for consumer projects by using the `FLTKConfig.cmake` file. That's how I understand the "modern CMake" approach with CMake config files, and this is what happens now: the target `fltk::fltk` contains all informations needed by the consumer project. System introspection ("finding" dependencies) by the consumer project is not necessary and IMHO not desirable. We do already know all dependencies and if we tried to "find" them in the `FLTKConfig.cmake` file (again) we might even find the wrong versions of some libraries and make the link fail or produce runtime errors.

Therefore I believe that our current solution is good as it is. Thanks for all your help.

jwillikers:
I understand and am very happy with the current solution. Thanks for adding those `target_include_directories` calls for me.

I'm not sure this applies to fltk, but just so you're aware, the only thing that might cause problems is packaging up the `FLTKConfig.cmake` with the fltk library and installing it on a different machine, i.e. building a package on Debian and installing on Fedora. This would have the wrong hard-coded paths to libraries due to the differences between the standard library directories, i.e. `/usr/lib64` and `/usr/lib/<arch>-linux-gnu`.

Albrecht-S:
> I understand and am very happy with the current solution.

:+1: @jwillikers I'm happy that you are happy.

>Thanks for adding those target_include_directories calls for me.

Not only for you. Your report made clear that adding the proper include directories is necessary in general. There may be other systems where some libs and headers can be found in non-system directories.

> the only thing that might cause problems is packaging up the FLTKConfig.cmake with the fltk library and installing it on a different machine, ...

I don't think that this is a common scenario. If you link to shared libs on one system and try to run an application on another system you might have much more important (ABI) incompatibilities than only the search paths...

PR #958, opened 2024-04-24
Solved: STR 2145: Fix box types and focus frames.

MatthiasWM:
[2145](https://www.fltk.org/str.php?L2145+P0+S-2+C0+I0+O0+E0+V1.+Q)

This is a big PR and it needs some explanation. The STR itself asks to fix the issue that the focus frame on some box types is just a rectangle and does not match the graphics of the frame or box. I fixed this by adding a focus-drawing callback to the box type array. That part was a few lines of code.

As a bigger refactoring, I removed all the box types starting with an underline, like `_FL_PLASTIC_UP_BOX` and replaced the no-underscore `FL_PLASTIC_UP_BOX` macros with direct enumeration values. The original code jumped through some really interesting hoops that helped linkers optimize unused code away. If an app never called `_FL_PLASTIC_UP_BOX`, the code to draw the plastic boxes would never be referenced, and a good linker could optimize and remove all plastic drawing routines.

Switch to 2024, and FLTK offers `-scheme` as a command line argument. This means that we always must link all possible drawing code, including the Plastic scheme and all others. So the whole macro setup and dynamic creation of the box type array is obsolete and wastes time and space, and adds complexity.

So I removed all the code that is no longer needed. Comments welcome.

### DIFF
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -591,12 +591,6 @@ enum Fl_Callback_Reason {
 
     This enum defines the standard box types included with FLTK.
 
-   \note The documented \p enum \p Fl_Boxtype contains some values (names)
-      with leading underscores, e.g. \p \b _FL_SHADOW_BOX. This is due to
-      technical reasons - please use the same values (names) without the
-      leading underscore in your code! Enum values with leading underscores
-      are reserved for internal use and subject to change without notice!
-
     FL_NO_BOX means nothing is drawn at all, so whatever is already
     on the screen remains. The FL_..._FRAME types only draw their edges,
     leaving the interior unchanged. The blue color in the image below
@@ -617,12 +611,6 @@ enum Fl_Callback_Reason {
     \brief FLTK standard box types
 
     This enum defines the standard box types included with FLTK.
-
-   \note The documented \p enum \p Fl_Boxtype contains some values (names)
-      with leading underscores, e.g. \p \b _FL_SHADOW_BOX. This is due to
-      technical reasons - please use the same values (names) without the
-      leading underscore in your code! Enum values with leading underscores
-      are reserved for internal use and subject to change without notice!
 */
 enum Fl_Boxtype { // boxtypes (if you change these you must also change fl_boxtype.cxx):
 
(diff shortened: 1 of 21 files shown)


MatthiasWM:
PS: this can be tested nicely by running "test/boxtype"

Albrecht-S:
Is this intended as a "bug fix" for 1.4.0 or is it a "feature request" to be implemented in 1.5.0?

Some thoughts:

I'm always for removing overhead and for simplifying code but I'm afraid that the removal of the enum values with underscore might break working 1.3.x code that used these enum values as box types [1]. Note that the warnings in the docs like "_plastic version of_ `FL_UP_BOX`, _use_ `FL_PLASTIC_UP_BOX`" for enum value `_FL_PLASTIC_UP_BOX` (with underscore) have been added in 1.4.0, they don't exist in 1.3.x.

These changes might also affect users that add their own schemes but since we don't know such user code [2] I'm also reluctant to introduce code changes that **_might_** affect schemes in 1.4.0 (1.4.x).

That said, I support these changes in general but I don't think that the entire "refactoring" is suitable for 1.4.0, it should probably be postponed (1.5.0) although I would like to have a much smaller patch that introduces the new focus box drawing in 1.4.0 (if this could be done in a backwards compatible way).

----
[1] I also considered replacing the names with underscore with pure macros, like
```
#define _FL_PLASTIC_UP_BOX FL_PLASTIC_UP_BOX
```
which might work for 1.3.x backwards compatibility but it might also confuse users and devs. These names would have to be documented as deprecated in 1.4.0 and would have to be removed in 1.5.0, or something like that.

[2] IIRC the [TigerVNC](https://github.com/TigerVNC/tigervnc) project defines an [own scheme](https://github.com/TigerVNC/tigervnc/blob/master/vncviewer/fltk/theme.cxx) or at least they're doing something very similar but there may be other projects as well.

Albrecht-S:
PS: as a test I can imagine to build TigerVNC (vncviewer) from sources with FLTK 1.4.0. The last time I tried this it worked with a few minor fixes (mainly missing headers) but I do also believe that I opened an issue or PR to fix these issues. Hence it would be a good test to see if it works before and after applying this PR.

MatthiasWM:
Thanks @Albrecht-S . I tested the patch with TigerVNC and with the addition of the enums, it compiles fine(*). I will put the additional _FL const inside and `#ifndef DOXYGEN` to avoid cluttering the docs.

Since I can never be 100% sure that the patch has no side effects, I think I will introduce this patch in multiple smaller commits into master. If there is a regression, it will be easier to find. Is that ok with you?

*) The cmake/Makefiles/install sequence generates `/usr/local/FLTK.framework` on my Mac, but the contents of that directory is not a macOS framework at all (directories ending in `.framework` have a special meaning on macOS and bundle headers and libraries together) . But the TigerVNC cmake finds `/usr/local/FLTK.framework` before it finds `/usr/local/lib/libfltk.dylib` and fails to link. Deleting `/usr/local/FLTK.framework` fixes things for TigerVNC.

Albrecht-S:
> I will introduce this patch in multiple smaller commits into master. If there is a regression, it will be easier to find. Is that ok with you?

Sure, go ahead (I noticed that you did already push a commit). The new focus drawing looks really good, thank you very much!

OT:
> The cmake/Makefiles/install sequence generates /usr/local/FLTK.framework on my Mac, but the contents of that directory is not a macOS framework at all ...

Thanks for mentioning it, this is a known problem. Manolo provided some insights and I have this on my todo list. The current code is obviously from an earlier attempt to provide a framework.
/OT

Albrecht-S:
One thing I noticed during quick testing with `test/boxtype.cxx` is that `*_FRAME` (and FL_NO_BOX) boxtypes never restore the look w/o the focus frame - this is not a regression, it's obvious because these boxtypes don't draw a background and has been this way "forever".

We **_may consider_** not to draw a focus box at all on boxtypes w/o a solid background because we can't "erase" it when the widget loses the focus. This should be possible with the new focus box drawing function per boxtype (just set it to an empty function (not NULL, obviously) for `FL_NO_BOX` and `FL_*_FRAME` types. What do you think?

MatthiasWM:
> One thing I noticed during quick testing with `test/boxtype.cxx` is that `*_FRAME` (and FL_NO_BOX) boxtypes never restore the look w/o the focus frame

Yes, I would leave it like that and assume that the users know why they chose a frame over a box. Maybe they draw their own background, maybe the parent does. And they can always disable the focus for every widget with a simple flag.

Albrecht-S:
> they can always disable the focus for every widget with a simple flag.

OK, that makes sense.

Albrecht-S:
@MatthiasWM Can we close this PR? It hasn't been merged but implemented differently.

Is there still anything we need to do, maybe regarding the "underscore" boxtypes, or ... ?

MatthiasWM:
I will close this issue. It's no longer needed. The process is not done yet, but I will likely move it to 1.5. Don't let this hold you back from anything.

PR #965, opened 2024-05-01
use MAC_OS_X_VERSION_MIN_REQUIRED to avoid building error when using …

cstkingkey:
…newer sdk and setting old target

> MAC_OS_X_VERSION_MIN_REQUIRED specifies the earliest OS version you support, and corresponds to MACOSX_DEPLOYMENT_TARGET in your project/ target configuration.

> MAC_OS_X_VERSION_MAX_ALLOWED specifies the maximum OS version your code knows about, and corresponds to SDKROOT in your project/target configuration.

### DIFF
--- a/src/Fl_Native_File_Chooser_MAC.mm
+++ b/src/Fl_Native_File_Chooser_MAC.mm
@@ -28,7 +28,7 @@
 #include <FL/fl_string_functions.h>
 #define MAXFILTERS      80
 #import <Cocoa/Cocoa.h>
-#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_VERSION_11_0
 #  import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
 #endif
 
@@ -532,7 +532,7 @@ static char *prepareMacFilter(int count, const char *filter, char **patterns) {
 - (void)control_allowed_types:(const char *)p
 {
   NSString *ext = [NSString stringWithUTF8String:p];
-#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_VERSION_11_0
   if (fl_mac_os_version >= 110000) {
     UTType *type = [UTType typeWithFilenameExtension:ext]; // 11.0 + framework UniformTypeIdentifiers
     [dialog setAllowedContentTypes:[NSArray arrayWithObject:type]]; // 11.0


Albrecht-S:
Thanks for the PR. I don't know how to use the macOS "version magic" stuff, this seems like something @ManoloFLTK can tell best (maybe @MatthiasWM as well, I'm not sure).

ManoloFLTK:
At this point in time, I don't agree with the proposed change. The macOS-specific code of FLTK is written to allow building with many SDK versions (either old or recent) and to run resulting apps on either recent or old macOS systems. That is visible in the source code
```
#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
  if (fl_mac_os_version >= 110000) {
    UTType *type = [UTType typeWithFilenameExtension:ext]; // 11.0 + framework UniformTypeIdentifiers
    [dialog setAllowedContentTypes:[NSArray arrayWithObject:type]]; // 11.0
  }
  else
#endif
  if (fl_mac_os_version >= 100900) {
    [dialog performSelector:@selector(setAllowedFileTypes:) 
                 withObject:[NSArray arrayWithObject:ext]];
  }
```
Line
`#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0`
ensures that the following new in macOS 11 API is used only if building with an SDK of version 11 or more
and line
`  if (fl_mac_os_version >= 110000) {`
ensures that if an app built using a recent API runs on an system older than macOS 11, the new API is not called.
Consider that `fl_mac_os_version` is computed by FLTK early to indicate the version of the running OS such that
macOS 14.3 is 140300 and macOS 10.13 is 101300.

FLTK has been using this type of construct for years. Is it wrong?

The link error observed when cross-compiling FLTK from Linux to macOS suggests the value of CMAKE_SYSTEM_VERSION when cross-compiling does not match the version of the SDK with which the compilation step is performed.
If a symbol from framework UniformTypeIdentifiers is missing at link time, it means that this symbol was compiled-in and thus that the SDK version used is ≥ 11.0
Then, the FLTK build procedure uses condition CMAKE_SYSTEM_VERSION ≥ 20 to determine whether the link command should contain "`-framework UniformTypeIdentifiers`". I've just verified here that if the FLTK build procedure finds that variable CMAKE_SYSTEM_VERSION has a value ≥ 20.0 (the two numbering systems in use don't facilitate things), cross compilation of FLTK on Linux for macOS with the macOS 14.0 SDK runs well. It fails if that variable is set to 19.0.

In a nutshell, I believe that during your cross-build procedure, the  SDK version is ≥ 11 and CMAKE_SYSTEM_VERSION is < 20. These 2 things are not coherent with each other. The minimum macOS version on which the resulting apps can run is something else determined, if I'm not wrong, by CMAKE_OSX_DEPLOYMENT_TARGET.

cstkingkey:
> At this point in time, I don't agree with the proposed change. The macOS-specific code of FLTK is written to allow building with many SDK versions (either old or recent) and to run resulting apps on either recent or old macOS systems. That is visible in the source code
> 
> ```
> #if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
>   if (fl_mac_os_version >= 110000) {
>     UTType *type = [UTType typeWithFilenameExtension:ext]; // 11.0 + framework UniformTypeIdentifiers
>     [dialog setAllowedContentTypes:[NSArray arrayWithObject:type]]; // 11.0
>   }
>   else
> #endif
>   if (fl_mac_os_version >= 100900) {
>     [dialog performSelector:@selector(setAllowedFileTypes:) 
>                  withObject:[NSArray arrayWithObject:ext]];
>   }
> ```
> 
> Line `#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0` ensures that the following new in macOS 11 API is used only if building with an SDK of version 11 or more and line ` if (fl_mac_os_version >= 110000) {` ensures that if an app built using a recent API runs on an system older than macOS 11, the new API is not called. Consider that `fl_mac_os_version` is computed by FLTK early to indicate the version of the running OS such that macOS 14.3 is 140300 and macOS 10.13 is 101300.
> 
> FLTK has been using this type of construct for years. Is it wrong?
> 
> The link error observed when cross-compiling FLTK from Linux to macOS suggests the value of CMAKE_SYSTEM_VERSION when cross-compiling does not match the version of the SDK with which the compilation step is performed. If a symbol from framework UniformTypeIdentifiers is missing at link time, it means that this symbol was compiled-in and thus that the SDK version used is ≥ 11.0 Then, the FLTK build procedure uses condition CMAKE_SYSTEM_VERSION ≥ 20 to determine whether the link command should contain "`-framework UniformTypeIdentifiers`". I've just verified here that if the FLTK build procedure finds that variable CMAKE_SYSTEM_VERSION has a value ≥ 20.0 (the two numbering systems in use don't facilitate things), cross compilation of FLTK on Linux for macOS with the macOS 14.0 SDK runs well. It fails if that variable is set to 19.0.
> 
> In a nutshell, I believe that during your cross-build procedure, the SDK version is ≥ 11 and CMAKE_SYSTEM_VERSION is < 20. These 2 things are not coherent with each other. The minimum macOS version on which the resulting apps can run is something else determined, if I'm not wrong, by CMAKE_OSX_DEPLOYMENT_TARGET.

Thanks for your explanation.

ManoloFLTK:
> Thanks for your explanation.

You are welcome, but does that mean you now cross-build successfully FLTK?

I've successfully cross built FLTK here from Debian where I installed osxcross for the SDK 14.0 with this command :
```
cmake -B. -S $HOME/fltk -DCMAKE_C_COMPILER=/home/XXX/osxcross/target/bin/x86_64-apple-darwin23-clang -DCMAKE_CXX_COMPILER=/home/XXX/osxcross/target/bin/x86_64-apple-darwin23-clang++ -DCMAKE_SYSTEM_NAME=Darwin -DCMAKE_SYSTEM_VERSION=23 -DFLTK_BUILD_GL=0 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13
```

cstkingkey:
> > Thanks for your explanation.
> 
> You are welcome, but does that mean you now cross-build successfully FLTK?
> 
> I've successfully cross built FLTK here from Debian where I installed osxcross for the SDK 14.0 with this command :
> 
> ```
> cmake -B. -S $HOME/fltk -DCMAKE_C_COMPILER=/home/XXX/osxcross/target/bin/x86_64-apple-darwin23-clang -DCMAKE_CXX_COMPILER=/home/XXX/osxcross/target/bin/x86_64-apple-darwin23-clang++ -DCMAKE_SYSTEM_NAME=Darwin -DCMAKE_SYSTEM_VERSION=23 -DFLTK_BUILD_GL=0 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13
> ```

The issue is that other libraries don't have the same logic. I think the sdk itself use MACOSX_DEPLOYMENT_TARGET to control symbols visiblity.

cstkingkey:
thanks for your time

ManoloFLTK:
@cstkingkey I you have the possibility and the time to try with the new FLTK code after commit 70b84b4, I would appreciate being informed of what the situation is. I have installed osxcross inside a Debian VM so I should be able to reproduce issues if there are remaining ones.

cstkingkey:
Now it works.

Issue #991, opened 2024-06-20
Wayland Memory Leak: one leak (16 bytes) per key press


Albrecht-S:
**Bug Description**
Under Wayland a key repeat timeout causes a memory leak that accumulates per keypress (16 bytes each).

**To Reproduce**
Run any FLTK program under Wayland with valgrind, e.g. `valgrind --leak-check=full test/hello.cxx` or ASAN (see below). Then hit any keys and quit the program. Whether the program uses an input field or not doesn't matter.
Valgrind reports among other (false positive) leaks:
```
==808033== 64 bytes in 4 blocks are definitely lost in loss record 4,799 of 7,783
==808033==    at 0x4840F2F: operator new(unsigned long) (vg_replace_malloc.c:422)
==808033==    by 0x1A121B: wl_keyboard_key(void*, wl_keyboard*, unsigned int, unsigned int, unsigned int, unsigned int) (in /git/fltk/master/test/hello)
==808033==    by 0x5BD1F79: ??? (in /usr/lib/x86_64-linux-gnu/libffi.so.8.1.2)
==808033==    by 0x5BD140D: ??? (in /usr/lib/x86_64-linux-gnu/libffi.so.8.1.2)
==808033==    by 0x5BD1B0C: ffi_call (in /usr/lib/x86_64-linux-gnu/libffi.so.8.1.2)
==808033==    by 0x488B760: ??? (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==808033==    by 0x4887AA9: ??? (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==808033==    by 0x488941B: wl_display_dispatch_queue_pending (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==808033==    by 0x19FFAA: wayland_socket_callback(int, wl_display*) (in /git/fltk/master/test/hello)
==808033==    by 0x19F8B0: Fl_Unix_Screen_Driver::poll_or_select_with_delay(double) (in /git/fltk/master/test/hello)
==808033==    by 0x19ED94: Fl_Unix_System_Driver::wait(double) (in /git/fltk/master/test/hello)
==808033==    by 0x1644D6: Fl::run() (in /git/fltk/master/test/hello)
```

[...]

PR #992, opened 2024-06-21
Add Fl::remove_next_timeout(...) to remove only one timeout

Albrecht-S:
This method also returns the data pointer of the removed timeout.

This new method is generally useful for user code and particularly necessary to fix issue #991.


Albrecht-S:
This PR is for documentation only. I'm going to merge it now...

PR #995, opened 2024-06-27
Fl_Tree +/- buttons are no longer drawn using bitmaps, different default line color

### LINKED ISSUE #972: Fl_Tree doesn't use system colors
**Describe the bug**
Fl_Tree doesn't follow the scheme colors. The only correctly set color is the tree item selection color, others are not set. This can be seen best when the system uses a dark color scheme. On Windows, the FLTK colors are sourced from the classic theme colors (deprecated after Win7 but still functional as of Win11). On Linux the FLTK colors are sourced using some (presumably deprecated too) X mechanism, but it only works in the TDE and MATE desktop envinroments. Moreover, the line color appears to be hard coded, and the expando buttons are drawn from a XPM bitmap.

**To Reproduce**
Considering that this behaviour can't be (visibly) reproduced on modern windows without modding the system substantially, it's best to test it on a Win7 virtual machine.
1. On a win7 VM, select the classic theme and apply some crazy dark colors.
2. Run the demo tree program.

**Expected behavior**
The Fl_Tree should set all the colors correctly, the line color should be derived from one of the FLTK colors, and the expando buttons need to follow the color scheme too.

**Screenshots**
[imgur](https://i.imgur.com/2It2JX5.png)

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git
 - If from Git, branch: master
 
All default CMake options, sans the disabled FLTK_BUILD_GL due to some unrelated problems. The problem reported persists regardless of this setting anyways.

 - OS: Windows, Linux
 - OS Version: Windows 10 / Arch Linux with TDE

**Linux/Unix Runtime, if applicable:**
 - X11

**Additional context**
Two things:
- it's still possible to use the classic Win2000 theme on modern versions of windows, but it requires the use of third party tools.
- The way FLTK gets system colors on Linux/X11 doesn't work on any modern desktop envinroment (that I tried). I've only got it to work on TDE and Mate, which are forks of KDE3 and Gnome 2 respectively. To fix this, I think it's necessary to detect what desktop envinroment is running, and read the colors from the DE config files. I've done this for the LXQT desktop envinroment as a proof of concept and it works (still no success with treeview colors, though).

CyprinusCarpio:
The XPM bitmaps for default Tree buttons were removed, instead the buttons are drawn using FLTK functions when `closeicon()` or `openicon()` return NULL, which is now the case unless button bitmaps are explicitly set. Fixes #972

### DIFF
--- a/src/Fl_Tree_Prefs.cxx
+++ b/src/Fl_Tree_Prefs.cxx
@@ -35,7 +35,7 @@
 
 void Fl_System_Driver::tree_draw_expando_button(int x, int y, bool state, bool active) {
   fl_rectf(x, y, 11, 11, active ? FL_BACKGROUND2_COLOR : fl_inactive(FL_BACKGROUND2_COLOR));
-  fl_rect(x, y, 11, 11, Fl_Color(8));
+  fl_rect(x, y, 11, 11, FL_INACTIVE_COLOR);
   fl_color(active ? FL_FOREGROUND_COLOR : fl_inactive(FL_FOREGROUND_COLOR));
   fl_line(x + 2, y + 5, x + 8, y + 5);
   if (state) {
@@ -102,7 +102,7 @@ Fl_Tree_Prefs::Fl_Tree_Prefs() {
   _linespacing            = 0;
   _labelfgcolor           = FL_FOREGROUND_COLOR;
   _labelbgcolor           = 0xffffffff;         // we use this as 'transparent'
-  _connectorcolor         = Fl_Color(8);
+  _connectorcolor         = FL_INACTIVE_COLOR;
   _connectorstyle         = (Fl_Tree_Connector)Fl::system_driver()->tree_connector_style();
   _openimage              = 0;
   _closeimage             = 0;
--- a/src/Fl_Tree_Prefs.cxx
+++ b/src/Fl_Tree_Prefs.cxx
@@ -37,7 +37,7 @@
 //    │ + │   │ - │
 //    └───┘   └───┘
 void Fl_System_Driver::tree_draw_expando_button(int x, int y, bool state, bool active) {
-  fl_rectf(x, y, 11, 11, FL_BACKGROUND2_COLOR);         // fill
+  fl_rectf(x, y, 11, 11, active ? FL_BACKGROUND2_COLOR : fl_inactive(FL_BACKGROUND2_COLOR)); // fill
   fl_rect(x, y, 11, 11, FL_INACTIVE_COLOR);             // outline
   fl_color(active ? FL_FOREGROUND_COLOR : FL_INACTIVE_COLOR);
   fl_line(x + 3, y + 5, x + 7, y + 5);                  // horiz line
--- a/src/Fl_Tree_Prefs.cxx
+++ b/src/Fl_Tree_Prefs.cxx
@@ -34,10 +34,9 @@
 
 
 void Fl_System_Driver::tree_draw_expando_button(int x, int y, bool state, bool active) {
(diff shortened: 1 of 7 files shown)


Albrecht-S:
Thanks for the PR. This is something Greg (@erco77) should review because he knows best what to do. I requested his review.

Albrecht-S:
@CyprinusCarpio Please see also my [comment here](https://github.com/fltk/fltk/issues/972#issuecomment-2198124861) about numerical vs. named colors.

erco77:
Seems like the approach I was thinking of; I'll get in deep to double check for things like null derefs, which is the biggest concern for changes like this, and will do the merge if all looks ok.

erco77:
OK, looks good -- reviewed and tested on Mac + Linux so far.

Pushed a few small mods as 3938e6d, specifically:

- Small fixes for the mac applied to Fl_Darwin_System_Driver.cxx:

  - Added include for fl_draw.H
  - Changed `Fl_System_Driver::tree_draw_expando_button(..)`
  to instead `Fl_Darwin_System_Driver::tree_draw_expando_button(..)` 

- Added new methods to access the icon's w() and h() which do the NULL deref checks, getting that logic out of the inline code.

- One pixel mods to the "⊞" and "⊟" symbols to more closely match the original bitmaps, removed an unused call to fl_color(), small code formatting for fewer lines.

erco77 (approved):
See above PR comments -- approved with patch 3938e6d applied.

src/Fl_Tree_Prefs.cxx
```diff
@@ -22,74 +22,26 @@
...
-
-Fl_Pixmap *Fl_System_Driver::tree_closepixmap() {
-  static Fl_Pixmap *pixmap = new Fl_Pixmap(tree_close_xpm);
-  return pixmap;
+// Draw non-OS specific Fl_Tree open/close icons
+//    ┌───┐   ┌───┐
+//    │ + │   │ - │
+//    └───┘   └───┘
+void Fl_System_Driver::tree_draw_expando_button(int x, int y, bool state, bool active) {
+  fl_rectf(x, y, 11, 11, FL_BACKGROUND2_COLOR);         // fill
```
CyprinusCarpio, line 40: Removing the ternary for the fill color is not a good change, it messes up the color when the tree is disabled and the system uses a colorful theme (before/after):
bez tytułu
I'll revert that change and push.

Albrecht-S:
> looks OK to me to squash/merge this PR. Albrecht; can you handle that? I still have apprehension with handling PR's, can't remember if squash/merge or rebase/merge is the way to go.

@erco77 after you wrote the above note there have been a few more commits, hence I hesitate to do the merge.

However, it's pretty easy to find out which one to choose: click on the "Squash and merge" button's pulldown triangle and you'll see something like this:
image

In this screenshot I moved the cursor over the "right" selection which colored it in blue.

It's as simple as reading the comments to each selection to find out which one we want:

**_Squash and merge:_** "the <n> commits from this branch will be **_combined into one commit in the base branch_**"

... whereas "Rebase and merge" would copy (rebase) every commit, i.e. create 6 distinct commits (as of now) which is usually not what we want.

I'm leaving this to you to review the latest commit(s) and then I assume you can do the `Squash and merge` yourself. If you're not yet sure, please let me know.

erco77:
@Albrecht-S Thanks, done.
I seem to recall a recent PR where it was messy because there wasn't a branch or something up with the OP's fork, and I think this one was commits without a branch too, so wasn't sure. Anyway it squash/merged OK, and the build checks are running.

Issue #1001, opened 2024-07-01
Regression: CMake error when generating a VS project with default settings


CyprinusCarpio:
**Describe the bug**
CMake errors during generation of a VS 2022 project with default settings since a6651e10f. The error pertains to the GLU include directory, which is not found. Using CMake gui, the problem can be circumvented by setting the variable `OPENGL_GLU_INCLUDE_DIR` to empty, and clicking Generate again, which yields a fully functional FLTK project, GLU and all. CMake log attached.
[cmake log.txt](https://github.com/user-attachments/files/16056034/cmake.log.txt)

**To Reproduce**
1. Clone the repo and checkout a6651e10f
2. Using CMake GUI, try configuring and generating a VS 2022 project.
3. Observe that the error goes away once `OPENGL_GLU_INCLUDE_DIR` is set to empty, and the library can be fully built without problems
4. Likewise, checkout 5de880a and observe that there is no error to begin with.

**Expected behavior**
No error.

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git
 - Branch master
 - Commit a6651e10f

**FLTK Configure / Build Options**
Default

**Operating System / Platform:**
 - OS: Windows 10

Albrecht-S:
Thanks for the report, I'll check it...

Albrecht-S:
Checked with
- Windows 10
- VS 2019
- CMake 3.29.2
- Git commit 1c6a0c1a8fb254945179276ecbd359cc1926fbb1 from Jun 29, 2024 (because that was a commit I had checked out on my VM).

I can't confirm the issue, the build works w/o issues, and `glpuzzle` and `fractals` work as expected.
Result of CMake (in cache):
`OPENGL_GLU_INCLUDE_DIR:PATH=OPENGL_GLU_INCLUDE_DIR-NOTFOUND`.

We need more info:

- Did you make sure that the cache was clear(ed)? If not, please try again in a fresh build folder.
- What is the contents of `OPENGL_GLU_INCLUDE_DIR` before you set it "to empty"?
- Which CMake version?
- Please post full CMake errors if the issue remains.
- Please let us know **exactly** how you run CMake (particularly if you use the included version of VS 2022)

Albrecht-S:
Sorry, I missed that you had included the CMake log.
There is an obvious error:
```
CMake Error in src/CMakeLists.txt:
  Target "fltk_gl" INTERFACE_INCLUDE_DIRECTORIES property contains path:

[...]

PR #1002, opened 2024-07-01
CMake: fix error #1001

CyprinusCarpio:
For some reason `OPENGL_INCLUDE_DIRS` may contain the value `OPENGL_GLU_INCLUDE_DIR-NOTFOUND`, which I checked to be the case using some good old `message` in `src/CMakeLists.txt` in the if statement at line 843. 
I added some calls to print out the included directories in the target `fltk_gl` before and after the line 846, which appears to be the culprit (although it seems that it's only a symptom of a deeper problem). It's clear as day that the `OPENGL_GLU_INCLUDE_DIR-NOTFOUND` path is added by this line. To "fix" this I added some if statements to check for the validity of the values, as per Albrecht's instruction.

This isn't a fix per se obviously, it's only checking for and correcting a so far unexplainable scenario. Again I'm no CMake expert but I don't think this patch has the potential to mess up anything, but please advise nonetheless.


Albrecht-S:
This PR **may** be considered eventually but **_should not_** (yet) be merged as-is, as long we don't know what's really causing the issue on the OP's system.

Please refer to issue #1001 for details and further investigation.

CyprinusCarpio:
Closing as hackfix, now that a proper solution is possible.

PR #1007, opened 2024-07-07
gl_draw: fix drawing of text outside of viewport

### LINKED ISSUE #1006: Weird display of text outside of a viewport in a Fl_GL_Window
**Describe the bug**
Text drawn using `gl_draw` in a GL window behaves weirdly when it's screen origin coordinate falls outside of the viewport - instead of just being culled, it's drawn on top of one of the previously drawn texts that are inside the viewport.
I can reproduce this issue both on my Win10 host and on Arch Linux guest in VBox.

Minimal example program: 
[gltext.txt](https://github.com/user-attachments/files/16116862/gltext.txt)

**To Reproduce**
1. Run the example program
2. Pan the viewport around using the right mouse button and scale it using the wheel until the problem is observed

**Expected behavior**
Text ceases to be drawn if it's outside of the viewport.

**Screenshots**
bez tytułu

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git master b551644

On Windows, `FLTK_MSVC_RUNTIME_DLL` disabled. On Linux, all default.

 - OS: Windows 10, Arch Linux 

**Linux/Unix Runtime, if applicable:**
 - X11

**Additional context**
I'm pretty sure that it's a regression from 1.3.9, but I don't know what commit is guilty. I first observed this when switching my program to FLTK 1.4.0 from git several weeks ago. In my program I just do some maths to ensure that the `gl_text` isn't called when the text origin is outside of a viewport, as to circumvent this problem.

CyprinusCarpio:
The function responsible for binding and displaying the prepared text texture is `gl_texture_fifo::display_texture`, but it doesn't check for the validity (inside the viewport in this case) of the raster position `GL_CURRENT_RASTER_POSITION_VALID`. When it is not valid, then drawing the raster results in undefined behaviour:

> The object coordinates presented by glRasterPos are treated just like those of a glVertex command: They are transformed by the current modelview and projection matrices and passed to the clipping stage. If the vertex is not culled, then it is projected and scaled to window coordinates, which become the new current raster position, and the GL_CURRENT_RASTER_POSITION_VALID flag is set. If the vertex is culled, then the valid bit is cleared and the current raster position and associated color and texture coordinates are undefined.

https://linux.die.net/man/3/glrasterpos

This PR adds a check for validity of the raster pos. If it's not valid, return before doing anything. This fixes #1006  for me.

### DIFF
--- a/src/gl_draw.cxx
+++ b/src/gl_draw.cxx
@@ -510,6 +510,11 @@ void Fl_Gl_Window_Driver::draw_string_legacy(const char* str, int n)
 /** draws a utf8 string using an OpenGL texture */
 void Fl_Gl_Window_Driver::draw_string_with_texture(const char* str, int n)
 {
+  // Check if the raster pos is valid.
+  // If it's not, then drawing it results in undefined behaviour (#1006, #1007).
+  GLint valid;
+  glGetIntegerv(GL_CURRENT_RASTER_POSITION_VALID, &valid);
+  if (!valid) return;
   Fl_Gl_Window *gwin = Fl_Window::current()->as_gl_window();
   gl_scale = (gwin ? gwin->pixels_per_unit() : 1);
   if (!gl_fifo) gl_fifo = new gl_texture_fifo();


src/gl_draw.cxx
```diff
@@ -340,6 +340,11 @@ void gl_texture_fifo::display_texture(int rank)
   // GL_ENABLE_BIT for GL_DEPTH_TEST, GL_LIGHTING
   // GL_TEXTURE_BIT for GL_TEXTURE_RECTANGLE_ARB
   // GL_COLOR_BUFFER_BIT for GL_BLEND and glBlendFunc,
+  // check if the raster pos is valid.
```
Albrecht-S, line 343: 1. If this patch **would** be correct (see below) it should be moved up **above** all the preceding comments which are related to the following statement.
2. However, since `display_texture` is called only once it should IMHO be moved "up" to the callee, i.e. at the beginning of `Fl_Gl_Window_Driver::draw_string_with_texture` before the texture is computed at all. This would avoid to compute textures that are never drawn.
3. See more comments in the "conversation" tab.

Albrecht-S:
@CyprinusCarpio Thanks for the PR, I can confirm that it "works" with the new demo program [posted in this comment](https://github.com/fltk/fltk/issues/1006#issuecomment-2212507380).

However, this fixes only a part of the issue.
1. Text that starts right of (i.e. `x > w`) or below (i.e. `y > h`) the viewport is correctly skipped (not drawn at all) **[good]**
2. Text that is positioned inside the viewport and goes beyond the viewport at the right edge is correctly clipped (as before) **[good]**
3. If the text starts "left of" the viewport (x < 0) and would **_extend into the viewport_** it is not drawn at all **[bad]**

Point (3) is an improvement compared to the status quo because it fixes drawing the text at another, undefined position inside the viewport. See more on this in a follow-up comment on issue #1006.

CyprinusCarpio:
The text disappearing if it's origin is left or bottom of the viewport is the behaviour in 1.3.9 and can be considered a base line for the issue at hand.

So the options are as follows:
1. Merge this PR, with the validation moved to `Fl_Gl_Window_Driver::draw_string_with_texture`, which fixes the problem just as well while avoiding some unnecesary overhead. This will bring the effects of `gl_draw` in line to what it was in 1.3.9.
2. Close this PR, and implement the text culling in such a way that only the part intersecting with the viewport of the text is drawn. If there's no intersection, skip the text altogether. There are a couple of considerations to this. Should the raster be computed in full regardless of how much of it will be visible?
3. There might be some OpenGL mechanism unknown to me that may allow setting the raster pos outside of the viewport and have it be valid, and I'm trying to change the piston rings thru the exhaust.

Albrecht-S:
@CyprinusCarpio Thanks for updating the PR as suggested.

I would be fine with option 1. above (merge as-is) if someone else with more OpenGL knowledge than me (@MatthiasWM, @ManoloFLTK, maybe?) can verify that this patch (PR) is correct in the sense that it doesn't break anything else.

Options 2. and 3. might be subject of another issue or PR, I wouldn't care for now.

ManoloFLTK:
@CyprinusCarpio Please, change the comment lines included in this PR to:
```
  // Check if the raster pos is valid.
  // If it's not, then drawing it results in undefined behaviour (#1006, #1007).
```
and squash this PR into a single commit. After that, it will be merged into the FLTK source code.

ManoloFLTK (changes requested):
see my comment in conversation

ManoloFLTK (approved):

Issue #1041, opened 2024-08-16
[Bug] Fl_Sys_Menu_Bar: Clicking on submenu sometimes deadlocks


dannye:
**Describe the bug**

On Mac, clicking a submenu in the Fl_Sys_Menu_Bar can sometimes cause the menu to get stuck open and the whole program remains indefinitely deadlocked.

**To Reproduce**

1. Compile the current master branch of FLTK like normal
2. Run ./bin/test/menubar
3. In the Sys menubar, click File -> Submenus
4. Repeat Step 3 until the File menu magically reopens and observe that the program is deadlocked

**Expected behavior**
The File menu should not automatically reopen and the program should not deadlock.

**Video**

The deadlock occurs at 1:10 in the video (it takes a lot of clicks since it is random).

https://github.com/user-attachments/assets/bbc349cd-b8d4-44d7-968e-a3fd47be17b2

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git
 - If from Git, branch: master
 - If from Git, commit: afc2072

**FLTK Configure / Build Options**
 - cmake -D CMAKE_BUILD_TYPE=Debug

**Operating System / Platform:**
 - OS: macOS
 - OS Version: macOS 14.5
 - Processor if relevant: M3 Pro

**Additional context**

Sometimes, the Edit or Checkbox menu will unexpectedly open instead of the File menu (as seen a few times in the video), but the program is not deadlocked in this case. Clicking into the top of the Sys menu bar will get the program unstuck.

Albrecht-S:
What happens if you hit `Esc` when it's deadlocked?

MatthiasWM:
Tried for a while to replicate the behavior, but I can't.  Xcode 15.4, macOS 14.5, M1 . Do you have any other way to trigger this? Mostly, the System Menu Bar is handled by macOS. If I could get it to hang, I could at least trace back to the cause.

Can you launch the app with `lldb`, break it when it hangs, and get a backtrace (`bt`)?

dannye:
> What happens if you hit `Esc` when it's deadlocked?

@Albrecht-S Nothing happens when I hit Esc, or Cmd+Q.

I can Cmd+Tab to another application, and Tab back, but the window never repaints itself; only the frozen menu bar is re-shown when tabbing back.

> Do you have any other way to trigger this?

[...]

PR #1043, opened 2024-08-20
Submenus with no callback enter their submenu instead of closing the whole menu

dannye:
Based on discussion in #1041

This PR includes the following changes:

1. Keep the menu open if a submenu with no callback is clicked.
   - This change was applied to both the FLTK menubar and the Mac sys menubar.
3. Enter the submenu if Enter or Space is hit on a submenu with no callback.
   - This change was only applied to the FLTK menubar. This is already the behavior of the Mac sys menubar.

These modifications are not applied to top-level menubar menu items, so Enter, Space, and mouse click still toggle the menu for top-level items.

I think this is more user-friendly, and is pretty common among menu systems, including the Mac sys menubar.

Apply this patch to test/menubar.cxx to test out a submenu _with_ a callback (can easily be tested with both the FLTK menubar and the Mac sys menubar):

```diff
diff --git a/test/menubar.cxx b/test/menubar.cxx
index a0424732d..fba4d6aa5 100644
--- a/test/menubar.cxx
+++ b/test/menubar.cxx
@@ -56,6 +56,10 @@ void test_cb(Fl_Widget* w, void*) {
     G_tty->printf("%s\n", m->label());
 }
 
+void test_cb2(Fl_Widget* w, void*) {
+  G_tty->printf("Submenu test\n");
+}
+
 void quit_cb(Fl_Widget*, void*) {exit(0);}
 
 Fl_Menu_Item hugemenu[100];
@@ -84,7 +88,7 @@ Fl_Menu_Item menutable[] = {
     {"shortcut",FL_ALT+FL_SHIFT+FL_F+1},
     {"shortcut",FL_ALT+FL_CTRL+FL_F+1},
     {"shortcut",FL_ALT+FL_SHIFT+FL_CTRL+FL_F+1, 0,0, FL_MENU_DIVIDER},
-    {"&Submenus", FL_ALT+'S',   0, (void*)"Submenu1", FL_SUBMENU},
+    {"&Submenus", FL_ALT+'S',   test_cb2, (void*)"Submenu1", FL_SUBMENU},
       {"A very long menu item"},
       {"&submenu",FL_CTRL+'S',  0, (void*)"submenu2", FL_SUBMENU},
         {"item 1"},
```

Feel free to defer this until after 1.4.0, if the implications of this change need to be ironed out first. But since this change isn't backwards compatible, strictly speaking, it would be useful to have this change in 1.4.0 rather than 1.4.x.
I think the "risk" for existing 1.3-based applications is very small, but not non-existent.

### DIFF
--- a/src/Fl_MacOS_Sys_Menu_Bar.mm
+++ b/src/Fl_MacOS_Sys_Menu_Bar.mm
@@ -378,7 +378,9 @@ static void createSubMenu( NSMenu *mh, pFl_Menu_Item &mm,  const Fl_Menu_Item *m
       mm = mm->next(0);
       continue;
     }
-    miCnt = [FLMenuItem addNewItem:mm menu:submenu action:selector];
+    miCnt = [FLMenuItem addNewItem:mm menu:submenu
+                            action:( (mm->flags & (FL_SUBMENU+FL_SUBMENU_POINTER) && !mm->callback()) ? nil : selector)
+            ];
     setMenuFlags( submenu, miCnt, mm );
     setMenuShortcut( submenu, miCnt, mm );
     if (mitem && (mm->flags & FL_MENU_INACTIVE || mitem->flags & FL_MENU_INACTIVE)) {
--- a/src/Fl_Menu.cxx
+++ b/src/Fl_Menu.cxx
@@ -821,6 +821,7 @@ int menuwindow::handle_part1(int e) {
       }
       return 1;
     case FL_Right:
+    RIGHT:
       if (pp.menubar && (pp.menu_number<=0 || (pp.menu_number == pp.nummenus-1)))
         forward(0);
       else if (pp.menu_number < pp.nummenus-1) forward(pp.menu_number+1);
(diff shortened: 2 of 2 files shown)


Albrecht-S:
Thanks for this PR, it's very much appreciated. However, I'm not the person to look into this, other FLTK devs know more about menu handling.

> Feel free to defer this until after 1.4.0, if the implications of this change need to be ironed out first. But since this change isn't backwards compatible, strictly speaking, it would be useful to have this change in 1.4.0 rather than 1.4.x.

Actually I thought more about 1.5.0 rather than 1.4.x (x > 0) for this PR that changes behavior of the UI. Our goal with 1.4.x is to stay as close as possible to 1.3.x, and this includes behavioral changes of the UI. But that's only my thought (w/o testing), others may find this change tolerable...

MatthiasWM:
I like the change in behavior for the menu. It makes a lot of sense and duplicates the behavior of macOS menus. For example, pressing enter on a submenu item now opens the submenu instead of dismissing the entire menu. Makes sense to me. Esc for dismissing the menu still works.

I would apply the patch.

Albrecht-S:
> I would apply the patch.

@MatthiasWM IMHO you are the one who knows menu code and behavior best, hence I'd say if you think it's good to go, please feel free to merge it. I wouldn't mind...

MatthiasWM:
Verified. Seems to be working nicely.

PR #1068, opened 2024-09-12
Add support for forward and back mouse buttons on macOS and Windows

CendioHalim:
I've been working on adding forward and back button support for TigerVNC.

This patch will add support for these mouse buttons on macOS and Windows. On X11, these buttons map to 8 and 9.

I also have backports to branch 1.3: https://github.com/CendioHalim/fltk/commit/971e25416479ac33966eea2561a42096313de149 https://github.com/CendioHalim/fltk/commit/0cff6ec1519a387991c2adedaaf57aace7db03d8

Tested this with TigerVNC on Windows 10, 11, and macOS 14.5

### DIFF
--- a/src/Fl_cocoa.mm
+++ b/src/Fl_cocoa.mm
@@ -1046,7 +1046,7 @@ static void cocoaMagnifyHandler(NSEvent *theEvent)
  */
 static void cocoaMouseHandler(NSEvent *theEvent)
 {
-  static int keysym[] = { 0, FL_Button+1, FL_Button+3, FL_Button+2 };
+  static int keysym[] = { 0, FL_Button+1, FL_Button+3, FL_Button+2, FL_Button+8, FL_Button+9 };
   static int px, py;
 
   fl_lock_function();
--- a/src/Fl_win32.cxx
+++ b/src/Fl_win32.cxx
@@ -1347,6 +1347,21 @@ static LRESULT CALLBACK WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lPar
       case WM_RBUTTONUP:
         mouse_event(window, 2, 3, wParam, lParam);
         return 0;
+      case WM_XBUTTONDOWN: {
+        int xbutton = GET_XBUTTON_WPARAM(wParam) == XBUTTON1 ? 8 : 9;
+        mouse_event(window, 0, xbutton, wParam, lParam);
+        return 0;
+      }
+      case WM_XBUTTONDBLCLK: {
+        int xbutton = GET_XBUTTON_WPARAM(wParam) == XBUTTON1 ? 8 : 9;
+        mouse_event(window, 1, xbutton, wParam, lParam);
+        return 0;
+      }
+      case WM_XBUTTONUP: {
+        int xbutton = GET_XBUTTON_WPARAM(wParam) == XBUTTON1 ? 8 : 9;
+        mouse_event(window, 2, xbutton, wParam, lParam);
+        return 0;
+      }
 
       case WM_MOUSEMOVE:
 #ifdef USE_TRACK_MOUSE


MatthiasWM:
Thanks. I will hopefully have time to look at your PR in the next three days.

Albrecht-S:
@CendioHalim Thanks for this PR. You wrote:

> I've been working on adding forward and back button support [for TigerVNC].

I see your patches but I'd like to know what exactly "forward and back button support" is, i.e. what the user has to do to trigger these events, if necessary per platform, e.g. (just a guess) "on macOS press key `Command` + click mouse button 1/2 etc.". We'd also need to know if this is standard behavior on all our supported platforms.

Or, are these the extra "side buttons" some mice provide?

Thanks for any insights.

Albrecht-S:
@MatthiasWM I looked at the patch (proposed changes) and I believe it's probably not complete. The functionality may be enough but I believe we need to define (more) status bits for these extra buttons. However, this is only a quick guess, I'm leaving details up to you. Thanks for looking into it.

CendioHalim:
> @CendioHalim Thanks for this PR. You wrote:
> 
> > I've been working on adding forward and back button support [for TigerVNC].
> 
> I see your patches but I'd like to know what exactly "forward and back button support" is, i.e. what the user has to do to trigger these events, if necessary per platform, e.g. (just a guess) "on macOS press key `Command` + click mouse button 1/2 etc.". We'd also need to know if this is standard behavior on all our supported platforms.
> 
> Or, are these the extra "side buttons" some mice provide?
> 
> Thanks for any insights.

Indeed, these are the side buttons that mice provide. From my testing, (testing three different mice), the side buttons would trigger the same events on macOS and Windows.

CendioHalim:
> @MatthiasWM I looked at the patch (proposed changes) and I believe it's probably not complete. The functionality may be enough but I believe we need to define (more) status bits for these extra buttons. However, this is only a quick guess, I'm leaving details up to you. Thanks for looking into it.

I thought about adding the state bits for the other buttons, but I noticed that X11 only keeps the state for the standard three buttons (left, middle, right). This is reflected `Fl_x.cxx` as well: https://github.com/fltk/fltk/blob/bb917628ffa8ffc6cafc79da80344c8c9c653969/src/Fl_x.cxx#L1009

Keeping states for additional mouse buttons is not something that X11 does AFAIK, so this is something that could be added for all platforms as well.

Albrecht-S:
@CendioHalim Thanks for this quick additional info, it's very much appreciated.

As I wrote before, I'm leaving this PR to Matthias (or any other dev that wants to take it).

MatthiasWM:
Apologies for not getting to the patch in the last days. I will be traveling until Oct 2nd without a capable developer setup, but I am positive, we will get the patch in before RC1. Again, sorry for the delay.

CendioHalim:
No worries, safe travels!

Albrecht-S:
@MatthiasWM After looking at the proposed changes of this PR and #1076 I'm not that confident that we can or should merge this PR in FLTK 1.4.0 because the changes might be error prone and need to be done well on all platforms. This PR deals with macOS and Windows only whereas #1076 deals with X11 and is obviously incomplete (see [my comment](https://github.com/fltk/fltk/issues/1076#issuecomment-2374823366)).

Wayland is (AFAICT) not considered at all (yet).

Documentation would have to be added as well but we really need to understand what's going on and I'd be glad if we could implement **_documented behavior of the X11 platform_** rather than just **testing** what a current implementation does.

That said, I also wrote already that I ordered a mouse for testing (should arrive tomorrow) but I hesitate to do all this in FLTK 1.4.0 for the mentioned reasons. It appears to be too easy to do it wrong. I'm willing to try and test though...

CendioHalim:
> Documentation would have to be added as well but we really need to understand what's going on and I'd be glad if we could implement documented behavior of the X11 platform rather than just testing what a current implementation does.

When I started implementing this, I couldn't find any official documentation from X11, Windows or Apple that made the connection between mouse side buttons and back/forward navigation.

After my own testing, I had a look at how Firefox, Chromium and XQuartz handle this.

This is what Firefox does on:
  1. [macOS](https://searchfox.org/mozilla-central/source/widget/cocoa/nsCocoaUtils.mm#1124)
     * button 3 and 4 get mapped to `MouseButton::eX(1/2)`, which are used to navigate back/forward.
  2. [Windows](https://searchfox.org/mozilla-central/source/widget/windows/nsWindow.cpp)
      * `XBUTTON1` and `XBUTTON2` get mapped to `MouseButton::eX(1/2)`.

In Chromium, they seem to [map the constants](https://source.chromium.org/chromium/chromium/src/+/main:ui/events/ozone/evdev/event_factory_evdev.cc) `BTN_SIDE, BTN_BACK, BTN_FORWARD, BTN_EXTRA` :
```cc
  switch (button) {
    case BTN_LEFT:
      modifier = MODIFIER_LEFT_MOUSE_BUTTON;
      break;
    case BTN_RIGHT:
      modifier = MODIFIER_RIGHT_MOUSE_BUTTON;
      break;
    case BTN_MIDDLE:
      modifier = MODIFIER_MIDDLE_MOUSE_BUTTON;
      break;
    case BTN_BACK:
    case BTN_SIDE:
      modifier = MODIFIER_BACK_MOUSE_BUTTON;
      break;
    case BTN_FORWARD:
    case BTN_EXTRA:
      modifier = MODIFIER_FORWARD_MOUSE_BUTTON;
      break;
    default:
      return;
    }
  }
```
which are defined in the [Linux kernel](https://github.com/torvalds/linux/blob/11a299a7933e03c83818b431e6a1c53ad387423d/include/uapi/linux/input-event-codes.h#L359C1-L363C24): 
```c
#define BTN_SIDE		0x113
#define BTN_EXTRA		0x114
#define BTN_FORWARD		0x115
#define BTN_BACK		0x116
```
It seems like X11 tries to [match these](https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/hw/xfree86/drivers/inputtest/xf86-input-inputtest.c), but I haven't found where the mapping from the physical button is done (USB HID -> ... -> X11).

I found [this commit](https://gitlab.freedesktop.org/xorg/xserver/-/commit/4f27d1e05f67eb8953a9bbed7e6f34b03456c64f) in XQuartz which I thought was interesting. They simply map macOS buttons 4 and higher to 8 and higher on X11.

> I'm not a fan of "experimental programming" (TM).

Fair enough, me neither.

Albrecht-S:
**_Progress report:_**

- @MatthiasWM: I assigned myself as well because I started working on this PR to see how it worked out (for my own curiosity)
- @CendioHalim: I started with your PR but modified it in some aspects, see below for details
- I didn't want to "hijack" your PR, hence I created [my own branch 'mouse-buttons' on my fork](https://github.com/Albrecht-S/fltk/tree/mouse-buttons). I hope you don't mind.
- My branch is still work in progress but it seems to work well for Windows and macOS ! :-)
- X11 and Wayland support are still to be done.
- I updated the docs, but this may be incomplete and need more work.
- **_Everything is only preliminary and may or may not be included in FLTK 1.4.0 (depends on further progress)._**

**_Changes WRT @CendioHalim's  original PR:_**

- I renumbered buttons 8 and 9 to 4 (back) and 5 (forward), respectively. Rationale: the "pseudo buttons" (`4-7`) in X11 event messages that deal with scroll wheels don't appear anywhere in the public FLTK API. These "pseudo buttons" are not sticky and should not appear in `Fl::event_status()`, they were obviously a badly chosen compromise for the X11 protocol which defines only "buttons" 4 and 5 (vertical scroll) but uses also "buttons" 6 and 7 (horizontal scroll). Neither are needed (publicly visible) in FLTK and wouldn't be useful in the event status. My intention is to map pseudo buttons 8 and 9 of the X11 protocol (side buttons) to 4 and 5, respectively.
- Buttons 4 (back) and 5 (forward) appear now in `Fl::event_status()` while dragging as the new bit definitions `FL_BUTTON4` and `FL_BUTTON5`, respectively.
- If this strategy is not correct for any reason all comments and suggestions would be welcomed.
- `test/handle_events.cxx` can be used to test all mouse events. This is what I am using.

@CendioHalim: Thank you again for your great work, particularly on Windows and macOS. I could use it almost completely with only the few modifications mentioned above and the additional status handling. Your contributions are highly appreciated.

**_Further work:_**

I'd appreciate all tests and feedback from @CendioHalim (the OP) and others, using my branch `mouse-buttons` (see above) on macOS and Windows. Thanks in advance.

I'll investigate how to manage the other platforms (X11 + Wayland).

Albrecht-S:
FTR: commit c8dadeffe0661596e26d0bdd0ab899db86acb0e0 of branch `mouse-buttons` in [my fork](https://github.com/Albrecht-S/fltk/tree/mouse-buttons) should work with X11 and Wayland as well, i.e. this branch includes now support of `BACK` and `FORWARD` mouse buttons for all FLTK platforms.

@CendioHalim Please test if this fixes all issues on all FLTK platforms for you. TIA. If you confirm we may eventually close this PR (if we decide to merge my branch).

CendioHalim:
> * @CendioHalim: I started with your PR but modified it in some aspects, see below for details
> 
> * I didn't want to "hijack" your PR, hence I created [my own branch 'mouse-buttons' on my fork](https://github.com/Albrecht-S/fltk/tree/mouse-buttons). I hope you don't mind.

I don't mind at all :) 

> **_Changes WRT @CendioHalim's original PR:_**
> 
>   I renumbered buttons 8 and 9 to 4 (back) and 5 (forward), respectively. Rationale: the "pseudo buttons" (`4-7`) in X11 event messages that deal with scroll wheels don't appear anywhere in the public FLTK API. These "pseudo buttons" are not sticky and should not appear in `Fl::event_status()`, they were obviously a badly chosen compromise for the X11 protocol which defines only "buttons" 4 and 5 (vertical scroll) but uses also "buttons" 6 and 7 (horizontal scroll). Neither are needed (publicly visible) in FLTK and wouldn't be useful in the event status. My intention is to map pseudo buttons 8 and 9 of the X11 protocol (side buttons) to 4 and 5, respectively.
> 
> Buttons 4 (back) and 5 (forward) appear now in `Fl::event_status()` while dragging as the new bit definitions `FL_BUTTON4` and `FL_BUTTON5`, respectively.
> 
I tested your changes on X11, Wayland, macOS and Windows, they seem to work. Thanks for the effort to keep track of the states as well.

I have some feedback for the implementation in `Fl_x.cxx`:

* The following if statement in the `ButtonPress` case: `if (mb < 1 || mb > 9) return 0;`  Before this change, mouse buttons > 9 would still generate `FL_PUSH/RELEASE` correctly. I have a Logitech G400 which is capable of sending button 10, and I presume other mice with more buttons can send buttons even higher than that. Would it not be an OK compromise to implement state tracking for buttons 1-5, and be content with only sending the corresponding `FL_PUSH/RELEASE` for any arbitrary `xevent.xbutton.button > 9`? 

* I haven't double-checked, but wouldn't `xbutton_state` fit inside the already existing `e_state`? If so, there code could be simplified a bit.
  * I could test this some time this week.

* The code in `case ButtonRelease`: could be simplified a bit.

Here are some of my suggestions, feel free to try them out / comment on them:

```diff
diff --git a/src/Fl_x.cxx b/src/Fl_x.cxx
index 6a201a52f..89ca68b3d 100644
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -2043,7 +2043,6 @@ int fl_handle(const XEvent& thisevent)
 
   case ButtonPress: {
     int mb = xevent.xbutton.button;             // mouse button
-    if (mb < 1 || mb > 9) return 0;             // unknown or unsupported button, ignore
 
     // FIXME(?): here we set some event related variables although we *might*
     // ignore an event sent by X because we don't know or want it. This may lead to
@@ -2067,7 +2066,7 @@ int fl_handle(const XEvent& thisevent)
     } else if (mb == 7 || (mb == Button5 && Fl::event_shift())) {
         Fl::e_dx = +1;                          // right
         event = FL_MOUSEWHEEL;
-    } else if (mb < 4 || mb > 7) {              // real mouse *buttons*, not scroll wheel
+    } else {                                      // real mouse *buttons*, not scroll wheel
         if (mb > 7)                             // 8 = back, 9 = forward
           mb -= 4;                              // map to 4 and 5, resp.
         Fl::e_keysym = FL_Button + mb;          // note: mb might have changed
@@ -2076,8 +2075,6 @@ int fl_handle(const XEvent& thisevent)
         if (mb == 5) xbutton_state |= FL_BUTTON5; // save extra button state internally
         event = FL_PUSH;
         checkdouble();
-    } else { // unknown button or shift combination
-      return 0;
     }
 
 #if FLTK_CONSOLIDATE_MOTION
@@ -2091,18 +2088,13 @@ int fl_handle(const XEvent& thisevent)
 
   case ButtonRelease: {
     int mb = xevent.xbutton.button; // mouse button
-    switch (mb) {                   // figure out which real button this is
-      case 1:                       // left
-      case 2:                       // middle
-      case 3:                       // right
-        break;                      // continue
-      case 8:                       // side button 1 (back)
-      case 9:                       // side button 2 (forward)
-        mb -= 4;                    // map to 4 and 5, respectively
-        break;                      // continue
-      default:                      // unknown button or scroll wheel:
-        return 0;                   // don't send FL_RELEASE event
-    }
+
+    if (mb > 3 && mb < 7) // don't send FL_RELEASE event
+      return 0;
+
+    if (mb > 7)
+      mb -= 4;
+
     Fl::e_keysym = FL_Button + mb;  // == FL_BUTTON1 .. FL_BUTTON5
     set_event_xy(window);
 ```

**Comment by CendioHalim** [comment]
> * I haven't double-checked, but wouldn't `xbutton_state` fit inside the already existing `e_state`? If so, there code could be simplified a bit.

Ah, looks like this is needed since `set_event_xy()` overwrites `Fl::e_state` each time an event happens:
```cpp
Fl::e_state   = ((fl_xevent->xbutton.state & event_state_mask) << 16) | xbutton_state;
```

CendioHalim:
> * Would it not be an OK compromise to implement state tracking for buttons 1-5, and be content with only sending the corresponding `FL_PUSH/RELEASE` for any arbitrary `xevent.xbutton.button > 9`?

I had a look at the [Xorg evdev driver](https://gitlab.freedesktop.org/xorg/driver/xf86-input-evdev/-/blob/master/src/evdev.c#L2737) and [input-event-codes.h](https://github.com/torvalds/linux/blob/e32cde8d2bd7d251a8f9b434143977ddf13dcec6/include/uapi/linux/input-event-codes.h#L355C1-L363C24) in Linux, and it seems like X11 won't send a mouse button greater than `BTN_JOYSICK - 1 - BTN_SIDE`, which should come out to 12.

So to handle all normal X11 mouse button events, we should look at numbers between 1-12.

Albrecht-S:
@CendioHalim Thank you very much for your feedback, code suggestions, and pointers to Xorg docs. I appreciate your help very much.

As you noticed yourself, keeping the button states required a separate (static) variable. I didn't want to change the entire `Fl::e_state` implementation too much. However, I believe we're doing the entire status management ourselves on macOS and Wayland anyway, whereas we rely on the OS on Windows and X11. Never change a working system... ;-)

In the future (1.5.0 or later) I can imagine to revisit this and maybe consolidate it on all platforms (this might reduce platform specific code).

For now I hesitate to accept other buttons than "back" and "forward". Although it might be possible to support more mouse buttons on X11 (and maybe Wayland as well?) I hesitate to do it because this wouldn't be cross-platform. Windows doesn't support more than five buttons anyway, hence we'd need to make an exception for Linux/X11/Wayland (I don't know if we could do it on macOS). I don't want to preclude this for all times, but for now I believe it's better to allow only these well-defined buttons on **_all_** platforms. We should reserve to extend this for later when (**if!**) we can define a clear API how to deal with even more buttons. IIRC I read somewhere that (gaming) mice that provide more buttons typically send keyboard events instead of button events for these extra buttons.

This needs more investigation before we do something we'd have to redesign later. One point would be to decide how the button numbers of these extra buttons should be dealt with, for instance: would "X11 button" 10 be mapped to 6 or not (leave it at 10)? We should keep that in mind for a later update.

I fixed up my own branch (added some more docs and missing definitions), squashed and rebased it and created PR #1081. I'm waiting for more feedback and reviews of other FLTK devs to decide how to proceed. Once we decide to merge my PR we can close this one.

CendioHalim:
> For now I hesitate to accept other buttons than "back" and "forward". Although it might be possible to support more mouse buttons on X11 (and maybe Wayland as well?) I hesitate to do it because this wouldn't be cross-platform. Windows doesn't support more than five buttons anyway, hence we'd need to make an exception for Linux/X11/Wayland (I don't know if we could do it on macOS). I don't want to preclude this for all times, but for now I believe it's better to allow only these well-defined buttons on **_all_** platforms.

Fair point.

Thanks again for taking the time to do this work, much appreciated.

PR #1080, opened 2024-09-28
Fl_Text_Display: Fix text selection dragging off-by-one bug

qartis:
The bug can be reproduced in examples/texteditor-simple, and it's easier if you zoom in with Ctrl-+ or add edit->textsize(100). The endpoint of the text selection is always shifted right by one character which leads to strange selection behavior.

### DIFF
--- a/src/Fl_Text_Display.cxx
+++ b/src/Fl_Text_Display.cxx
@@ -4335,7 +4335,6 @@ int Fl_Text_Display::handle(int event) {
           scroll_direction = 0;
         }
         pos = xy_to_position(X, Y, CURSOR_POS);
-        pos = buffer()->next_char(pos);
       }
       fl_text_drag_me(pos, this);
       return 1;


Albrecht-S:
@qartis Thanks for the report. I tried to reproduce the issue and could not in my first try. Then I noticed that the code in question was about **_dragging the selection_**. I changed the title of this PR accordingly.

I confirm that the selection dragging behavior is weird, as the OP wrote. In my own words: the end of the selection "jumps early" to the next character and thus includes one more character than expected. This was observed when dragging the selection from left to right. Other operations not tested.

I didn't investigate further though, leaving this for someone more familiar with the code - or for later.

Albrecht-S:
More test results: when dragging either end of the selection from right to left the start or end of the selection seems to move (jump) too late, i.e. later than expected. A quick test with the patch seems to confirm that the patch works. Again, I'm not going to verify the patch now...

MatthiasWM:
Thanks for the fix.

MatthiasWM:
Thanks, Albrecht.

Issue #1076, opened 2024-09-24
X11: No FL_RELEASE when releasing the forward/back mouse button


CendioHalim:
**Describe the bug**
I don't get FL_RELEASE when I click on the forward/back side buttons on my mouse. These buttons map to `xevent.xbutton.button` 8 and 9.

**To Reproduce**
1. Click on the forward or back mouse button.

**Expected behavior**
I should get an `FL_RELEASE` when I release the mouse button.

**FLTK Version**
This commit https://github.com/fltk/fltk/commit/2252ba98390505f353ea35e8ade3e3914b46c977 broke this behaviour.

A simple fix would be to explicitly check if the button is between 4-7:

```diff
diff --git a/src/Fl_x.cxx b/src/Fl_x.cxx
index dcae946ab..bd2843db2 100644
--- a/src/Fl_x.cxx
+++ b/src/Fl_x.cxx
@@ -2089,7 +2089,7 @@ int fl_handle(const XEvent& thisevent)
     Fl::e_keysym = FL_Button + xevent.xbutton.button;
     set_event_xy(window);
     Fl::e_state &= ~(FL_BUTTON1 << (xevent.xbutton.button-1));
-    if (xevent.xbutton.button > Button3) { // "buttons" 4-7 = mousewheel events: don't send FL_RELEASE
+    if (xevent.xbutton.button > 3 && xevent.xbutton.button < 8 ) { // "buttons" 4-7 = mousewheel events: don't send FL_RELEASE
       return 0;
     }
```

CendioHalim:
Something else seems to be broken as well. If I hold down LMB and keep it pressed, then press and release either side button, I get an `FL_RELEASE` the patch above. But as soon as I move the mouse again, I will keep receiving `FL_PUSH` events for the side button until LMB is released.

Albrecht-S:
@CendioHalim Thank you for the report and the proposed patch. Your simple patch is not going to work for several reasons, but it's a good start. However, we would need to extend some constants (enums) for `Fl::event_state()` like for instance `FL_BUTTONS`, but very likely even more. All mouse event related code would need to be checked as well.

[...]

PR #1081, opened 2024-10-02
Support mouse buttons 4 + 5 (aka "side buttons") (#1076, #1068)

Albrecht-S:
This work is based on PR 1068 (patch by @CendioHalim) and extended to store button status (4,5) in Fl::event_state() like it's done for other mouse buttons (1-3).

Related: Issue #1076

Changes:
- new symbol: `FL_BUTTON4` = side button 1 = "back"
- new symbol: `FL_BUTTON5` = side button 2 = "forward"
- modified  : `FL_BUTTONS` now includes bits for two side buttons

Note: the status of these new buttons is not maintained by X11, therefore we need to maintain them in internal variables for this platform.

This PR has been tested on all supported platforms (Windows, macOS, Linux/X11, and Wayland).

### DIFF
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -548,9 +548,11 @@ enum Fl_Callback_Reason {
 
 /**@{*/
 
-#define FL_LEFT_MOUSE   1       ///< The left mouse button
-#define FL_MIDDLE_MOUSE 2       ///< The middle mouse button
-#define FL_RIGHT_MOUSE  3       ///< The right mouse button
+#define FL_LEFT_MOUSE     1       ///< The left mouse button
+#define FL_MIDDLE_MOUSE   2       ///< The middle mouse button
+#define FL_RIGHT_MOUSE    3       ///< The right mouse button
+#define FL_BACK_MOUSE     4       ///< The back mouse button (side button 1)
+#define FL_FORWARD_MOUSE  5       ///< The forward mouse button (side button 2)
 
 /**@}*/          // group: Mouse Buttons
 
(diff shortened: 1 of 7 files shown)


CendioHalim:
Is there any chance of getting this to branch-1.3 as well if I backport these changes?

Albrecht-S:
**_Notes:_**
1. This PR fixes a potential bug as well, where unexpected button numbers could set arbitrary (undefined) bits in `Fl::event_status()`, some inconsistencies, and includes documentation improvements.

2. This PR ignores any other mouse buttons than "back" and "forward". This is intentional because we'd have to define a clear API for more mouse buttons across platforms. [Windows doesn't support more mouse buttons](https://learn.microsoft.com/en-us/windows/win32/learnwin32/mouse-input) anyway (they call them `XBUTTON1` and `XBUTTON2`), hence having mouse buttons 4 and 5 on all platforms is an advantage.

3. I propose to include this PR in FLTK 1.4.0 although we're officially on feature freeze. The mentioned inconsistencies and bug fixes are IMHO worth it.

@ManoloFLTK, @MatthiasWM : please review the code, particularly on macOS and Wayland, but Windows and Linux would also be appreciated. Thanks in advance.

All: please feel free to test and give feedback here.

Albrecht-S:
> Is there any chance of getting this to branch-1.3 as well if I backport these changes?

Hmm, I don't know; I see how it could be useful for TigerVNC...

However, current status of branch-1.3 is that it gets only **_serious bug fixes_**, particularly because we want to avoid regressions. We should definitely wait (with porting attempts) until this code settles in current master (API, if/when merged) or at least for test and review results of other devs.

I'm open for comments and suggestions.

MatthiasWM:
Ok, the stupid app to allow macOS to configure my Logitech mouse is 250MB. They should have used FLTK and they could have written it in 2MB... . Testing on macOS now.

MatthiasWM:
Ok, yes, it all works on my macOS 14.6.1 M1. I added labels and the correct user value for the state indicators in `test/keyboard`. 

My cheap Favo mouse works perfectly well with this after I disabled the macOS default for the _back_ button. My comparatively expensive Logitec MX Anywhere however only sends a short burst after either of the buttons is released. More interestingly, after installing the Logictec 250MB mouse setup software, the mouse no longer send mouse button 4 and 5, but instead sends a _swipe_ gesture (`NSEventTypeSwipe`), and I have <rolleyes> not found a way to make it send mouse button events again without uninstalling their software! </rolleyes>

MatthiasWM (approved):

Albrecht-S:
> I added labels and the correct user value for the state indicators in test/keyboard.

@MatthiasWM Great, thanks for this addition. Do you think that this is ready to be merged in 1.4.0? I think we should do it.

MatthiasWM:
Yes, I think so.

Albrecht-S:
@MatthiasWM Done, thanks for your help.

PR #1104, opened 2024-11-03
macOS: Change required version for inclusion of ScreenCaptureKit

MatthiasWM:
Unverified. 

Changes the use of MAC_OS_X_VERSION_MAX_ALLOWED and MAC_OS_X_VERSION_MIN_REQUIRED for ScreenCaptureKit to be able to sync compile and CMake based linking with CMAKE_OSX_DEPLOYMENT_TARGET.


MatthiasWM:
I can't solve this. Sorry.

CMake/setup.cmake
```diff
@@ -117,17 +117,22 @@ if(APPLE)
...
+    set(CMAKE_OSX_DEPLOYMENT_TARGET 10.8)
+  endif()
   if(FLTK_BACKEND_X11)
     if(NOT(${CMAKE_SYSTEM_VERSION} VERSION_LESS 17.0.0)) # a.k.a. macOS version ≥ 10.13
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_LIBCPP_HAS_THREAD_API_PTHREAD")
     endif(NOT(${CMAKE_SYSTEM_VERSION} VERSION_LESS 17.0.0))
   else()
     set(FLTK_COCOA_FRAMEWORKS "-framework Cocoa")
+    message(STATUS "CMAKE_OSX_DEPLOYMENT_TARGET: ${CMAKE_OSX_DEPLOYMENT_TARGET}")
+    message(STATUS "CMAKE_OSX_SYSROOT: ${CMAKE_OSX_SYSROOT}")
```
Albrecht-S, line 130: Just a hint: logging contents of CMake variables should be done with `fl_debug_var(NAME)`. This has three advantages:
1. less typing
2. less potential for errors (you don't have to type the same text twice)
3. "nice" alignment of variable names and values if more than 1 variable is being logged

Sorry, I missed that in my previous review, but there's not need to change it since this would be commented out anyway...

CMake/setup.cmake
```diff
@@ -117,17 +117,22 @@ if(APPLE)
   set(HAVE_VSNPRINTF 1)
   set(HAVE_SCANDIR 1)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated")
+  if (NOT DEFINED ${CMAKE_OSX_DEPLOYMENT_TARGET})
+    set(CMAKE_OSX_DEPLOYMENT_TARGET 10.8)
+  endif()
```
Albrecht-S, line 122: Setting this is a normal CMake variable is probably not a good idea since the real variable is stored in the cache, and doing it like above will "shadow" the cache variable, i.e. the user doesn't have much control and can't "see" the contents and can't effectively change it.
I suggest to change the assignment to:
```
  set(CMAKE_OSX_DEPLOYMENT_TARGET 10.8 CACHE STRING "<comment>")
```
I'm not sure about the comment though. See more in my comments in "Conversation". Needs testing...

Albrecht-S:
Regarding the `<comment>` for the cache variable mentioned above: in cmake-gui the cache variable generated by CMake (after removal) displays the tooltip one can see in the CMake cache in the two lines above the variable:
```
$ grep -B2 -A0 DEPLOY CMakeCache.txt 
//Minimum OS X version to target for deployment (at runtime); newer
// APIs weak linked. Set to empty string for default value.
CMAKE_OSX_DEPLOYMENT_TARGET:STRING=
```

Note also that the `if` clause `if (NOT DEFINED ${CMAKE_OSX_DEPLOYMENT_TARGET})` is neither necessary nor useful for cache variables, i.e. we could write (w/o the `if`/`endif` statements):
```
set(CMAKE_OSX_DEPLOYMENT_TARGET 10.8 CACHE STRING "Minimum OS X version to target for deployment ...")

The result would be:
1. if the user sets or changes the cache variable we honor their choice
2. if the user doesn't set the cache variable we set the minimum deployment target to 10.8

... if (2.) is what we want.

Note that CMake also mentions: "Set to empty string for default value." Should we really override this default value?

Albrecht-S:
Note: reopened for discussion...

MatthiasWM:
Ah, I abandoned this Pull Request. This was just a very first rough layout. PR #1105 is the current proposed patch. 

I tried this way before I realized that CMAKE_OSX_DEPLOYMENT_TARGET has little influence on what the source code produces, and then what needs to be linked. #1105 has a completely different approach and is agnostic to CMAKE_OSX_DEPLOYMENT_TARGET.

Setting `CMAKE_OSX_DEPLOYMENT_TARGET` in the CMake files is not very useful. The dev should do that manually if they need it to be set to something other than the default.

Albrecht-S:
Ah, OK, I missed #1105. I'll take a look. Closing this one again.

Issue #1103, opened 2024-11-03
Xcode apps no longer compile


MatthiasWM:
**Describe the bug**
No apps compile anymore using the CMake setup on macOS due to ScreenCapture library no being linked anymore.

**To Reproduce**
- Update Xcode to 16.0, 
- delete the build folder
- rebuild the cmake build folder using `cmake -B build/Xcode -G Xcode .`
- launch Xcode, compile any test app, and you get 4 linke errors similar to: `Undefined symbol: _OBJC_CLASS_$_SCContentFilter`

**FLTK Version**
 - Version: 1.4.0 master
 - If from Git, branch: master

**FLTK Configure / Build Options**
Please add the complete configure and build command lines if you built FLTK yourself.
Please include all configure or CMake options and add all CMake options you changed
by using CMake GUI tools (if any).
 - `cmake -B build/Xcode -G Xcode .`

**Operating System / Platform:**
- macOS 14.6.1 (23G93)
- Xcode Version 16.0 (16A242d)
- cmake version 3.30.2
- uname -r  = 23.6.0
- CMake ${CURRENT_OSX_VERSION} 14.6.1

**Additional context**
In CMake/Ssetup.CMake, line 130 reads `if(${CMAKE_SYSTEM_VERSION} VERSION_GREATER_EQUAL 24.0)` but the variable is set to 23.6.0 on my system after the upgrade to Xcode 16.0. I upgraded all my Brew stuff and got CMake 3.30.5, but CMAKE_SYSTEM_VERSION stays at 23.6.0. As a result, `-weak_framework ScreenCaptureKit` is not added to FLTK_COCOA_FRAMEWORKS.

My macOS docs say, ScreenCaptureKit was introduced with macOS 12.3.

Would it make more sense to query CMAKE_OSX_DEPLOYMENT_TARGET in the CMake file, which is 14.6 on my machine, for example?

Albrecht-S:
Note: FLTK 1.4.0 RC3 is delayed until this bug has been fixed. Hopefully soon ;-)

MatthiasWM:
@ManoloFLTK 

Puttingg together more information. So CMAKE_OSX_DEPLOYMENT_TARGET changes MAC_OS_X_VERSION_MIN_REQUIRED, but the requirement to link the library depends on MAC_OS_X_VERSION_MAX_ALLOWED. I am reading up on those macros now, hoping I can give a solutions quickly.

MatthiasWM:
[...]

PR #1105, opened 2024-11-04
Using try_run to get __MAC_OS_X_VERSION_MAX_ALLOWED

MatthiasWM:
Possible solution for #1103 ?


CMake/setup.cmake
```diff
@@ -117,6 +117,17 @@ if(APPLE)
   set(HAVE_VSNPRINTF 1)
   set(HAVE_SCANDIR 1)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated")
+  try_run(
+    RUN_RESULT COMPILE_RESULT
+    SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/CMake/macosMaxAllowd.c
+    RUN_OUTPUT_VARIABLE MAC_OS_X_VERSION_MAX_ALLOWED
+  )
+  if (COMPILE_RESULT STREQUAL "FALSE")
```
Albrecht-S, line 125: Can probably be simplified to:
```
if (NOT COMPILE_RESULT)
```

CMakeLists.txt
```diff
@@ -345,3 +345,5 @@ if(0) # debug built library and fluid targets
   endforeach()
   message(STATUS "---------------------------------------------------------")
 endif()
+
```
Albrecht-S, line 348: Adding these two empty lines should be reverted. This file shouldn't be changed at all

CMake/macosMaxAllowd.c
```diff
@@ -0,0 +1,8 @@
+#include <AvailabilityMacros.h>
+#include <stdio.h>
+
+int main(int argc, char **argv) {
+  unsigned int ver = __MAC_OS_X_VERSION_MAX_ALLOWED;
+  printf("%d.%d.%d", ver/10000, (ver/100)%100, ver%100);
+  return 0;
+}
```
Albrecht-S, line 8: If this is going to be used, please rename this to `CMake/macosMaxAllowed.c` (fix spelling).

CMake/macOSMaxAllowed.c
```diff
@@ -0,0 +1,8 @@
+#include <AvailabilityMacros.h>
```
Albrecht-S, line 1: @MatthiasWM, @ManoloFLTK Please check if this header file is available on all targeted macOS versions. ISTR that the header `<Availability.h>` was used on older macOS versions.

CMake/macOSMaxAllowed.c
```diff
@@ -0,0 +1,8 @@
...
+#include <stdio.h>
+
+int main(int argc, char **argv) {
+  unsigned int ver = __MAC_OS_X_VERSION_MAX_ALLOWED;
+  printf("%d.%d.%d", ver/10000, (ver/100)%100, ver%100);
```
Albrecht-S, line 6: Same as above, should it fail if the "underscored" macro is not defined?

CMake/macOSMaxAllowed.c
Albrecht-S: _Hmm, I'm confused, I thought I added this already:_

Maybe it's intended to fail if `<AvailabilityMacros.h>` does not exist? Then it's correct, obviously.

MatthiasWM: `<Availability.h>` is the correct header, and `__MAC_OS_X_VERSION_MAX_ALLOWED` should always be defined there if we target OS X (vs. iPhone, watch, or what have you). If the include file does not exist, or the macro is not there, the caller falls back to the host system version `CURRENT_OSX_VERSION`.

CMake/macOSMaxAllowed.c
MatthiasWM: Fixed the include file, failure calls a reasonable fallback. Tested.

Issue #1116, opened 2024-11-09
[Bug] Windows: fullscreen_off does not correctly preserve window size


dannye:
**Describe the bug**

On Windows, If a fullscreened window is un-fullscreened with `fullscreen_off()`, its position doesn't match what it should.

1. In the case of a non-maximized window, the window is slightly smaller than it should be. If fullscreen is toggled on/off repeatedly, the window continues to slowly shrink.

2. In the case of a maximized window, the window is too tall and the bottom edge of the window is hiding behind the taskbar.

No such issue happens on Linux or Mac.

**To Reproduce**

I added a shortcut to the fullscreen button in fullscreen.cxx just to make it easier to trigger:

```diff
   window.b3 = new Fl_Toggle_Light_Button(50,y,window.w()-60,30,"FullScreen");
   window.b3->callback(fullscreen_cb,w);
+  window.b3->shortcut(FL_F + 11);
   y+=30;
```

1. Run test/fullscreen.exe
2. Either make the window maximized, or not
3. Toggle fullscreen on
4. Toggle fullscreen off
5. Repeat steps 3 and 4 many times (not necessary for a maximized window)
6. See that the window is either too small if non-maximized, or too tall if maximized

**Expected behavior**

Both a maximized and non-maximized window should be restored to its original size when fullscreen is toggled off.

**Videos**

1.

https://github.com/user-attachments/assets/51b52f68-34fe-4dd4-89a6-713445dab691

2.

https://github.com/user-attachments/assets/9540bd23-fb87-4344-a269-1a392dd25850

Notice that the "Update" button at the bottom of the window is not in view when fullscreen is toggled off at the 10 second mark.

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git
 - If from Git, branch: release-1.4.0rc3

**FLTK Configure / Build Options**
 - Default CMake build in Visual Studio

**Operating System / Platform:**
 - OS: Windows
 - OS Version: Windows 11

ManoloFLTK:
@dannye Please, close issue if you confirm the commit mentioned above fixes it.

dannye:
[...]

PR #1118, opened 2024-11-09
Windows: Fix flicker/animation when transitioning from fullscreen to maximized

dannye:
This is a continuation of #1116.

Avoid the flicker and animation when leaving fullscreen and restoring maximized state by using `SetWindowPos()` to set the window to use the whole work area of the current screen instead of calling `pWindow->maximize();`.

Before:

https://github.com/user-attachments/assets/4a23608e-0bcc-445c-9f3c-bdd64c34f853

After:

https://github.com/user-attachments/assets/5ace8ba8-fd16-4c7b-8c9b-91c94fbef36f


ManoloFLTK:
Many thanks for this nice improvement.

PR #1126, opened 2024-11-17
Fix MSVC Level4 compiler warnings in public header files

dannye:
If a user application wishes to enable Level4 compiler warnings in Visual Studio (eg, `<WarningLevel>Level4</WarningLevel>`) then the public FL header files will produce many warnings, such as:

> ...\include\FL\Fl_Widget.H(528,60): warning C4244: '=': conversion from 'Fl_Labeltype' to 'uchar', possible loss of data
> ...\include\FL\Fl_Button.H(161,44): warning C4244: '=': conversion from 'Fl_Boxtype' to 'uchar', possible loss of data

To fix this, the application developer must wrap all FL includes like so:

```cpp
#pragma warning(push, 0)
#include <FL/Fl_Widget.H>
#include <FL/Fl_Button.H>
#pragma warning(pop)
```

It would be nice to be able to finally delete all the `#pragma warning(push, 0)`/`#pragma warning(pop)`.

This PR does the bare minimum to add the appropriate casts to fix the warnings in the public header files only.

This was original reported in fltk.general: https://groups.google.com/g/fltkgeneral/c/SQ3RSqNgXbE/m/lIl2lnjdBQAJ

This possibly raises the question of why there are some mismatches between data member types and argument types (in setters, for example), but now is obviously not the time to address all of it. This PR changes no functionality at all.

### DIFF
--- a/FL/Fl_Button.H
+++ b/FL/Fl_Button.H
@@ -158,7 +158,7 @@ public:
 
     \param[in] b down box type
    */
-  void down_box(Fl_Boxtype b) {down_box_ = b;}
+  void down_box(Fl_Boxtype b) {down_box_ = (uchar)b;}
 
   /// (for backwards compatibility)
   void shortcut(const char *s) {shortcut(fl_old_shortcut(s));}
--- a/FL/Fl_File_Input.H
+++ b/FL/Fl_File_Input.H
@@ -66,7 +66,7 @@ public:
   /** Gets the box type used for the navigation bar. */
   Fl_Boxtype    down_box() const { return (Fl_Boxtype)down_box_; }
   /** Sets the box type to use for the navigation bar.  */
-  void          down_box(Fl_Boxtype b) { down_box_ = b; }
+  void          down_box(Fl_Boxtype b) { down_box_ = (uchar)b; }
 
   /**
     Gets the current error color.
--- a/FL/Fl_Grid.H
+++ b/FL/Fl_Grid.H
@@ -177,15 +177,15 @@ public:
 
     Cell(int row, int col) {    // constructor
       Cell_();
-      row_ = row;
-      col_ = col;
+      row_ = (short)row;
+      col_ = (short)col;
     }
 
     Cell(Fl_Widget *w, int row, int col) {  // widget assignment
       Cell_();
       widget_ = w;
-      row_ = row;
-      col_ = col;
+      row_ = (short)row;
+      col_ = (short)col;
     }
 
     /**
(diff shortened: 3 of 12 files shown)


dannye:
From the fltk.general thread, @Albrecht-S wrote:

> (3) "warning C4244" is a known issue but benign and nobody found the
time to fix it. We could add some casts that would likely remove the
warnings but this would make the code longer and ugly. Anyway, it might
be done if someone could find the time to do it. Patches (valid for all
platforms!) always welcome...

Yes, it probably makes the code just a tad uglier, but having a "spotlight" (the explicit cast) shining on the eyebrow-raising type mismatches is also a good thing IMO.

I'm maybe a bit worried about the "(valid for all platforms!)" comment.. If any of my casts are not truly cross-platform compatible then please let me know. I will also do my own testing. (Only tested on Windows at the time of writing.)

dannye:
FTR: An alternative option instead of adding the casts is to modify some of the enum declarations in Enumerations.H, for example:

```diff
- enum Fl_Boxtype {
+ enum Fl_Boxtype : uchar {
```

But this is probably a worse option.
This has the benefit of making `Fl_Boxtype` and `uchar` synonymous (making the cast unnecessary) but it is possibly worse for backward- and forward-compatibility.

MatthiasWM:
> ```diff
> - enum Fl_Boxtype {
> + enum Fl_Boxtype : uchar {
> ```

`enum name ﻿() : type { }` is a C++11 feature. FLTK 1.3 is still C++98 compatible (as in 1998). This will change with 1.5 (C++17 is my recommendation).

Your patch looks simple enough, but then again changes a lot of code between RC3 and 1.4.0 . I will talk to the core team today for a decision.

dannye:
> I will also do my own testing. (Only tested on Windows at the time of writing.)

I see no issues with this PR on my Linux and Mac machines.

> Your patch looks simple enough, but then again changes a lot of code between RC3 and 1.4.0 . I will talk to the core team today for a decision.

Thanks @MatthiasWM. Whatever decision you all make is OK with me.

MatthiasWM:
I finally found the time to fix the warning listed and a few more.

dannye:
Great, thank you @MatthiasWM!

Just noting one extremely minor point:

This whitespace misalignment on line 491 was handled by this PR but not by 1ba957e:

https://github.com/fltk/fltk/blob/1ba957eb6096c1df1e6476cccea5dbd3ec3b189d/FL/Fl_Terminal.H#L490-L492

Thanks again!

MatthiasWM:
> This whitespace misalignment on line 491 was handled by this PR but not by [1ba957e]

It's fixed and will be in the next commit.

PR #1131, opened 2024-11-18
Fix test/fltk-versions on Windows

dannye:
Some compilers allow variable-sized stack arrays, but msvc does not, so allocate it on the heap instead.

### DIFF
--- a/test/fltk-versions.cxx
+++ b/test/fltk-versions.cxx
@@ -120,7 +120,7 @@ int main(int argc, char **argv) {
   grid->row_height(0, 40);
   title->labelsize(20);
 
-  Fl_Box *box[versions];
+  Fl_Box **box = new Fl_Box*[versions];
   for (int i = 0; i < versions/3; i += 1) {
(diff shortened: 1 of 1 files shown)


dannye:
Other alternatives:

```cpp
Fl_Box *box[sizeof(version) / sizeof(version[0])];
```

or just:

```cpp
Fl_Box *box[9];
```

Albrecht-S:
Thanks for your report. I **knew** that I had forgotten something (building on VS and checking the CI builds). Sorry, but the last days (pushing release 1.4.0) were pretty hard.

I'll fix the code ASAP but I'm not sure if / when to issue a new release. Let's wait a while, particularly since this affects only **one** test/demo program.

Albrecht-S:
@dannye I didn't realize that this was a PR and you suggested a fix. My solution is different, hence this PR is no longer needed.

I also decided to publish a release update (`1.4.0-1`) to prevent bug reports and to provide users with a clean build (hopefully).

I'm closing this PR now. Thanks for your help.

dannye:
No problem at all. Thanks!

Issue #877, opened 2023-12-25
`FL_SHORTCUT` and `FL_KEYBOARD` do not honor `FL_WHEN_RELEASE` for `FL_RADIO_BUTTON`


dannye:
**Describe the bug**

The description for `FL_WHEN_RELEASE` is:

> FL_WHEN_RELEASE           =  4, ///< Do the callback when the button or key is released and the value changes

However, a key down event for the shortcut (`FL_SHORTCUT`) or spacebar on a focused widget (`FL_KEYBOARD`) will trigger the callback, even when the radio button is already set and `when_` is `FL_WHEN_RELEASE`. This is notably different from a mouse button release (`FL_RELEASE`), which will _not_ invoke the callback if the radio button is already set.

The actual behavior is more like what I would expect for if `when_` is set to `FL_WHEN_RELEASE_ALWAYS`.

**Additionally**: Astonishingly, `FL_SHORTCUT` <ins>never</ins> calls `set_changed()` and `FL_KEYBOARD` <ins>always</ins> calls `set_changed()` (in `Fl_Button::handle()`), regardless of whether the radio button value is changing or not. This thwarts any possibility of checking `if (button->changed())` in the callback to decide whether to run the rest of the callback or not.

Also, holding down the spacebar on a focused radio button invokes the callback on each repeat. Of course, if the callback was only invoked on value change, then this issue would "fix itself".

All of this together means it's unreasonably difficult to create a radio button that only invokes its callback _on change_, which is what `FL_WHEN_RELEASE` is intended for, afaict.

In order to work around this, one would either need to subclass to override `handle()` in order to return early if the radio button is already set:

```cpp
if ((event == FL_SHORTCUT || event == FL_KEYBOARD) && value()) {
	return 0;
}
```

or one would need to store some separate/redundant state to manually track radio button changes to know when the full callback should be executed.

Both of these workarounds seem non-ideal.

**To Reproduce**

This can be witnessed in the default test program, radio.cxx.

However, here is an alternate radio.cxx with a more practical example that demonstrates why having better control over only invoking the callback on change would be advantageous:

<details>

<summary>radio.cxx</summary>

```cpp
#include "radio.h"

#include <FL/Fl_Int_Input.H>

#include <string>

Fl_Int_Input *int_input=(Fl_Int_Input *)0;

static void inches_radio_button_cb(Fl_Button *b, void *) {
  std::string val = std::to_string(std::atoi(int_input->value()) * 12);
[...]

PR #1145, opened 2024-11-26
Fix button down state when triggered by shortcut

dannye:
Sorry, I should have noticed this when verifying #877 but I didn't catch it.

The `value(1)`/`value(0)` are not strictly necessary because it's already handled by `simulate_key_action()` and `key_release_timeout()`.
And resetting the value back to 0 so soon prevents the button from ever being drawn in its down state.

Before:

https://github.com/user-attachments/assets/4b6d903b-33fc-4024-b84e-4b8493a793df

After:

https://github.com/user-attachments/assets/feac7799-335d-4c3c-86cf-31a759b9b987

The only functional difference is that now the `do_callback(FL_REASON_RELEASED)` happens while the `value_` is still 1, since the early `value(0)` was removed.

I'm not sure what's best here..

The options I see are:

1. Let the `do_callback(FL_REASON_RELEASED)` happen while the `value_` is still 1. (This PR)
2. Do something like this instead:
    ```diff
                 value(0);
                 set_changed();
                 do_callback(FL_REASON_RELEASED);
    +            value(1);
               } else if (when() & FL_WHEN_RELEASE) {
                 value(0);
                 set_changed();
                 do_callback(FL_REASON_RELEASED);
    +            value(1);
               }
             }
             return 1;
    ```
    so that `do_callback(FL_REASON_RELEASED)` happens while `value_` is indeed 0 (if this is even important) but the button is still correctly drawn in the down state. (Simple change but maybe a little strange)
3. Let `do_callback(FL_REASON_RELEASED)` be the responsibility of `key_release_timeout()` instead of `handle()`. (Maybe the most complicated)

If the value of `value_` isn't particularly important during `do_callback(FL_REASON_RELEASED)` then this PR should be sufficient. But I'm happy to make more changes.

### DIFF
--- a/src/Fl_Button.cxx
+++ b/src/Fl_Button.cxx
@@ -189,17 +189,14 @@ int Fl_Button::handle(int event) {
             do_callback(FL_REASON_RELEASED);
         } else {
           simulate_key_action();
-          value(1);
           if (when() & FL_WHEN_CHANGED) {
             set_changed();
             Fl_Widget_Tracker wp(this);
             do_callback(FL_REASON_CHANGED);
             if (wp.deleted()) return 1;
-            value(0);
             set_changed();
             do_callback(FL_REASON_RELEASED);
           } else if (when() & FL_WHEN_RELEASE) {
-            value(0);
             set_changed();
             do_callback(FL_REASON_RELEASED);
           }


dannye:
The reason I vote for this PR as-is is because the callback reason of `FL_REASON_RELEASED` kind of already communicates that the button has been released, and this logic is just for regular buttons, not check/radio buttons, so the button's "value" isn't significant in the same way.

Albrecht-S:
@dannye Thanks for this PR, it will be considered before the release of 1.4.1. If it is acceptable it will be merged into the 1.4.1 release...

dannye:
@MatthiasWM Your input on this would be appreciated as well. I really like in 554bccb how you added several more potential callbacks to the shortcut flow which make shortcuts mimic mouse presses much more closely. The only issue is that the `value(0)` introduces a visual regression in 1.4.0 (even though it _does_ make the behavior of the callback match mouse presses even more closely). I think the right compromise is to just let shortcuts differ from mouse presses in this one small detail.

Albrecht-S:
@dannye I checked your PR out and tested it. I can see the changes in the visual effect, i.e. the button gets drawn in the "down" state when hitting a corresponding key, either the shortcut or space if the button has the focus. So far, so good.

However, I'm not sure whether this change should be implemented (now). It introduces an inconsistency in callback handling in the following test scenario:
1. click the top left button with the mouse
2. press `space` key
3. press `F` key (shortcut of the top left button).

#### Old code (master @ ff161187e78bb958adac6a76bb7c325146892b10):
```
Label: '&Fl_Button A1'
Value: 0
Changed: 128
Label: '&Fl_Button A1'
Value: 0
Changed: 128
Label: '&Fl_Button A1'
Value: 0
Changed: 128
```

#### Modified code (this PR):
```
Label: '&Fl_Button A1'
Value: 0
Changed: 128
Label: '&Fl_Button A1'
Value: 1
Changed: 128
Label: '&Fl_Button A1'
Value: 1
Changed: 128
```

- The old code is consistent, i.e. `Value` is always `0` in the callback.
- The modified code has `Value == 0` on mouse click and `Value == 1` on shortcut events.

I took a look at the modified code but didn't really check it. I just **tested** it so far. There are at least the following questions:
1. Which behavior (mouse vs. shortcut `value`) is **correct**?
2. Should we care about this inconsistency?
3. Would the changed behavior cause regressions in user code?
4. Should we go further and change the mouse click behavior as well so it sets `value == 1` as well?
5. Is this behavior documented, and if yes, what do the docs define? (not checked yet).

@MatthiasWM and others: what do you think?

As it stands I'd tend not to merge this PR in master now (i.e. in 1.4.1). I'm not aware of any bug reports (except this PR). This might be further evaluated (and eventually documented) as soon as 1.5 development begins.

All comments and opinions appreciated.

Albrecht-S:
Surprise, surprise... :-(

Using `test/radio` of `branch-1.3` I get this (same test as above):
```
Label: '&Fl_Button A1'
Value: 0
Label: '&Fl_Button A1'
Value: 1
Label: '&Fl_Button A1'
Value: 1
```

**_This looks pretty much like the modified code of this PR !!!_**

Further tests reveal: building the 1.4 source of test/radio(.fl) with FLTK 1.3 displays:
```
Label: '&Fl_Button A1'
Value: 0
Changed: 128
Label: '&Fl_Button A1'
Value: 1
Changed: 0
Label: '&Fl_Button A1'
Value: 1
Changed: 0
```
**_Note:_** the "changed" flag is always 0 (as opposed to FLTK 1.4). This is just a fact, but: what is correct, what is a regression, what should be "fixed"? I dunno.

dannye:
Thank you @Albrecht-S for taking a look.

> However, I'm not sure whether this change should be implemented (now). It introduces an inconsistency in callback handling in the following test scenario:
> 
> (...)
> 
> * The old code is consistent, i.e. `Value` is always `0` in the callback.
> 
> * The modified code has `Value == 0` on mouse click and `Value == 1` on shortcut events.

Yes, this is the exact "inconsistency" I refer to above.

However, this PR doesn't *create* the inconsistency. It returns to the inconsistency of 1.3.x!

Commit 554bccb made things "consistent" at the cost of removing all visual indicators that a button has been pressed by a keyboard shortcut.

This is \**especially*\* bad for applications that disable visible focus! (Like all the buttons in my application.)

> 1. Which behavior (mouse vs. shortcut `value`) is **correct**?

That's actually an interesting (debatable) question. My answer is related to the next question...

> 2. Should we care about this inconsistency?

IMHO, no.
For the reason I stated [here](https://github.com/fltk/fltk/pull/1145#issuecomment-2499551589):

> the callback reason of `FL_REASON_RELEASED` kind of already communicates that the button has been released, and this logic is just for regular buttons, not check/radio buttons, so the button's "value" isn't significant in the same way.

> 3. Would the changed behavior cause regressions in user code?

No, because this PR returns behavior of value_ to the 1.3.x behavior.

I'll leave the last two questions to @MatthiasWM and others.

> As it stands I'd tend not to merge this PR in master now (i.e. in 1.4.1).

I politely implore you to merge it. It's true this could continue to get improved in the future (in the ways that you've described) but this PR easily gets things "back on track" for the time being.

Albrecht-S:
> I politely implore you to merge it. It's true this could continue to get improved in the future (in the ways that you've described) but this PR easily gets things "back on track" for the time being.

I'm already reconsidering my above statement, see my follow-up comment which happened to be posted almost at the time as your comment...

Albrecht-S:
I'm now following @dannye's arguments and my vote is to merge this PR. The main reasons are:
- important: consistency with FLTK 1.3, no regression
- this includes the visual button press effects when using `space` and the button shortcut which **are** available in 1.3

The inconsistency between mouse clicks and keyboard events is less important because it existed in 1.3 (no regression).

But anyway, I'd like to leave the final decision to @MatthiasWM who changed the button code recently (including to merge another PR of @dannye IIRC).

dannye:
> > 1. Which behavior (mouse vs. shortcut `value`) is **correct**?
> 
> That's actually an interesting (debatable) question.

To elaborate on my thoughts about this: I think they are **both** correct. (ie, the "inconsistency" simply isn't an issue.)

The real and true reason for the inconsistency is the difference between shortcut handling and mouse handling chronologically.

Hitting a shortcut key causes the following (in 1.3 logic):

1. Shortcut key goes does
2. Button visually goes down (value=1)
3. Callback is invoked (_while value=1_)
4. 0.15 seconds later, button visually goes up (value=0)
5. It doesn't matter much when the shortcut key goes up (ignoring repeats)

Whereas for mouse presses:

1. Mouse button goes down
2. Button visually goes down (value=1)
3. Mouse button goes up
4. Button visually goes up (value=0)
5. Callback is invoked (_while value=0_)

The key point is that keyboard shortcuts invoke the callback _as soon as the button goes down_ but mouse presses invoke the callback _as soon as the button goes up_.

I think this is actually fine and there's no reason to change anything. If we _really_ wanted to resolve the inconsistency, I described two alternatives in the description of this PR, but I think they are both worse than the current flow.

MatthiasWM:
The visual feedback is really important. I was not aware that my patch suppressed that. And for the state values, I find it best to get back to what 1.3 did at this point. `value` is essential for check boxes and radio and toggle buttons. On push buttons, it has really not much of a meaning.

dannye:
Thanks @MatthiasWM! FLTK 1.4.1 is going to be the most rock solid version yet! I'm very excited! Thank you all so much.

PR #1150, opened 2024-12-01
Use a constant frame rate (25 fps) in test/cube demo

Albrecht-S:
The old version was "synchronized" by calling Fl::check() in a loop which resulted in the highest speed possible on any platform.

It turned out that this was way too fast on the Wayland platform as `valerius` posted on IRC/Matrix on Nov 30, 2024. Citation: "anyone else notice the cube demo on Wayland is mega ultra hyper speed?".

I took the opportunity to change the demo to use a timer with a given frame rate (see constant 'FPS') to synchronize drawing. Note: the 'speed' slider is independent of the drawing frequency.

Finally this commit adds statistics output when the 'Stats / Exit' button is pressed like

  Iterations =  304, runtime = 12.167 sec, fps = 24.99, requested: 25

with the requested frame rate (default = 25) and allows the user to continue to see more statistics.

To do (very likely not in 1.4.x):
 1. integrate running statistics in the display
 2. make the desired frame rate variable at runtime (new slider)

### DIFF
--- a/test/cube.cxx
+++ b/test/cube.cxx
@@ -16,7 +16,18 @@
 //     https://www.fltk.org/bugs.php
 //
 
-#include <config.h>
+// Define 'HAVE_GL=1' on the compiler commandline to build this program
+// w/o 'config.h' (needs FLTK lib with GL).
+//   $ fltk-config --use-gl --compile cube.cxx -DHAVE_GL=1
+// Use '-DHAVE_GL=0' to build and test w/o OpenGL support.
+
+#ifndef HAVE_GL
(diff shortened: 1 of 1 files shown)


MatthiasWM:
IIRC test/glpuzzle has a very similar problem.

Albrecht-S:
> IIRC test/glpuzzle has a very similar problem.

Very likely. Making the puzzle "spin" by dragging (possibly unintentionally) seems to be very platform dependent. Also, the "solve" function could be much slower so a user can follow it...

But I'm not going to work on that one, at least not in the near future.

PR #1143, opened 2024-11-24
Windows: Fix rescale bug while window is maximized or fullscreen

dannye:
From #1141

After lots of testing I feel very confident about the fix, so I decided to open a pull request.

Feel free to suggest improvements if you can think of issues that I'm not spotting.

Before:

https://github.com/user-attachments/assets/14598699-8dc6-4bc0-9129-42534ddc70e6

After:

https://github.com/user-attachments/assets/1779311b-ffc8-4f55-8141-ed46c79162e6

PR #1160, opened 2024-12-09
Prevent clicking inactive menubar item from opening menu window - refactor

### LINKED ISSUE #1159: Mouse hover+Enter key selects inactive menu items
**Describe the bug**

Hovering over an inactive menu item and pressing Enter picks the item and invokes its callback.

I have no idea how I've never noticed this in over two years.

This means that marking a menu item as FL_MENU_INACTIVE is merely a suggestion to the user.

**To Reproduce**

1. Run test/menubar
2. Click on any menubar item (even an inactive one!)
3. Hover over an inactive menu item (in the menubar or in a submenu)
4. Press Enter
5. See that the menu item was picked and its callback was invoked

**Expected behavior**

Pressing Enter while hovering over an inactive menu item should do nothing.

**Video**

https://github.com/user-attachments/assets/680399de-0d8d-41e7-b9e0-dd4d78724650

**FLTK Version**
 - Version: 1.4.0
 - Downloaded and self-built from: Git
 - If from Git, branch: master
 - If from Git, commit: 5ff42dd

 - Default CMake options

 - OS: Windows
 - OS Version: Windows 11

**Additional context**

As you could expect, this can be very destructive for applications that assume inactive menu items will never be picked.

Here's an amusing example from my editor where a whole song can be deleted _while it's playing_, which is absolutely not supposed to be possible.

https://github.com/user-attachments/assets/e74e9f8e-791f-4d33-b17e-2f8369808f17

Many of my menu item callbacks have safeguards that return early if conditions don't make sense, but not all of them do. In some cases, the program crashes.

dannye:
~~Fixes #1159~~

### DIFF
--- a/src/Fl_Menu.cxx
+++ b/src/Fl_Menu.cxx
@@ -1002,7 +1002,7 @@ const Fl_Menu_Item* Fl_Menu_Item::pulldown(
   pp.current_item = 0; pp.menu_number = 0; pp.item_number = -1;
   if (menubar) {
     // find the initial menu
-    if (!mw.handle(FL_DRAG)) {
+    if (!mw.handle(FL_DRAG) || (pp.current_item && !pp.current_item->activevisible())) {
       Fl::grab(0);
       return 0;
     }
--- a/src/Fl_Menu.cxx
+++ b/src/Fl_Menu.cxx
@@ -842,6 +842,7 @@ int menuwindow::handle_part1(int e) {
       if (pp.current_item && (!pp.menubar || pp.menu_number > 0) &&
           pp.current_item->activevisible() && pp.current_item->submenu() && !pp.current_item->callback_)
         goto RIGHT;
+      if (pp.current_item && !pp.current_item->activevisible()) return 0;
       pp.state = DONE_STATE;
       return 1;
     case FL_Escape:


dannye:
I would also like to fix the fact that clicking a top-level inactive menubar item still "pulls down" the menu despite there being no menu to show. The non-existent menu window still becomes the Fl::grab(), so the main window cannot be interacted with, but there is no visual indicator of this, which is awfully confusing.

But fixing this seems fairly more complicated. When a menubar is pulled down, the initial menu item is chosen with a fake FL_DRAG event:

https://github.com/fltk/fltk/blob/5ff42dd37a5fa071f54a6e413e349f26f4837340/src/Fl_Menu.cxx#L1004

So in order to "abort" if the pressed top-level menubar item is inactive, `menuwindow::handle_part1()` would have to return 0.
But I'm not immediately sure how to tell this fake FL_DRAG event apart from a legitimate FL_DRAG event.

If someone else has a good idea for how to fix this particular aspect, that would be really great. But otherwise, I think it's not urgent to fix this aspect because it's mostly harmless (unlike the bug being fixed by this PR).

dannye:
> So in order to "abort" if the pressed top-level menubar item is inactive, `menuwindow::handle_part1()` would have to return 0.

I was overlooking the simple solution. I simply added the "abort" to `pulldown()` directly instead of trying to modify `handle_part1()`.

ManoloFLTK:
@dannye: please consolidate all proposed changes in a single commit relative to the head of the git FLTK repository.

dannye:
I *can* do that, but it doesn't really change anything. If you're considering how it will look when it is merged and want it to be added to the master branch as a single commit, then use the "Squash and merge" option in GitHub. That would have made #1143 added to master as a single commit instead of 2, for example.

dannye:
I updated the title to reflect that the bugs have already been fixed. Feel free to merge or close.

MatthiasWM:
The current changes and the patch are identical in functionality. Thanks for both contributions. I will close this PR at this point.

PR #1167, opened 2024-12-12
Correctly detect ucrt64 environment - CMake 

alexbora:
CMake ver.  3.31.1-1 and 3.31.2-3 (mingw-w64-ucrt-x86_64-cmake), MSYS2/ucrt64, OS win10. 

CMake does not properly detect the ucrt64 environment; it seems to happen apparently when MSYS might not be correctly installed.
CMake will then wrongly include the mingw headers, instead of ucrt64 ones, leading to compilation errors. 

Updated the CMAKE_PREFIX_PATH var to /ucrt64 (instead of /mingw, when ucrt64 detected), re-tested under win10/11 as standalone build as well as a lib included in a CMake project, compiles cleanly. 

Autotools detect the environment w/o issues.

### DIFF
--- a/CMake/setup.cmake
+++ b/CMake/setup.cmake
@@ -172,7 +172,6 @@ if(WIN32)
 
  if(MINGW AND EXISTS /mingw)
     if(DEFINED ENV{MSYSTEM} AND $ENV{MSYSTEM} STREQUAL "UCRT64") # UCRT64
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_WIN32") # suppress compiler warnings
       list(APPEND CMAKE_PREFIX_PATH "$ENV{UCRT64}")
     else()
      list(APPEND CMAKE_PREFIX_PATH /mingw)


Albrecht-S:
Thanks for the report and the PR, I'm working on it ...

Albrecht-S:
@alexbora Feedback (still working and testing more cases): your code
```CMake
    if(DEFINED ENV{MSYSTEM} AND $ENV{MSYSTEM} STREQUAL "UCRT64") # UCRT64
      list(APPEND CMAKE_PREFIX_PATH "$ENV{UCRT64}")
    else()
     list(APPEND CMAKE_PREFIX_PATH /mingw)
    endif() # UCRT64
```
is not correct. It works only by accident.
This was likely a typo:
```
  list(APPEND CMAKE_PREFIX_PATH "$ENV{UCRT64}")
```
and was maybe intended to read
```
  list(APPEND CMAKE_PREFIX_PATH "$ENV{MSYSTEM}")
```
but this is also incorrect. It would add `UCRT64` to `CMAKE_PREFIX_PATH` which is also wrong. A better statement would be:
```
  list(APPEND CMAKE_PREFIX_PATH "/ucrt64")
```

**_Why does your patch "work" anyway?_** The old code would add "/mingw" to the `CMAKE_PREFIX_PATH` which finds the **wrong headers**, your code would add an empty string to `CMAKE_PREFIX_PATH` which is harmless, the correct headers are found anyway. Weird, isn't it?

Well, this led me to think about the addition of "/mingw" in the old code.
- Why was it done in the first place? I don't know (this was done before I took over).
- Is it correct? I don't think so.

According to [MSYS docs on "Environments"](https://www.msys2.org/docs/environments/) it should likely be "`/mingw64`" or, in "legacy" mode "`/mingw32`". Basically we'd have to test for all possible "environments" and add the corresponding "prefix".

But very likely it would be the best not to add anything to `CMAKE_PREFIX_PATH` **in all cases of MSYS "Environments"**, hence we could probably remove the entire code that adds something to `CMAKE_PREFIX_PATH` - which would simplify the code rather than adding more complexity. Sigh. Windows. ;-)

That's why I'm still investigating...

Albrecht-S:
@alexbora What happens if you remove the assignment of "`/mingw`" on your system instead of adding another one of "`/ucrt64?`" - or if you comment out both `list(APPEND CMAKE_PREFIX_PATH xxxx)` statements? Does it work with your `UCRT64` setup?

I'll try this here as well but I'd like to know if it works for you (I think it should). Thanks in advance.

Albrecht-S:
@alexbora Welcome, and thank you for your bug report, tests, and feedback. We definitely need user feedback for such border cases.

I'll test this more tomorrow, maybe with other "MSYS2 Environments" as well, and then I'll decide. It's very likely that the best solution is to remove that statement (or of course the entire code block).

Albrecht-S:
Note: I tested several configurations and I conclude that the best solution is to remove the faulty assignment altogether as written above. Tested with MinGW (aka mingw32), MSYS/MinGW-w64, and MSYS/MinGW-w64/UCRT64. All configurations build and work well.

Thanks for the PR and the heads-up nevertheless and for your testing and feedback.

Closing this PR now because the committed solution is different than this PR.

PR #1171, opened 2024-12-15
gears: Add gears demo

carmiker:
This adds a simple port of the "gears" demo to FLTK. This is useful for testing GL functionality and refresh rate and/or redraw interval.


MatthiasWM:
We already have several OpenGL demos that do very similar things. This is also a well known GLUT app that works as is in its original form. We also already have demos for that.

This particular version mixes FLTK and GLUT which is possible, but I would not recommend that. Lastly, this version uses the idle callback to draw the OpenGL graphics directly which is not a good idea. Recommended implementation would derive Fl_GL_Window and implement a custom draw() function.

Sorry, but I'll have to reject this PR.

PR #1178, opened 2024-12-23
Win32: handle WM_SYSCOLORCHANGE

CyprinusCarpio:
This simple PR ensures that FLTK programs update their colors when the system color theme is changed on Windows.

Perhaps, a developer may use `show()` instead of `show(argc, argv)` for the top window with the intent of using the default FLTK colors. As per the documentation of `show(argc, argv)` in `Fl_Window.cxx` this isn't correct:

> This call should be used for top-level windows, at least for the first (main) window.

Nonetheless, in such a case the new `WM_SYSCOLORCHANGE` handler may be at odds with that decision.

Also, what if the color theme is changeable in the program itself, and the `WM_SYSCOLORCHANGE` handler does more harm than good?

### DIFF
--- a/src/Fl_win32.cxx
+++ b/src/Fl_win32.cxx
@@ -1220,6 +1220,18 @@ static LRESULT CALLBACK WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lPar
   if (window) {
     switch (uMsg) {
 
+      case WM_SYSCOLORCHANGE: {
+        Fl::get_system_colors(); // For selection color
+        DWORD bg = GetSysColor(COLOR_BTNFACE);
+        Fl::background(uchar(bg&255), uchar(bg>>8), uchar(bg>>16));
+        DWORD bg2 = GetSysColor(COLOR_WINDOW);
+        Fl::background2(uchar(bg2&255), uchar(bg2>>8), uchar(bg2>>16));
+        DWORD fg = GetSysColor(COLOR_WINDOWTEXT);
+        Fl::foreground(uchar(fg&255), uchar(fg>>8), uchar(fg>>16));
+        Fl::redraw();
+        return 0;
+      }
+
       case WM_DPICHANGED: { // 0x02E0, after display re-scaling and followed by WM_DISPLAYCHANGE
         if (is_dpi_aware && !Fl_WinAPI_Window_Driver::data_for_resize_window_between_screens_.busy) {
           RECT r, *lParam_rect = (RECT*)lParam;


Albrecht-S:
@CyprinusCarpio Thanks for your PR with this Windows specific code.

This will be considered for FLTK 1.5 but it can't be merged in FLTK 1.4 for several reasons:

We'll (try to) implement "_themes_" and _dark_ and _light mode_ in FLTK 1.5 and this PR can perhaps be a part of the final implementation.

All the issues you mention above are (some of the) reasons why this "simple PR" can't be merged as-is. If we merged it users (i.e. program developers) wouldn't have control over what happens when Windows posts the `WM_SYSCOLORCHANGE` message to the application. We need something to enable the developer of a program to control this behavior.

Another reason is that we can't make such changes (like adding "themes" support) only to Windows. FLTK is a cross-platform toolkit and needs similar support of new features on all platforms.

There's another patch (branch) pending to support dark mode on macOS. I tested it and found that there is much more needed to make this work in a defined way than only a "simple PR".

But anyway, thanks for your work and for sharing the code.

CyprinusCarpio:
Thanks for the input.

I see, this isn't the optimal solution, at least not in it's current form. 

> Another reason is that we can't make such changes (like adding "themes" support) only to Windows. FLTK is a cross-platform toolkit and needs similar support of new features on all platforms.

As far as support for dark/light mode I agree, but `WM_SYSCOLORCHANGE` is not the same. What this PR achieves is really just listening for system color changes. This isn't any more in conflict with cross-platform design of FLTK than getting the Win32 system colors using `GetSysColor` in `Fl_WinAPI_Screen_Driver.cxx`.

Granted, more finesse is needed than I did in this PR to make it work properly, and that finesse would probably need to come from a solid theming foundation in FLTK itself, which indeed needs to be cross-platform. I don't know what the 1.5 theming support plans are, but I think there needs to be a way to tell between the "default" "system" theme and one set by the program, so that one of the problems of this PR is avoided, that is unwanted changing of app colors in a `WM_SYSCOLORCHANGE`. I have some more ideas on theming, but they go beyond the scope of this PR.

If anyone has more input on this issue, that is applicable to the way `WM_SYSCOLORCHANGE` should be handled, please share. As for me, I think that this PR can be closed.

dannye:
@CyprinusCarpio Can you elaborate on how to trigger a `WM_SYSCOLORCHANGE` event? I tried tinkering with this but couldn't make the event happen on Windows 11.

Under Settings -> Personalization -> Colors, neither "mode" (light vs dark) nor "accent color" seem to trigger `WM_SYSCOLORCHANGE`. Although, it did trigger a `WM_SETTINGCHANGE` event. Just curious.

===

Also, as a partially on-topic/partially off-topic note, when FLTK does support proper event handling for dark mode toggling, certain widgets will need to be made to handle the event as well. For example, Fl_Help_View will need to recompute its hyperlink text color.

Currently, since dark mode toggling is the responsibility of the application, the application code must ensure that Fl_Help_View's `format()` function gets called when the theme is changed to guarantee that hyperlinks have proper contrast.

However, the `format()` function is private, and so it must be called in a roundabout way, like:
`help_view->textsize(help_view->textsize());`

[Example](https://github.com/dannye/crystal-tracker/commit/b8bb9c5246445beb609604175fc619a22941b8f0)

I didn't check if this is handled already by @Albrecht-S's experimental/WIP darkmode branch.

MatthiasWM:
This is quite a niche request. For the time being, if this functionality is important to you, you can add your own handler with `Fl::add_handler()` or `Fl::add_system_handler()`. https://www.fltk.org/doc-1.4/group__fl__events.html#ga50d1df4bc40158be2514f120f2f7e60c . 

I will look at this for 1.5, but as you already said, if an app developer does not call `show(args, argv)`, or if a user chose colors via command line, this event should be ignored. 

This issue has nothing to do with dark mode. That's a different can of worms ;-) .

CyprinusCarpio:
> Can you elaborate on how to trigger a `WM_SYSCOLORCHANGE` event? I tried tinkering with this but couldn't make the event happen on Windows 11.

I should have mentioned, `WM_SYSCOLORCHANGE` was *propably* deprecated some time after Win7. I don't know about 8, but on 10 and 11 it appears as if the only way to trigger this message is to use third-party theming tools, such as those used to (hack-ishly?) enable the Classic theme. Overall, Windows 10 and 11 have 3 themes, Classic (or more accurately, lack of theme, deprecated a long time ago, requires hooking explorer.exe for it not to crash), Basic (related to the Luna XP theme, is used when DWM is not running), and Aero (the default, uses DWM). 

> For the time being, if this functionality is important to you, you can add your own handler with `Fl::add_handler()` or `Fl::add_system_handler()`

This is a good solution, I'll make sure to do it

CyprinusCarpio:
Closing, it will be easier to implement what this PR achieves in a more fleshed out commit than try to merge this one.

PR #1187, opened 2025-01-11
Fix Fl_Table_Row crashes (1.3.x branch, possibly 1.4.x)

basiliscos:
Hello,

I faced with the following issue: sometimes when a new `Fl_Table_Row` is created, then new rows to it are assigned (i.e. via method `rows(int)` ).  In the current implementation,  it forward the rows assignment to the parent class, and only after that it allocates internal structures for the selected rows. 

The problem is that parent class (`Fl_Table`) in its `rows(int)` implementation might trigger some events processing, right after assigning new rows number... and the events processing in the `Fl_Table_Row` assume the internal structures are already prepared (allocated), while they are not.  Which leads to segfault.

My patch fixes that. Also, I think the bug  exits on 1.4.x branch too.

I attach the screenshots with backtraces of segfault. 

I tested my patch, and it works fine.

Thank you!

2025-01-11_xx

### DIFF
--- a/src/Fl_Table_Row.cxx
+++ b/src/Fl_Table_Row.cxx
@@ -150,9 +150,9 @@ void Fl_Table_Row::select_all_rows(int flag) {
 
 // Set number of rows
 void Fl_Table_Row::rows(int val) {
-  Fl_Table::rows(val);
   while ( val > (int)_rowselect.size() ) { _rowselect.push_back(0); }	// enlarge
   while ( val < (int)_rowselect.size() ) { _rowselect.pop_back(); }	// shrink
+  Fl_Table::rows(val);
 }
 
 //#define DEBUG 1


Albrecht-S:
@basiliscos Thank you for the patch. This is something @erco77 (the original author) should take a look at.

Is it possible that you provide a simple but complete (compileable and linkable) demo source code that shows the issue? This would help testing and verifying that your patch is OK. TIA.

basiliscos:
>Is it possible that you provide a simple but complete (compileable and linkable) demo source code that shows the issue? 

Unfortunately, no.

I just can add, that the issue happens on `Windows`, and does not on Linux.

erco77:
I'll have to look at this when my head is clear; while the fires in Los Angeles are mostly out, there's new wind threats through Tuesday, so we may have to go through this again.

Anyway, offhand I want to look further into this, as changing the order of calling rows() seems fragile. There's probably a more robust way to solve for interference from the parent class during the method's execution.

Albrecht-S:
@erco77 Hi Greg, thanks for looking into this. Meanwhile I took a look as well and tried to figure out what's happening. Unfortunately I couldn't reproduce the issue so far...

That said, since I don't **know** what is happening, here are some thoughts:

Looking at the patch and **_assuming it's on the right track_** I would at least suggest a modification like this:
```C++
void Fl_Table_Row::rows(int val) {
  while ( val > (int)_rowselect.size() ) { _rowselect.push_back(0); }	// enlarge
  Fl_Table::rows(val);
  while ( val < (int)_rowselect.size() ) { _rowselect.pop_back(); }	// shrink
}
```
This alternate patch would **enlarge** the `_rowselect` array **_before_** `Fl_Table::rows(val)` is called and **shrink** the `_rowselect` array **_after_** `Fl_Table::rows(val)` is called if necessary.

I'm not convinced that this is the correct / best solution though, but anyway, I hope it helps.

That said, the stack trace reveals that the OP's (@basiliscos) own derived class calls `Fl_Table_Row::row_selected(0)` in the top-most stack entry (`0`) whereas the row assignment is called in stack entry `19`. The former is obviously caused by an `FL_MOVE` event sent to the window (and thus to the derived class(es)) by `fl_fix_focus()` in stack entry `8`. The latter is caused by calling `hide()` on the horizontal scrollbar in `recalc_dimensions()` (Fl_Table.cxx, line 499, stack entry `11`).

Another patch I tried was to harden `Fl_Table_Row::row_selected(int row)`:
```C++
// Is row selected?
int Fl_Table_Row::row_selected(int row) {
  if ( row < 0 || row >= rows() ) return(-1);
  if ( row >= (int)_rowselect.size() ) {
    printf("row_selected(%3d) out of range(_rowselect.size() = %3d)\n", row, (int)_rowselect.size());
    fflush(stdout);
    return -2;
  }
  return(_rowselect[row]);
}
```

This would return `-2` instead of `-1` when `row` is outside the range of `_rowselect`, i.e. in the case the OP wants to fix. Unfortunately I could never see this case in my tests. Note that I tested on Linux (including a build with Address Sanitizer and MinGW (cross-compiled on Linux) but not with a real Windows box.

For my tests I used `examples/table-with-keynav.cxx` with some modifications so I could assign row sizes etc. but w/o "success" (i.e. I couldn't reproduce the crash).

Note also that the OP is referring to 1.3.x but I tried to test with both `branch-1.3` and `master` and I couldn't reproduce the issue.

PS: crossing fingers that no new fires will come up!

Albrecht-S:
@basiliscos Can you describe in more detail when the error happens, what the geometry of your widgets is, and whatever could help to find the real cause. It would be helpful to see more code of your class, particularly the handle() method and the assignment of row sizes. Thanks in advance for all info.

basiliscos:
> Note that I tested on Linux (including a build with Address Sanitizer and MinGW (cross-compiled on Linux) but not with a real Windows box.

me too have not luck on linux (with sanitizer and valgrind). Linux is my main development platform :)

Albrecht-S:
@basiliscos Are you sure that your statement:

> ... the issue happens on Windows, and does not on Linux.

is true, or is it maybe the opposite?

How are you building your program, do you use MinGW or MSYS?

It would also be interesting to see what happens around line `119` in your `static_table.cpp` where `Fl_Table::CONTEXT_RC_RESIZE` is handled by your `draw_cell()` code, or the code of `assign_rows()`. If possible, please post the full code of `static_table.cpp` (and related headers).

basiliscos:
Here is my derived table, where the code triggers segfault



 i.e. it asks the count of rows, and then asks whether each row is selected.
 
 here is all the code: 
 
 https://github.com/basiliscos/syncspirit/blob/v0.4.0-dev/src/ui-fltk/static_table.cpp#L221

Albrecht-S:
What is `mxe` ? Wikipedia says `Methoxetamine (MXE) is a dissociative hallucinogen ...`
I hope that's not the `mxe` you are using. ;-)

Thanks for all further info. I ran out of time and I'm leaving the investigation to Greg (@erco77) for now...

basiliscos:
>  https://github.com/mxe/mxe

yep. I use mingw-compiler from this toolchain instead of the one from my distro.

Albrecht-S:
@basiliscos Independent of this issue I suggest to modify your code posted [here](https://github.com/basiliscos/syncspirit/blob/810fb3e5fcb6e093d4d20d4e06b221ad2028fb43/src/ui-fltk/static_table.cpp#L219): This code is executed for **every** event, even for `FL_MOVE` events which don't change anything of your widget. This appears to be wasteful, there should be a better opportunity to achieve what you want.

However, as it stands, this code is what triggers the issue which maybe needs a fix, hence thank you for reporting and trying to fix it.

basiliscos:
@Albrecht-S Thank you for your suggestions )

Albrecht-S:
@erco77 While we're at it: the documentation of `int Fl_Table_Row::row_selected( int row	)`
> Checks to see if 'row' is selected.
> Returns 1 if selected, 0 if not.

is misleading or at least incomplete. In fact this method returns `-1` if `row` is out of range as stated in a [comment of the implementation](https://github.com/fltk/fltk/blob/06418e4b219add0ff31186effbdd01ecb0f00a28/FL/Fl_Table_Row.H#L165) in the header file:
> // is row selected? (0=no, 1=yes, -1=range err)

Unfortunately code like [the OP's code here](https://github.com/basiliscos/syncspirit/blob/810fb3e5fcb6e093d4d20d4e06b221ad2028fb43/src/ui-fltk/static_table.cpp#L221) interprets the result `-1` as `true` which is another cause of the issue at hand. BTW, I made the same "mistake" as well in my own tests by using the result as `bool`.

That said, we can either document this fact - but this doesn't fix existing code - or we can fix the code to do what the current doc says: return **only** `1` or `0` so interpreting the result like a `bool` works correctly.

I suggest strongly to do the latter (fix the code rather than the documentation).

erco77:
> I suggest strongly to do the latter (fix the FLTK code rather than the documentation).

Hmm, ok, looks like `Fl_Browser::selected(int line)` returns 0 if '`line`' is out of range, so that would at least be consistent.

Albrecht-S:
> looks like Fl_Browser::selected(int line) returns 0 if 'line' is out of range, so that would at least be consistent.

Good, thanks for checking. Since it wasn't documented we can IMHO change it anyway to do what the docs say. If we ever wanted to distinguish the error code we could add another return parameter to the method but this shouldn't be necessary. User code should not call this method with out-of-bounds values anyway, so returning 0 for invalid rows should be fine.

Albrecht-S:
@erco77 I took a stab at fixing the docs and the related code as discussed above in commit a48ebc5db02960165bf1b9fb64443a5cdf9635d5. See the commit log for changes. I apologize if this commit affects your work.

#### Details:
This commit changes the return value of `Fl_Table_Row::row_selected(int)` (only 0 and 1) and improves (clarifies) the doxygen docs.

My commit does **_not_** fix the potential "out-of-sync behavior" of `Fl_Table::rows()` and the size of the `Fl_Table_Row::_rowselect` array. I'm leaving the decision to you but I think my suggested modification of this PR (the first patch in [this comment](https://github.com/fltk/fltk/pull/1187#issuecomment-2587698200)) could be a good start or at least a minimal fix.

My commit **assumes** that the above mentioned "out-of-sync behavior" is fixed and thus `rows()` and the size of the `Fl_Table_Row::_rowselect` array are always in sync. Therefore I did not include the "hardening" of `int Fl_Table_Row::row_selected(int row)` as shown in the second patch of [this comment](https://github.com/fltk/fltk/pull/1187#issuecomment-2587698200). If you think it is necessary then it should of course return 0 and not `-2` as in my previous patch used for testing.

PS: I'd appreciate a quick decision since FLTK 1.4.2 is due and should be released soon to make it into Debian 13 (Trixie). But I also understand if you are too busy (with real life things) and I can do it for you if you agree. I would use my first suggested patch rather than the original fix of this PR. Please let me know what you think...

MatthiasWM:
Sorry for having been away for a while. I think that @Albrecht-S 's solution looks good. @erco77 , die you have time to verify?

erco77:
Sorry guys, I'm just worlds away in electronics and deadlines. I guess just go ahead with the patch; I don't think there's anything too complex here.

Albrecht-S:
OK, thanks, @MatthiasWM and @erco77, I assigned myself and I'll go ahead and push the patch the OP proposed with my changes). This - together with the previous commit a48ebc5db02960165bf1b9fb64443a5cdf9635d5 - will then be backported to branch-1.3 (1.3.11), the next and final release of branch-1.3 (hopefully, i.e. unless something really unforeseen happens).

Albrecht-S:
Done. Closing this PR now. The modified fix has been applied to `master` (commit 0c2aa40bbc37c7e8b236e6db7c2f07b90f619e16, will be in 1.4.2) and backported to branch-1.3 (commit 93830563e9ce976ca46f056f69e4155d9cde1243, will be in 1.3.11).

@basiliscos Thanks for your PR, I'm closing it now w/o merge because the applied fix is different.

basiliscos:
@Albrecht-S thank you. Is there any terms of the fresh 1.3.x release with the included fix?

Albrecht-S:
I'm not sure yet if we need another backport or fix for issue #1192, hence I'd like to wait some time until this has been clarified. Making a release is a non-negligible overhead, and the current goal is to finish the 1.3 branch with the next (1.3.11), hopefully final release. That's why I'm still hesitating, but it should be done ASAP.

ManoloFLTK:
@Albrecht-S About whether another backport or fix for issue #1192 is needed: my view is that the proposal given at #1192 to write slightly differently and in a platform-independent way the user code that turns on or off a multi-screen window solves completely that issue without need to change the source code of FLTK 1.3 nor 1.4. But I'd like to see the OP confirm that.

PR #1221, opened 2025-03-15
Fixed pointer overflow on international characters like Kanji

ggarra13:
On Windows there can be an overflow in font names when using Fl::set_fonts("-*").

For the whole thread that kept my program failing, see:

https://github.com/ggarra13/mrv2/issues/321

### DIFF
--- a/src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx
@@ -1347,13 +1347,20 @@ bool Fl_Wayland_Window_Driver::process_menu_or_tooltip(struct wld_window *new_wi
     xdg_positioner_set_anchor_rect(positioner, popup_x, 0, 1, 1);
     popup_y++;
   }
-  int V, screen_H;
-  Fl::screen_work_area(V, V, V, screen_H, origin_win->screen_num());
-  // Under KWIN, we need to have in screen_H the size of the current screen's work-area.
-  // We get this exactly only when there's a single screen. Otherwise, we get the screen height.
-  bool tall_kde = (pWindow->h() > screen_H) &&
-    (Fl_Wayland_Screen_Driver::compositor == Fl_Wayland_Screen_Driver::KWIN);
-  xdg_positioner_set_size(positioner, pWindow->w() * f , (tall_kde ? screen_H : pWindow->h()) * f );
+  int positioner_H = pWindow->h();
+  if (Fl_Wayland_Screen_Driver::compositor == Fl_Wayland_Screen_Driver::KWIN) {
+    // Under KWIN, limiting the height of the positioner to the work area height
+    // results in tall popup windows starting at the top of the screen, what we want.
+    // Unfortunately, we know the work area height exactly only for single-screen systems,
+    // otherwise FLTK returns work area height == screen height. In that case we estimate
+    // work area height ≈ screen height - 40.
+    int V, work_area_H, screen_H;
+    Fl::screen_work_area(V, V, V, work_area_H, origin_win->screen_num());
+    Fl::screen_xywh(V, V, V, screen_H, origin_win->screen_num());
+    if (work_area_H == screen_H) work_area_H -= 40;
+    if (positioner_H > work_area_H) positioner_H = work_area_H;
+  }
+  xdg_positioner_set_size(positioner, pWindow->w() * f , positioner_H * f );
   xdg_positioner_set_anchor(positioner, XDG_POSITIONER_ANCHOR_BOTTOM_LEFT);
   xdg_positioner_set_gravity(positioner, XDG_POSITIONER_GRAVITY_BOTTOM_RIGHT);
   // prevent menuwindow from expanding beyond display limits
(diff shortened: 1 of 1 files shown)


Albrecht-S:
@ggarra13 Thanks for the PR. I see what you want to achieve but I'm still investigating whether your patch is correct, and I suspect that it is not (although it may fix the issue). I need to dive deeper into these Windows specific functions and structures.

_Note to devs: please don't merge this as-is before I concluded my checks!_

Side note: this code looks like something that asks for using std::string in a somewhat larger patch. Maybe.

Albrecht-S:
@ggarra13 Can you give me a hint on how to reproduce the issue?
Do you know some specific font names that cause this crash?
If you do, pasting the (UTF-8) font name here could likely be helpful.

ggarra13:
> Can you give me a hint on how to reproduce the issue?

Sadly, no.  I have one main user with the issue happening, and three others with similar issues.  Switching to a Windows' locale does not make the issue happen as you need to have fonts like Kanji, etc. which are multibyte (2 or 3 bytes) and make them overflow the buffer.

> Do you know some specific font names that cause this crash?

Let me create a test program that prints all the fonts on their system to give it to the users that had the crashes.

Albrecht-S:
> Let me create a test program that prints all the fonts on their system to give it to the users that had the crashes.

Thanks for the offer, but please wait a moment. I'm currently investigating and maybe I don't need that effort of you and your user...

ggarra13:
I think constraining it to LF_FACESIZE as done first is correct.  See:

https://groups.google.com/g/microsoft.public.vc.language/c/RfqKiBg-OE4

The strncpy() is superflous, but cannot hurt.

Albrecht-S:
> I think constraining it to LF_FACESIZE as done first is correct. ...

This is the wrong place to fix it, the bug is further down in the code when allocating (on the stack):
`  char buffer[LF_FACESIZE + 1];`

`LF_FACESIZE` is a constant (32) and is the max. length of `wchar`s a font name can have. Assume an "Eastern" font name has only characters in the respective Unicode encoding. Let's further assume that each "Eastern" Unicode character is converted to three or four UTF-8 bytes, then the required buffer size would be `LF_FACESIZE * 4 + 1` and not `LF_FACESIZE + 1` which is way too small. This is from the times before we used UTF-8 where one character of the original fontname was also one byte in the allocated buffer.

Please try my patch, applied to the original code. It is much bigger than only allocating a larger buffer because I also reformatted and optimized it a bit to avoid unnecessary string copies. And, BTW, no, I didn't use `std::string` in this patch.
[win-enumfonts.diff.txt](https://github.com/user-attachments/files/19264550/win-enumfonts.diff.txt)

This patch includes some debug statements. I used `test/utf8.cxx` for my tests, but in my environment I didn't have fonts that would cause an overflow. However, I have some fontnames that are up to the max. (31 characters) length. I built FLTK in VS in debug mode and when you run from the debugger you should see something like this for all font names **longer than 30 characters**:
```
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift Light SemiCondense"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiLight SemiCond"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiBold SemiConde"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiLight Condense"
```

The important output is `dstlen=31` (never higher **_in my tests_**). If you have a fontname that causes an overflow in the original code you should see a value of `dstlen > 31`, and since the buffer (`char *fn` in the new code) is now `malloc`ed in the correct size there shouldn't be an overflow.

I'd appreciate if you could test and post the output here. If you could reproduce the issue before, then you should see some lines with `dstlen > 31`.

ggarra13:
I checked your code and it is indeed a more correct fix.  You really know your Unicode!

Albrecht-S:
@ggarra13 OK, Gonzalo, it's fine that you agree that my fix is more correct, but did you also test it, and did you see `dstlen > 31` ? I'm always striving to see a **positive** confirmation of (a) the bug and (b) the fix. Running any program that exhibited the bug (either yours of test/utf-8) would be a confirmation for me. TIA.

ggarra13:
Okay.  Will do.  I cannot reproduce it on my machine as I don't have fonts > 31.  I am trying to patch my FLTK with cmake to submit my program to those with the problem.
You'll have to be patient.

Albrecht-S:
@ggarra13 Well, I think it's not necessary to bother your users with such tests. I'll make some more tests, commit my solution, and let you know. You can then decide how to proceed - or you do as you suggested above if you also want to do this ...

Please post here if you have news, I'll be noticed. We don't need to reopen.

ggarra13:
Here's the output with your patch unchanged (ie 32 - 2).

$ mrv2
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift Light SemiCondense"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiLight SemiCond"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiBold SemiConde"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Bahnschrift SemiLight Condense"
enumcbw: dstlen=30, LF_FACESIZE=32, L=29, ANSI=Y fontname="Bahnschrift SemiBold Condense"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Small Semili"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Small Semibo"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Text Semilig"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Text Semibol"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Display Ligh"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Display Semi"
enumcbw: dstlen=31, LF_FACESIZE=32, L=30, ANSI=Y fontname="Segoe UI Variable Display Semi"
        [main] mrv2 v1.3.6 - Built Mar 15 2025 15:34:37 - Release
        [main] Set Language to en.UTF-8, Numbers to C
        [main] Build Environment:
        [main]  Distribution: Microsoft Windows 11 Pro
        [main]  Desktop Environment: Windows (GDI+)
        [main]  Kernel Info: Microsoft Windows [Version 10.0.26100.3476]
        [main] Running Environment:
        [main]  Distribution: Microsoft Windows 11 Pro
        [main]  Desktop Environment: Windows (GDI+)
        [main]  Kernel Info: Microsoft Windows [Version 10.0.26100.3476]
        [main] Install Location:
        [main]  D:/code/applications/mrv2/BUILD-MSys-amd64/Release/install
        [main] Preferences Location:
        [main]  C:\msys64\home\User-PC/.filmaura/
        [ocio] Setting OCIO config to default:
        [ocio]  ocio://default

Albrecht-S:
OK, thanks.

Albrecht-S:
Fixed in commit 2e1730d2f0f131d792c9b0561672f5252a868785. Feedback appreciated.

ggarra13:
I've given your patch to the users compiled in my program and they report that it does not work.  However, they report my original patch does work fine.

Albrecht-S:
Yes, the patch was different, technically, but it should do the same thing.

PR #1222, opened 2025-03-16
Fluid Rejuvenation for 1.5

MatthiasWM:
Hello Fluid users, hi maintainers and developers.

I mentioned many times that Fluid has grown to be a pretty cool app full of features over the last two decades. Much was added without plan or documentation. With the start of FLTK 1.5 and the introduction of C++11, it is high time to finally prune, refactor, streamline, declutter, rearchitect and rejuvenate Fluid.

The goal is to make Fluid better maintainable and better documented, so current features become bug-free, and new features can be added without guess work and cross referencing all over the place.

Here is a list of things in this PR:
- file size: some source files had close to 4000 lines. I have started to split files into functional units and more files
- file list: I have added a hierarchy into the source code directory to make it easier to focus on features instead of a long list of files
- naming conventions: som classes started with `Fl_`, some with `Fluid_`, some with `Fd_`. I decided to remove these and instead put classes into a two level namespace that often reflects the directory structure and functional structure
- removing all globals: Fluid is packed with "states" that are stored in obscure global variables. I started to move all these into the Application class, Project class,  or wherever they make sense locally.
- multiple document interface: it's far from ready, but in the long run, Fluid should be able to have multiple documents open at the same time. This not only helps the user, but also helps us developers to separate application features from project features
- general naming: I tried to use more common nouns for class and method names to make it more obvious what they do. "Types" are now "Nodes", "types" are now "prototypes", and a bunch more
- C++11: `FL_OVERRIDE` is now `override`. `0`, `0L`, and `NULL` are now `nullptr` where appropriate. I am also slowly introducing std library containers and std::string where that makes sense. `enum`s become `enum classes if useful.

I am planning to leave this PR here for a few days, just in case anyone wants to look at the code and comment. I would like to keep this run going though, and I may add changes here, or merge the PR into master at any point.


MatthiasWM:
OK, I am quite happy with all the changes to Fluid. Most changes just move stuff around and rename things. Some minor changes introduce C++11. Everything compiles on macOS without warning despite "pedantic" and other flags set. So fingers crossed and thumbs squeezed that this works for all user configurations out there.

PR #1226, opened 2025-03-17
Reactivating Mergeback functionality.

MatthiasWM:
This is an experimental addition. It's not ready yet, but I put it here in case anyone likes to try.

Mergeback is a new Fluid feature that allows users to edit large parts of the generated C++ code (callbacks and "Code" blocks) and merge the changes back into a running interactive Fluid.

It solves the issue where a user writes a callback function in the widget dialog inside Fluid, compiles the code in the IDE, only to find that there is a typo in the callback. The IDE will present the C++ code, and the typos is quickly fixes, and the app compiles. But back in Fluid, the changes in the C++ file will be overridden when the files are generated again. Mergeback can find changes in the C++ file and read them back into the project.

Future version of Mergeback will do this automatically and eventually even write back changes all the way back into the project file. I have been using Mergeback for a while and found it very helpful.


MatthiasWM:
TODO: add CRC for header file as well!

PR #1240, opened 2025-03-31
options.cmake: Check Threads_FOUND, not CMAKE_HAVE_THREADS_LIBRARY.

ucko:
The former's been available since CMake v2.8 and documented since v3.17; the latter was never documented and quietly went away in CMake v3.24.

This will be helpful in 1.4 as well; please let me know if you'd like a formal PR on that front too.

### DIFF
--- a/CMake/options.cmake
+++ b/CMake/options.cmake
@@ -739,11 +739,11 @@ if(FLTK_USE_PTHREADS)
 
   include(FindThreads)
 
-  if(CMAKE_HAVE_THREADS_LIBRARY)
+  if(Threads_FOUND)
     add_definitions("-D_THREAD_SAFE -D_REENTRANT")
     set(USE_THREADS 1)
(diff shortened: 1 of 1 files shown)


Albrecht-S:
@ucko Merged "manually" in master (8b28e38942ce8b61966da90af644c9948a007eda), hence I'm closing this PR.

> This will be helpful in 1.4 as well; please let me know if you'd like a formal PR on that front too.

Not necessary, thanks. I cherry-picked your commit in branch-1.4 (2e6e94d11857664979f5ecf466f3afbed3a04976).

_**Side note:** I wonder how it "worked" with the old, undocumented variable. On Debian Bookworm I have `cmake 3.25.1` which doesn't set this CMAKE_HAVE_THREADS_LIBRARY variable, as you wrote (I verified it in CMake sources). Strange... But this is moot, I'm not going to dig deeper._

Thank you very much for this patch.

PR #1241, opened 2025-03-31
Use CMake's built-in timestamp formatting.

ucko:
It notably honors SOURCE_DATE_EPOCH if set, making for reproducible output.  For even better reproducibility, use UTC.  (Unlike the date command's output, the result is already always in English.)  Extend this approach to the book, introducing appropriately formatted PDF_DATE and TODAY variables for its PDF metadata and title page respectively and making make_pdf configurable.

1.4 will need some additional changes to the traditional build system; I'll open a separate PR for that branch.


Albrecht-S:
@ucko Great job, thanks. I like the simplification (CMake string(TIMESTAMP ...)) and the additional PDF metadata (checked after build with `pdfinfo`). Awesome.

I also tested setting `SOURCE_DATE_EPOCH` which worked flawlessly, i.e. as intended. However, there's a minor point with this: if someone sets it unintentionally (maybe from a former build of another project), then the build date of the docs gets this date. Therefore it might be useful to enable it explicitly, rather than just relying on an environment variable being set or not. I learned that Debian strives to enable reproducible builds, but here's my question: how do other projects deal with this, particularly with this environment variable?

Please don't misunderstand this, I'm willing to merge your branch w/o changes, it's just for my curiosity.

ucko:
Fair question. `SOURCE_DATE_EPOCH` settings are typically transient and automatic; for instance, Debian builds canonically go through `dpkg- buildpackage`, which sets the variable for its children (whose descendants then pick it up) on the basis of `debian/changelog`.

Albrecht-S:
@ucko Thanks again, merged now in master (678c85027294c0ae5162d02b20a59fde758ac77a).

I'm now going to work on #1242 for branch-1.4 ...

Issue #1238, opened 2025-03-28
fully support using own shared libraries internally


ucko:
**Is your feature request related to a problem? Please describe.**
FLTK's build system makes minimal use of its own shared libraries, leading to excess disk usage for anyone who will have the shared libraries around anyway.  The Autoconf-based build system offers alternative shared versions of `fluid` and `fltk-options`, but the CMake build system doesn't even do that, and both link all other executables (including the games) statically.

**I am using FLTK**
 - 1.4.x.

**Describe the solution you'd like**
A configuration option to request not merely building static libraries but moreover linking FLTK's own executables against them.

**Describe alternatives you've considered**

**Additional context**
AIUI, it's best practice in Debian to make maximal use of shared libraries (with an exception for system recovery tools that obviously doesn't apply here).

Albrecht-S:
@ucko Hi Aaron, thanks for taking the time to work on FLTK 1.4 for Debian.

Currently we support linking selected test applications (`cairo_test`, `hello`, `pixmap_browser`, `unittests`, `glpuzzle`, and `shape`) against the shared libraries if they are built, but this was done with testing the build procedure in mind, and not for installation.

I can see your point though regarding `fluid`, `fltk-options`, and the applications that are installed in Debian, which are probably some of the games. All other test programs are only for building and testing and should not affect Debian packages.

I'll investigate to build `fluid` and `fltk-options` (which should be installed as system tools) and I'll see which option we can provide to do this.

I'll also take a look at linking all the games against the shared libs. Should we provide an option to build and install the "games"? Currently our default (CMake) procedure doesn't do that, IIRC.

Please tell us which FLTK applications you intend to install in Debian, and I'll try build alternative versions linked against the shared libs. This will be done exclusively for the CMake build (which you should use anyway), because we don't want to change the configure/Makefiles in FLTK 1.4 anymore.

Please feel free to tell us if you need building any apps linked against the shared libs **_using autoconf_** (although I hope you don't need that).

Note to all readers (I'm aware that Aaron knows this): FLTK 1.5 will no longer support `configure/Makefiles`.

Albrecht-S:
@MatthiasWM I agree with all your points, but IMHO installing applications system-wide as in Aaron's case justifies this extra effort. This would also apply to all Linux and other distributions.

MatthiasWM:
Never mind my mail. I did not look close enough at the sender and answered for an app dev scenario.

Albrecht-S:
[...]

PR #1242, opened 2025-03-31
Use CMake's built-in timestamp formatting.

ucko:
It notably honors SOURCE_DATE_EPOCH if set, making for reproducible output.  For even better reproducibility, use UTC.  (Unlike the date command's output, the result is already always in English.)  Extend this approach to the book, introducing appropriately formatted PDF_DATE and TODAY variables for its PDF metadata and title page respectively and making make_pdf configurable.

Under the traditional build system, don't attempt to handle SOURCE_DATE_EPOCH because date's portable interface is too limited, just supply PDF_DATE and TODAY values corresponding to the current date (and time zone) and use the C (English) locale for month names and abbreviations.


Albrecht-S:
@ucko I took the freedom to fix some minor issues in the configure/Makefiles system and pushed it to your branch.

Now I'm going to merge it...

Albrecht-S:
Merged in commit 6b8409fc6cb4ef4eec93028665b314f6075e13c6.

@ucko Now all your PR's are merged, please let me know if there's more to do to get FLTK 1.4 in Debian Trixie. I still need to work on the "shared libraries" issue (#1238).

As discussed earlier I can release 1.4.3 if this is helpful (I think it is). I would likely need one or two days to prepare the release, but I can already begin updating version numbers [1] if we agree to use 1.4.3 for Debian. I would also want to check for some more backports from master (1.5) that are worth having in 1.4.3...

----
[1] Sigh, the version number management should be simplified, but it's an old issue and on my todo list for 1.5.

ucko:
Thanks, and sorry for slightly fumbling the changes to the `Makefile`.  Please do go ahead with a 1.4.3 release once you've wrapped up #1238.  I meanwhile went ahead with a 1.4.2-based upload, which with any luck will make it out of the `NEW` queue in time for the freeze.

Albrecht-S:
FTR: Issue #1238 has been fixed in 1.5 and backported to 1.4.3.

PR #1248, opened 2025-04-11
Windows: fix "heap-use-after-free" in resize_after_screen_change()

timothytylee:
When a native window is destroyed, make sure the callback to Fl_WinAPI_Window_Driver::resize_after_screen_change() is removed.

### DIFF
--- a/src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
+++ b/src/drivers/WinAPI/Fl_WinAPI_Window_Driver.cxx
@@ -50,6 +50,7 @@ Fl_WinAPI_Window_Driver::Fl_WinAPI_Window_Driver(Fl_Window *win)
 
 Fl_WinAPI_Window_Driver::~Fl_WinAPI_Window_Driver()
 {
+  Fl::remove_timeout(resize_after_screen_change, pWindow);
   if (shape_data_) {
     delete shape_data_->effective_bitmap_;
     delete shape_data_;


timothytylee:
The "heap-use-after-free" error can be triggered under the following conditions:
 * There are two screens with different scaling factors
 * The window is moved from one screen to the other
 * Then the window is closed within 1 second (ie. before the `resize_after_screen_change()` timeout callback is invoked)

timothytylee:
On a separate note, the same bug might also be present for `Fl_X11_Window_Driver::resize_after_screen_change()`, which requires a similar fix.

Albrecht-S:
Thanks for the PR, the fix looks sensible. You're right, X11 would need a similar fix AFAICT.

I also see related code in [src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx, line 1609](https://github.com/fltk/fltk/blob/76e76ba80a10bc547eff9ffb0988337193fdf534/src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx#L1609). However, I'm not sure how this is triggered, I can't find `Fl::add_timeout(..., Fl_Wayland_Window_Driver::resize_after_screen_change)` for Wayland as in the X11 and Windows specific code.

I'd like to wait for a comment from @ManoloFLTK because he is the main author of this stuff and I can't test with two screens on my test system (I could try to arrange something on an older Linux version though if needed...).

ManoloFLTK:
I’ll be away from my desk for 2 weeks.Manolo Le 12 avr. 2025 à 13:19, Albrecht Schlosser ***@***.***> a écrit :﻿
  Thanks for the PR, the fix looks sensible. You're right, X11 would need a similar fix AFAICT.
I also see related code in src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx, line 1609. However, I'm not sure how this is triggered, I can't find Fl::add_timeout(..., Fl_Wayland_Window_Driver::resize_after_screen_change) for Wayland as in the X11 and Windows specific code.
I'd like to wait for a comment from @ManoloFLTK because he is the main author of this stuff and I can't test with two screens on my test system (I could try to arrange something on an older Linux version though if needed...).—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



Albrecht-S left a comment (fltk/fltk#1248)
Thanks for the PR, the fix looks sensible. You're right, X11 would need a similar fix AFAICT.
I also see related code in src/drivers/Wayland/Fl_Wayland_Window_Driver.cxx, line 1609. However, I'm not sure how this is triggered, I can't find Fl::add_timeout(..., Fl_Wayland_Window_Driver::resize_after_screen_change) for Wayland as in the X11 and Windows specific code.
I'd like to wait for a comment from @ManoloFLTK because he is the main author of this stuff and I can't test with two screens on my test system (I could try to arrange something on an older Linux version though if needed...).

ManoloFLTK:
Wayland includes means to detect when a window changes screen. So there's no Wayland equivalent of the delayed rescaling used under Windows and X11. This MR therefore does not need to consider the Wayland case.

Albrecht-S:
@timothytylee Thanks again for your PR. I committed your changes together with the X11 changes as discussed above in commit 002b69f8fe90108faacaf2f728814fe52ff35095, hence I couldn't use your PR for a direct merge and I'm closing the PR now.

@ManoloFLTK FYI: I also removed obsolete/unused code and data from `Fl_Wayland_Window_Driver` in commit 374f22b8bf76c3c3e74640c098138b4a13748817 which is related to this PR.

PR #1251, opened 2025-04-20
CMake: Fallback to libOpenGL (no-X11 OpenGL)

lanodan:


ManoloFLTK:
@lanodan Could you, please, describe in what setup the proposed change proves useful?

lanodan:
For systems without X11 present but where OpenGL is available (libGL.so being OpenGL+GLX, and libOpenGL.so purely about OpenGL), for example Gentoo and Buildroot allows this.

ManoloFLTK:
Thanks for this explanation. I'll not be available until next week to consider this PR.

CMake/options.cmake
```diff
@@ -651,7 +651,8 @@ if(FLTK_BUILD_GL)
     unset(HAVE_GL_GLU_H CACHE)
     find_file(HAVE_GL_GLU_H GL/glu.h PATHS ${X11_INCLUDE_DIR})
   else()
-    find_package(OpenGL)
+    find_package(OpenGL COMPONENTS OpenGL)
+    set(OPENGL_LIBRARIES -lGLU -lOpenGL)
```
Albrecht-S, line 655: This seems wrong. Adding explicit linker options `-lGLU -lOpenGL` unconditionally w/o even checking the result of `find_package(OpenGL COMPONENTS OpenGL)` doesn't seem correct. Those linker options should be taken from the result of `find_package()` (targets) and not by directly specifying them (even if this is done elsewhere in our CMake files).

MatthiasWM:
Thanks for linking back to me @Albrecht-S .  From what I can see, this fixes an issue if a user has Wayland and OpenGL installed, but not the X11 OpenGL links, right? Maybe @ManoloFLTK knows more?

@lanodan TO test this, I'd have to set up a virtual machine. What Linux distro in what version would expose the faulty behavior? Which packages must be installed, and, probably more important, which packages must *not* be installed to expose this?

ManoloFLTK:
@lanodan Commit f2db90c should support systems without X11 present but where OpenGL is available. Do you confirm?

lanodan:
>@lanodan Commit f2db90c should support systems without X11 present but where OpenGL is available. Do you confirm?

Tested master (f2db90cb77466d6720c348be78e491a7d00c9e8c) since it got merged and yeah, I confirm it works, thanks.

PR #1252, opened 2025-04-20
CMake: Fix dbus automagic

lanodan:
Allows to disable dbus usage even on systems where dbus is installed.


ManoloFLTK:
@lanodan could you, please, describe a scenario where it would be useful to build FlTK without dbus even in presence of dbus? TIA

Albrecht-S:
@ManoloFLTK Can you please explain why the current implementation uses `dbus` at all? What is it used for, and what would not work w/o `dbus` support? What I mean is: why not make it optional, as @lanodan requested?

But I agree that @lanodan should also explain why this would be useful. I'd like to see the rationale for both options.

ManoloFLTK:
FLTK uses dbus to get the value of the cursor theme which determines the series of cursor icons associated to a given cursor shape. There's a fallback mechanism used if for some reason the cursor theme is not given by dbus that queries environment variables for the cursor theme.

Albrecht-S:
@ManoloFLTK Thanks for the explanation, that seems to be useful.

ManoloFLTK:
"It's only used for vendored libdecor"
That's not what I would write. It's used both by libdecor to set the cursor shape when it's above a window titlebar AND by FLTK when the cursor is above FLTK windows. The cursor theme name libdecor discovers isn't transmitted to FLTK which needs to access dbus to discover itself the same information.

Is dbus available on Gentoo?

ManoloFLTK:
I understand now the "automagic dependency" problem. I'm not available until next week to consider this PR. Thanks for  proposing a fix.

Albrecht-S:
> I understand now the "automagic dependency" problem. ...

Thanks to @lanodan for the link. I believe that the patch is technically correct and could be applied, but I'm leaving the final decision to Manolo.

Points to consider:
1. As I understand we'd lose functionality (cursor theme detection) when using the bundled libdecor - or even if we use the system libdecor (?) - _**if the option is turned off**_. This should be documented clearly in `README.CMake.txt`. Documentation of this option is missing in the patch anyway.
2. Therefore the option text should also include a warning, such as "not recommended to turn off" or something like that, or for instance:
```CMake
option(FLTK_USE_DBUS "use D-Bus - required to detect the cursor theme" ON)
```

@lanodan Thanks for the patch, I'm definitely not requesting you to change it because I believe that the details mentioned above are up to FLTK devs (in this case @ManoloFLTK).

@ManoloFLTK Please feel free to correct my statements above, or apply better changes. TIA.

ManoloFLTK:
@lanodan Many thanks for this PR. It's now committed to the master branch with a reference to this PR and your name. I expect the commit to remove the dbus automagic. If there are remaining issues, please, open an issue.

PR #1287, opened 2025-07-30
Fl_Gl_Window::make_current() does not crash if the window hasn't been shown yet

dkogan:
Hi. Currently if you call any Fl_Gl_Window functions before the window has been shown, a crash results: make_current() calls create_gl_context(g=NULL), and creaate_gl_context() then dereferences this NULL pointer.

There's no reason to actually touch Fl_Gl_Window that early usually, but I've done it many times by mistake, and crashing isn't what should happen in that case. This patch fixes this by making make_current() do nothing in this case. It's a weird way to do it (it should look at g and not at shown(), and it should probably report an error somehow instead of doing nothing), but that's how this code does it in the other place where create_gl_context() is called.

I tested it in my cases, and it makes things work

### DIFF
--- a/src/Fl_Gl_Window.cxx
+++ b/src/Fl_Gl_Window.cxx
@@ -118,6 +118,9 @@ int Fl_Gl_Window::mode(int m, const int *a) {
 void Fl_Gl_Window::make_current() {
 //  puts("Fl_Gl_Window::make_current()");
 //  printf("make_current: context_=%p\n", context_);
+
+  if (!shown()) return;
+
   pGlWindowDriver->make_current_before();
   if (!context_) {
     mode_ &= ~NON_LOCAL_CONTEXT;


Albrecht-S:
Thanks for the report and the PR. It's always good if FLTK doesn't crash by dereferencing a `nullptr`, hence I consider your patch as a "minimal solution" (more as a workaround than a solution).

However, after calling `make_current()` the code would likely call GL drawing. I wonder what happens if GL drawing calls are made after such a skipped `make_current()` call.

Perhaps it would be a better solution to issue a qualified error message and terminate the program rather than doing "nothing" because this is a user programming error that should be fixed. ISTR that there were similar cases where we did that.

You wrote:
> I tested it in my cases, and it makes things work

What things does is make work in your case?

dkogan:
Hi. Thanks for looking!

I agree that this patch is not as complete a fix as it could be. The reasons for this minimal proposal are what I said in my report:

1. The other call to =create_gl_context()= in the same file is protected in the same way
2. This makes my pre-bug-fix application work

I just typed up the details of my application, but they don't matter: the patch worked around that one bug, but would not work in general. Reporting an error would be better. Doing this correctly would need to

1. Make =make_current()= report an error, even though it currently returns =void=
2. Treat =flush()= the same way (it also calls =create_gl_context()=). It also returns =void=, and has lots of other logic

This might need to be handled by somebody more involved in fltk developemnt than me.

Thanks.

MatthiasWM:
Not crashing is always a good thing. FLTK does not have a concept of calling assert if there are user errors, so unless we start that now, this patch is fine with me.

Issue #208, opened 2021-04-03
Suggestion: Please add ability to use tablet features like pen pressure sensitivity, tilt, and other stuff.


ghost:
I want to use FLTK, but I truly need ability to use tablet features such as pen pressure sensitivity and other stuff related to the tablet. How difficult would it be to implement it into FLTK?

By tablet I mean my drawing tablet like a wacom tablet.

fire-eggs:
You don't tell us what platform(s) you wish to target. Drivers for tablets are platform and sometimes manufacturer specific. A fair amount of work is necessary to handle devices in a cross-platform fashion. [Someone more knowledgeable than I might know of a "standard" interface for tablets and similar devices.]

For instance, if you were targeting Wacom compatible tablets on Linux, you could install or use the [Wacom driver](https://github.com/linuxwacom/xf86-input-wacom) and write your code to use the tablet, following the test programs in that repo.

erco77:
Yes, we should add this, but no one's been able to come up with this cross platform as yet.

Back in 2016 I proposed a Mac specific patch in STR#3329 to support pen movement + pressure/tilt that worked for me on a Cintiq (Wacom monitor/tablet combo used for drawing), see:

[1] https://www.fltk.org/newsgroups.php?gfltk.general+v:40003
[2] https://www.fltk.org/str.php?L3329

But that's as far as I got.

I was not sure what the proper API should be, and sought input, but since none was offered, the STR has languished.
Could be revived though, we just need a way that (a) works for cross platform, (b) does not impact performance on non-tablet devices (since this causes a lot of excess event traffic, or at least in my case it certainly did).

fire-eggs:
I was going to point out that Gimp supports tablets [via GTK+], but then I read that their tablet support is broken on Windows and Mac.

erco77:
Inkscape might be a good reference for an app that uses e.g. pen pressure.
I use Gimp on a mac with the Wacom Cintiq and it works fine. If it's broken, I'm not aware of it.

But we already have an example of Mac <-> FLTK <-> Cintiq in STR#3329. What's needed is Win and Linux.
I'm pretty sure Linux probably uses the same stuff SGI did back in the day, as X input extensions, IIRC.

I wrote some pure X windows/OpenGL code very long ago (1998) on an SGI using X input extensions to read pen pressure and tilt from a Wacom Artz tablet, code for that is here: http://seriss.com/people/erco/wacom/

[...]

PR #1326, opened 2025-11-16
Pen/stylus/tablet API and driver implementation for macOS

MatthiasWM:
See #208.

The API is quite stable and 90% of the driver is implemented. For testing, if you have a tablet device, take a look at `test/penpal`.

Features:
 - driver is in namespace Fl::Pen
 - event system similar to FLTK core system
 - fallback to mouse events if not tablet found
 - supports tip, eraser, pressure, barrel pressure, tilt, twist, and more

There are still a handful of missing features. This pull request is mainly for demos and testing.

### DIFF
--- a/FL/core/pen_events.H
+++ b/FL/core/pen_events.H
@@ -32,12 +32,13 @@ class Fl_Widget;
 
 namespace Fl {
 
+/** FLTK Pen/Stylus/Tablet input driver API. */
 namespace Pen {
 
 /**
  \defgroup fl_pen_events Pen and tablet event handling
  \ingroup fl_events
- \brief The Fl::Pen namespace API, declared in <FL/core/pen_events.H>
+ \brief This chapter documents the Fl::Pen namespace API, declared in <FL/core/pen_events.H>
 
  The FL::Pen namespace contains everything needed to work with a pen type input
  device, either in connection with an external tablet, or as a stylus for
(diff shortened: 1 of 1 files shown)


erco77:
I tried penpal - works for me! I can draw fine and pressure changes the pen size (see comments below), and eraser draws in white with pressure changing its size.

In my case, a Wacom Cintiq DTK220 (22") which has pen and eraser pressure, two axis tilt, and two buttons on the pen that are I think usually used for popup menus or other programmable options. There are also 9 buttons on the left, and 9 buttons on the right side of the screen. I don't usually use them, but some seem to be preprogrammed to open things like radial menus and such. Not sure what those are for.

In penpal, my only comment is I found it hard to control pen pressure.. since it affects line thickness, when I try to touch carefully to get thin lines, it jumps between a large dot and small size really quickly. Other than that, works fine for harder pressures; more controllable.
Not sure what controls color change; just by using it, it seems maybe when there's a focus change (e.g. moving the pen over the edges of the boxes)?

Attaching a screenshot:
image

MatthiasWM:
@erco77 Thanks for trying out the PR! I changed the pressure to use the squared value to calculate the radius. The result is much more controllable. What the buttons on the tablet do is not part of the driver. Wacom has a setup app that maps them to shortcuts depending on the frontmost app. Similarly for the barrel buttons: users can assign popup menus, double clicks, etc. , but the FLTK driver takes only what come through from the Wacom Driver.

MatthiasWM:
Tested the code thoroughly on macOS, including event delivery. Test app "Penpal" works and make sense in the way it works. I am confident that this can be added to master.

HTML documentation is generated in "Topics / Events handling functions / Pen and tablet event handling".

erco77:
Matt: noticed some new build warnings in 1.5.x master showing up on mac builds:
___
```
[ 58%] Building CXX object src/CMakeFiles/fltk-shared.dir/drivers/Cocoa/Fl_Cocoa_Pen_Events.mm.o
/Users/runner/work/fltk/fltk/src/drivers/Cocoa/Fl_Cocoa_Pen_Events.mm:300:1: warning: non-void function does not return a value [-Wreturn-type]
  300 | }
      | ^
/Users/runner/work/fltk/fltk/src/drivers/Cocoa/Fl_Cocoa_Pen_Events.mm:487:7: warning: variable 'ret' set but not used [-Wunused-but-set-variable]
  487 |   int ret = 0;
      |       ^
2 warnings generated.
```
____

MatthiasWM:
> 2 warnings generated.

Thanks, fixed. No clue why my local machine didn't give me those.

PR #1344, opened 2025-12-01
fix table-sort dir output parsing with strtok()

phetdam:
## Summary

This PR updates the `table-sort` example program to handle the output of `dir` on Windows better.

It ensures that the "Filename" column correctly displays the file/directory names and the "Size" column will have the file sizes. The unnamed third column will hold the `"<DIR>"` marker for any directory listed.

## Details

Currently, after building + running `table-sort` on Windows, one will see the following window (if run in repo root):

table-sort-win32-current

The reason for the empty "Filename" column is that `MyTable::load_command()` reads tokens from a line of text using `strtok()`, which is fine when using `ls -l` on POSIX systems, but doesn't work as expected with `dir` output on Windows. `dir` output requires some special handling because the `dir` output on the FLTK repo root will look something like the following:

```
    ... CMD header (5 lines) ...
2025-11-30  23:42    <DIR>          .
2025-11-30  22:31    <DIR>          ..
2025-11-30  22:31             5,161 .clang-format
2025-11-30  23:45    <DIR>          .git
2025-11-30  22:31             1,291 .gitattributes
  ... other directory entries ...
              36 File(s)        597,038 bytes
              26 Dir(s)  489,786,810,368 bytes free
```

One can see that each directory entry, if parsed using `strtok()` with `" \t\n"` as delimiters (which is what is currently used), each directory entry line yields only 4 tokens from 5 text columns. Therefore, the 5th column, "Filename", ends up blank.

We fix this by ensuring that on Windows (i.e. enclosed in a `#ifdef _WIN32` block) after reading the first 2 tokens of the `dir` output, we check if the 3rd token is `"<DIR>"` or not, appropriately inserting an empty string after/before appropriately so that each table row gets 5 columns. Since the last row of the `dir` output has 5 tokens (results in 6 table columns), we also remember to resize the table back to 5 columns after dropping the last 2 rows of the `dir` output.

After these changes, `table-sort` now more appropriately mirrors the `dir` output:

table-sort-win32-updated

### DIFF
--- a/examples/table-sort.cxx
+++ b/examples/table-sort.cxx
@@ -277,7 +277,8 @@ void MyTable::fill_from_dircmd() {
     // will get an extra empty, headerless column in our table due to the extra
     // "padding" we do when parsing the 3rd strtok() token for a line
     if ( rowdata_.size() > 2 ) {
-        rowdata_.pop_back(); rowdata_.pop_back();
+        rowdata_.pop_back();
+        rowdata_.pop_back();
         cols(rowdata_.back().cols.size());
     }
 #endif


erco77:
Thanks for the PR, will give it a look.

Of interest, the issue with commas in numeric values can probably be solved by adding the `/-C` flag to DIR, e.g.
```
C:\TEMP\FLTK_1.5.x.GIT> DIR /-C
[..]
12/01/2025  07:58 AM              3957 README.abi-version.txt
12/01/2025  07:58 AM              9304 README.Cairo.txt
12/01/2025  07:58 AM             45420 README.CMake.txt
12/01/2025  07:58 AM              2497 README.CPack.txt
12/01/2025  07:58 AM              5660 README.documentation.txt
[..]
              34 File(s)         502096 bytes
              23 Dir(s)    281166856192 bytes free
```
This leaves out the 1000's separator from all the numeric file size values, which would fix numeric sorting.
According to [this page](https://www.robvanderwoude.com/allhelpw2ksp4_en.php#DIR) the `/-C` flag is supported as far back as Win2k, so should be safe for us to use.

erco77:
Looks like on your system DIR doesn't show an AM/PM field after the time, which adds another wrinkle to the parser, making this get messy depending on the locale.. ugh.

I'd thought about using a different command on Windows that's easier to parse, like `netstat /n`.

Given all the apparent DIR output variations, that might be a better solution to keep the example program's parser as simple as possible. Will try some experiments and follow up.

erco77:
Suggesting this alternative v1 patch against the current fltk master, to solve the issue:
[fix-1344_v1.patch](https://github.com/user-attachments/files/23861755/fix-1344_v1.patch)

..here's the results I get on my Win10 machine (left) and Linux machine (right) with the US/EN locale:

image

Perhaps you can confirm if this patch works on your Windows system as well.

erco77:
Applied my v1 patch (with small optimization) to fltk master as commit 2e16977.
If you can confirm that solves the parsing issues, I'll close as "solved".

phetdam:
> Looks like on your system DIR doesn't show an AM/PM field after the time

Yes, you're right. That does make things more tangled...

> I'd thought about using a different command on Windows that's easier to parse, like `netstat /n`.

Although I see you've already applied a patch by tweaking the `dir` command to exclude directories and omit the commas in the file sizes, I had actually been thinking about just using made-up table values that would work regardless of platform. In particular, one thing I had thought of was to make a table with the following columns:

1. word, such as "unforgivable"
2. word length (this is an easy numeric column)
3. [ROT13](https://en.wikipedia.org/wiki/ROT13) word (since ROT13 should be easy to do)
4. vowels in word (e.g. cumulative count of a, e, i, o, u characters)
5. `std::hash<std::string>` value (another easy numeric column)
6. XOR with `0x4` (`0b100`) word (shifts each `char` value up or down by 4)

I figured this might be a not-too-difficult, yet not 100% trivial, way to produce some made-up table values so we don't need to worry about `popen`, parsing using `strtok`, etc. which adds layers to the example.

> Applied my v1 patch (with small optimization) to fltk master as commit [2e16977](https://github.com/fltk/fltk/commit/2e169770396fed12b3b9200b3354322924b36537).

Looks good for me locally (Windows 11) so yes please close this PR.

Let me know if the above idea I have for making up table content is of interest; if so, I can make a separate PR later this week.

RokerHRO:
Maybe the program should _not_ parse the – plattform & locale dependent – output of _another_ program but just do its own directory listing using C++17's `std::filesystem` functions? 🤔

phetdam:
> Maybe the program should _not_ parse the – plattform & locale dependent – output of _another_ program

Yes, that's the reason for my proposal to make up our own table data and avoid any platform/locale issues.

> just do its own directory listing using C++17's `std::filesystem` functions? 🤔

Well, FLTK currently only requires C++11 support. Although one could check [`CMAKE_CXX_COMPILE_FEATURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_COMPILE_FEATURES.html) for `cxx_std_17` to conditionally build the program when a client has a C++17 compiler, I'm not sure it's worth the effort.

RokerHRO:
1) a test/example program might have other/tighter platform requirements than the base library.

2) Just ask yourself: What's the purpose of the test/example program?
I think, you're right: a set of algorithmically generated test data would fulfill the purpose "show how to use `Fl_Table`". So: Go for it! :-)

erco77:
> Looks good for me locally (Windows 11) so yes please close this PR.

OK great; closing.

Regarding "fixed data", this example was designed to show table sorting of actual dynamic, cross-platform real world data from an external program - file browsers being something most people were familiar with.

It's probably good that in this case the real world data has some gotchyas, and showing how we can navigate around them.

The demo was also designed to show real world use of qsort() to dynamically detect data types of the sort data on the fly (numeric vs. alphabetic), as well as the usefulness of the popen() function on both Unix and Windows; popen() is not well known to most Windows users, who normally have to struggle with the WIN32-specific CreateProcess()/CreatePipe()/JobObject()/etc APIs which are /horribly/ complex.

Anyway, leaving the app "as is" to demonstrate all these things together, as sometimes people ask for reference to how to do many of these things, and table-sort makes a good reference for this.

If we feel we need a separate example that's simpler, e.g. "table-sort-simple", that could be a separate RFE.

phetdam:
> this example was designed to show table sorting of actual dynamic, cross-platform real world data from an external program

Platform portability was why I proposed using made-up data as then there wouldn't be any need to conditionally define a macro for `popen`, select between `ls` and `dir`, and need to deal with different output formats.

> It's probably good that in this case the real world data has some gotchyas, and showing how we can navigate around them.

True, but `ls` and `dir` have fundamentally different table schema (headers). Since we know the type in advance of each header, the columns could be statically typed. Of course, we could also choose to type the columns using some sort of variant.

But either way, if I just wanted to know how to create a sortable table, I'd rather just have simple data since my focus would be on using FLTK correctly. Designing the table schema/storage scheme would be more application-specific.
 
> struggle with the WIN32-specific CreateProcess()/CreatePipe()/JobObject()/etc APIs

Very true...
 
> Anyway, leaving the app "as is" to demonstrate all these things together, as sometimes people ask for reference to how to do many of these things, and table-sort makes a good reference for this.

I think that's a fair justification for keeping some of the sharp edges in this example.

> If we feel we need a separate example that's simpler, e.g. "table-sort-simple", that could be a separate RFE.

Sounds good. When I have some time I may try this; I have a working local modification of `table-sort` that uses the generated data I mentioned in [my previous comment](https://github.com/fltk/fltk/pull/1344#issuecomment-3600541494). The current state (when unsorted) looks like this:

table-sort-win32-new

Anyways, thanks for taking a look at this PR. I'm happy to see that FLTK is still actively maintained (and improving with the times--CMake integration is definitely appreciated given how commonly it is used in the C/C++ ecosystem).

erco77:
OK, so for a followup RFE, keep in mind there's already 8 table related examples, so there's some pressure to keep the number of files down to prevent overwhelming new users with _too much_ info.

That said, tables have so many different use cases, multiple examples are kinda required.

So if the table-sort example is really too hard to see the widget code due to the command output parsing, I'm open to a more stripped down "table-sort-simple". Either that, or making the existing code "clearer" so it's easier to see the separation between the widget code and the parser/command.

These final notes would be a good reference for a followup RFE.

phetdam:
> there's already 8 table related examples, so there's some pressure to keep the number of files down to prevent overwhelming new users with _too much_ info
>
> That said, tables have so many different use cases, multiple examples are kinda required.

Very reasonable. And so each example should ideally be targeted for a distinct use case and not overlap too much.

> So if the table-sort example is really too hard to see the widget code due to the command output parsing, I'm open to a more stripped down "table-sort-simple". Either that, or making the existing code "clearer" so it's easier to see the separation between the widget code and the parser/command.

Sounds good and thanks for the feedback :) I will think about how I might turn my local changes into a `table-sort-simple` example. There's some things that could be C++11-ed but the overall idea would be to simplify where possible.

Issue #1365, opened 2026-02-22
Blurry fonts with cairo/pango on FLTK >= 1.4


rodarima:
**Description of the Bug Report**

While testing FLTK 1.4.4 on Dillo I saw that fonts seem to be more blurry. I reproduced the issue with the fonts example program from FLTK 1.3.3 and the latest commit from master 1be412ab69fd689639764b3f649795e9b99594aa while selecting the "Dejavu Sans" font (notice there is no book in FLTK 1.3, but it should use the same):

<img width="891" height="680" alt="Image" src="https://github.com/user-attachments/assets/74c6b2ed-873f-4d4b-a5b7-c1cc1fb827a9" />

I'm currently using Xorg with these Xft settings, but I suspect this problem has nothing to do with the DPI:

```
Xft.autohint: 1
Xft.dpi: 101
```

I suspect that Cairo/Pango is not doing the proper hinting and that causes the font to become more blurry.

The wikipedia has a nice diagram: https://en.wikipedia.org/wiki/Font_hinting

I can reproduce a similar problem by changing the hinting with the pango-view program (`--hinting=none` vs `--hinting=full`). I'm not sure if there is a configuration file or environment variable that users can change to modify the rendering. This is how that tool does it:

https://github.com/bratsche/pango/blob/d0830820f84eb28a27502306b6acbf0f2b57fc7d/pango-view/viewer-pangocairo.c#L54-L60

It has more options:

https://www.cairographics.org/manual/cairo-cairo-font-options-t.html#cairo-hint-style-t

I probably would also need to adjust the other options. Is there a possibility to modify these options from FLTK? I would imagine a configuration section for cairo in the FLTK configuration file `~/.fltk/fltk.org/fltk.prefs` may be the best place so that all programs look as the user wants? I haven't seen any way to specify this in a cairo configuration file.

Notice that I also want to apply these options to the FLTK menus, not only to the Dillo canvas.

**FLTK Configure / Build Options**

 - `cmake -S . -G build-test`

Here is the CMake summary:

```
-- Configuration Summary for FLTK 1.5.0 generated by CMake 4.2.1 --

-- The following OPTIONAL packages have been found:

 * Doxygen
 * LATEX
 * ZLIB
 * PNG
 * JPEG

-- Static libraries    will be built in: /home/ram/dev/dillo/fltk/git/build-test/lib
-- Shared libraries    will not be built (set FLTK_BUILD_SHARED_LIBS=ON to build)
[...]

PR #1368, opened 2026-03-01
Correct draw translation by 0.5 pixels only

rodarima:
Using 1.5 pixels causes text to be rendered one pixel offset to the top left. This is visible when text is selected as the characters touch the top left border.

The fl_text_extents is also updated to make sure the bounding box is properly placed.

See also #1365

---

Here is the utf8fonts and unittests examples from before:

old-selection
old-extents

And from after:

new-selection
new-extents


ManoloFLTK:
@rodarima I do see there's a 1-pixel difference between the 2 images, but I don't see which one is preferable to the other, and why is it so.

rodarima:
The problem is that the text is now drawn with a (-1,-1) offset and it touches the right side of the selection area, which didn't happen before (in FLTK 1.3). This is visible in the location bar of Dillo where the "h" (of http) touches the selection. The "File" menu label is also not centered.

Here are the three versions, FLTK 1.3, before this patch and after:

dillo

I added a small red arrow in the selected text, let me show it side to side with zoom. The effect is also visible in the number of fonts text.

diff

PR #1385, opened 2026-03-25
[Idea] Fl_Scrollbar: Remove arrow buttons for very thin scrollbars

dannye:
**Is your feature request related to a problem? Please describe.**

Currently the "arrow buttons" on both ends of a Fl_Scrollbar disappear if the scrollbar is not at least 3 times as long as it is wide. This is so that the "track" between the buttons is guaranteed to be at least 1/3 of the total length of the scrollbar; if the scrollbar is not long enough, the buttons disappear so that the "track" gets 100% of the length of the scrollbar. [<sup>1</sup>]

However, I think the buttons should also disappear if the scrollbar is too thin.

If the scrollbar is too thin, the buttons are not rendered well and are not very user-friendly.

For example, at 6 pixels wide:

image

Personally, I think the arrow buttons are only suitable when the scrollbar is at least 7 pixels wide or wider.

When the scrollbar is less than 7 pixels wide (for example, 6 pixels wide), I think it works much better without the arrow buttons:

image

**I am using FLTK**

 - Version 1.4.x

**Describe the solution you'd like**

I would really like to add a "thin" scrollbar like this to my FLTK 1.4.x app, but I think currently it is not possible. The proposed solution in this PR would not break the ABI and could be considered for 1.4.5 if there were no objections (although I understand that a "last minute" suggestion/addition like this is not ideal).

My chosen "minimum width" of "7" could be made to be customizable to the application programmer, which might be even better/more flexible, but I think that would break the ABI.

**Describe alternatives you've considered**

I don't think I can change this in my app without making this change directly to FLTK, unless I copy/paste the whole Fl_Scroll widget into a custom widget that is identical to Fl_Scroll except that it uses a custom subclass of Fl_Scrollbar instead of the "real" Fl_Scrollbar class. This seems like a lot of effort for such a small change (a change which I think would also be an appropriate improvement to FLTK itself).

Thanks for reading.

---

[<sup>1</sup>]: My point in explaining this is to point out that it is not unprecedented for Fl_Scrollbar to intelligently hide its arrow buttons at its own discretion.

### DIFF
--- a/src/Fl_Scrollbar.cxx
+++ b/src/Fl_Scrollbar.cxx
@@ -68,9 +68,9 @@ int Fl_Scrollbar::handle(int event) {
 
   // adjust slider area to be inside the arrow buttons:
   if (horizontal()) {
-    if (W >= 3*H) {X += H; W -= 2*H;}
+    if (W >= 3*H && H >= 7) {X += H; W -= 2*H;}
   } else {
-    if (H >= 3*W) {Y += W; H -= 2*W;}
+    if (H >= 3*W && W >= 7) {Y += W; H -= 2*W;}
   }
 
   // which widget part is highlighted?
@@ -208,7 +208,7 @@ void Fl_Scrollbar::draw() {
     inset = 1;
 
   if (horizontal()) {
-    if (W < 3*H) {
+    if (W < 3*H || H < 7) {
       Fl_Slider::draw(X, Y, W, H);
       return;
     }
@@ -228,7 +228,7 @@ void Fl_Scrollbar::draw() {
       fl_draw_arrow(ab, FL_ARROW_SINGLE, FL_ORIENT_RIGHT, arrowcolor); // right arrow
     }
   } else { // vertical
-    if (H < 3*W) {
+    if (H < 3*W || W < 7) {
       Fl_Slider::draw(X, Y, W, H);
       return;
     }


Albrecht-S:
@dannye Thank you for this PR, this is an interesting feature, and the patch looks simple and very likely correct (not tested). However, according to our policy, this would be a new feature and would only be implemented in 1.5.0 (if accepted).

**_Regarding ABI changes:_** Theoretically, we could break the ABI in version 1.4.x by using so-called "ABI guards", which you are probably familiar with. Users who want such new ABI features or bug fixes could build e.g. for ABI 1.4.5 (CMake option: `FLTK_ABI_VERSION=10405`).

In 1.5.0 we can break the ABI, and thus it would be an option to make this feature user settable. Options that come to mind:
1. a threshold for scrollbar width that suppress the arrow widgets, default would be 7 as you suggest, or 0 (no suppression)
2. a general option to suppress the arrows per `Fl_Scroll` widget
3. an option to suppress the arrows per `Fl_Scrollbar` widget

The latter (3.) could be applied by user code to each scrollbar of an `Fl_Scroll` or any other widget with scrollbars. This would probably be the most flexible solution. Option 2. could also be implemented and would set the "no-arrows" option on both scrollbars.

### Slightly OT but related

Last but not least, here's an _almost concrete_ vision (experimental code exists): I worked on a new `Fl_Scroll` widget (name yet unknown, maybe `Fl_Scroll2` or `Fl_Scroll_New`). This widget would allow the user to replace the scrollbars with their own scrollbars, would **_NOT_** include the scrollbars in the list of children, and thus don't count the scrollbars in `Fl_Scroll::children()` (which in my opinion is an annoyance in the current code). It's still experimental though, but this could (would!) also implement the "no-arrows" option in the future, no matter whether we implement it in the clasic `Fl_Scroll` (in 1.5).

Albrecht-S:
Addition/correction: when I wrote the above comment I was **thinking** of Fl_Scroll explicitly, but now I noticed that your patch would change `Fl_Scrollbar` in general. That's the correct choice, and some of my comments above regarding `Fl_Scroll` don't apply to your PR. Sorry for the confusion.

But a new `Fl_Scroll` widget with above mentioned features is in the pipeline anyway. ;-)

dannye:
> However, according to our policy, this would be a new feature and would only be implemented in 1.5.0 (if accepted).

No problem, that's totally understandable. Especially because as you pointed out there are many options/considerations to make to decide the best way to allow for this flexibility, so there's no reason to rush/accept this PR.

> Last but not least, here's an _almost concrete_ vision (experimental code exists): I worked on a new `Fl_Scroll` widget (name yet unknown, maybe `Fl_Scroll2` or `Fl_Scroll_New`). This widget would allow the user to replace the scrollbars with their own scrollbars, would **_NOT_** include the scrollbars in the list of children, and thus don't count the scrollbars in `Fl_Scroll::children()` (which in my opinion is an annoyance in the current code). It's still experimental though, but this could (would!) also implement the "no-arrows" option in the future, no matter whether we implement it in the clasic `Fl_Scroll` (in 1.5).

I think a new "Fl_Scroll2" widget would be fantastic!

In addition to the "no-arrows" option, there are a few more real inconveniences with the existing Fl_Scroll widget that I think provide great opportunities for improvement with the new "Fl_Scroll2" replacement, if they're not already on your radar:

1. It's difficult to keep the Fl_Scroll clamped to the max (right or bottom) when the scroll is resized.
   * For example, if the Fl_Scroll is already scrolled all the way to the right, and then the scroll/window is made wider, the "xposition" (horizontal scroll value) of the Fl_Scroll is not updated (reduced), so it ends up scrolled too far to the right "beyond the maximum".
2. It's difficult to have a Fl_Scroll with a scrollbar that is invisible but still functional.
   * As in, have a Fl_Scroll that can still be scrolled with the mouse wheel, but the scrollbar is not drawn.
   * This is especially difficult because so much state management for the scrollbars is done in the draw function.
3. It's difficult to have a Fl_Scroll where its scrollbars ignore arrow key events (and home/end/page up/page down).
   * Specifically, it's difficult to block the arrow key events from being received by the scrollbars without also blocking more shortcut events from being received by other (non scrollbar) children of the Fl_Scroll.

I've solved for all of these issues in my own application with subclasses of Fl_Scroll, but some of the code is not that great (especially for issues # 2 and 3; the fix for issue # 1 is not that bad but I think it's something that Fl_Scroll should offer automatically).
And my solution for issue # 2 only allows for the Fl_Scroll to have 1 scrollbar. I don't think it's possible to have a Fl_Scroll with two functional-but-invisible scrollbars at the same time.

Just sharing some thoughts! I would gladly try out the experimental "Fl_Scroll2" and provide feedback if you ever want.

Albrecht-S:
@dannye wrote:
> I think a new "Fl_Scroll2" widget would be fantastic!

Great! :-)

> ... many interesting thoughts elided ...
> I would gladly try out the experimental "Fl_Scroll2" and provide feedback if you ever want.

Thank you very much for your constructive comments and your thoughts. This gave me a new boost to think even more about a new Fl_Scroll2 widget, and maybe more widgets that suffer from non-optimal (TM) scrollbar integration.

I appreciate this, but we should stop the discussion here (because it's OT) and continue in a new "GitHub Discussion" thread. I'll open one when I'm ready to propose and discuss the code of the new widget (it's not yet ready, but I'll check what I can do). I'll refer to this PR when I open the new discussion [1] so your thoughts will not be lost!

-----
[1] Feel free to open a "GitHub Discussion" thread in the "[Development category](https://github.com/fltk/fltk/discussions/categories/development)" with your thoughts. If you do, I'll add my proposal to the discussion later.

MatthiasWM:
I added your suggestion to 1.5 master. I like it, and six pixel wide boxes with 4 pixel wide arrows do make very little sense.

I also moved your Fl_Scrollbar issues into a separate Issue #1584

dannye:
Thank you, Matthias!

PR #1390, opened 2026-04-01
Add missing libdecor dispatching

jadahl:
Libdecor has its own libdecor_dispatch() function that must be called
for things to work properly, and this function was never called, meaning
anything other than the Wayland socket of the libdecor context would
never be dispatched.


ManoloFLTK:
@jadahl Many thanks for these improvements. The essence of this PR has been committed to master in a slightly different form (8e9e2b7).

Issue #1391, opened 2026-04-02
Allow setting wayland app_id


electricbrass:
**Is your feature request related to a problem? Please describe.**
Currently the app_id for a wayland window is always set based on the filename of the executable. 

**I am using FLTK**
 - Version 1.4.1

**Describe the solution you'd like**
A function to set the app_id similar to the existing solution for X11 (`Fl_Window::xclass`).

**Describe alternatives you've considered**
Make `Fl_Window::xclass` set the wayland app_id as well.

**Additional context**
Currently, this means things like .desktop files can break just by the executable being renamed. This is also a problem for Flatpaks, where apps must follow particular naming conventions for .desktop files, so without the ability to rename the desktop file, the ability to set the app_id is needed instead.

ManoloFLTK:
@electricbrass I propose to use the `Fl_Window::iconlabel(const char*)` member function that is currently unused by the Wayland platform. I attach the corresponding patch to master. The patch shows with test program `hello.cxx` how the API could be used. Would that solve your demand? If yes, I could also commit the same change to branch-1.4.

[1391.patch](https://github.com/user-attachments/files/26439047/1391.patch)

electricbrass:
That would work for my needs, yes. I'm a bit confused why it would be done with iconlabel though. Since that's not really what this is, it's an identifier for the window/application, similar to X11's WM_CLASS.

ManoloFLTK:
@electricbrass That's right. This is better in `Fl_Window::xclass()`. That's now committed to master. Please, close issue if you confirm it's fixed. I will then backport the commit to branch-1.4.

PR #1392, opened 2026-04-02
Add ability to override default wayland app id

electricbrass:
Addresses #1391


Albrecht-S:
Thanks for the proposed patch, but AFAICT after a quick scan, it would break the ABI, and thus we can't support this in `branch-1.4`. Besides that, we don't allow new features in FLTK 1.4 because it's in maintenance mode.

@electricbrass It seems that Manolo applied a fix for #1391 in FLTK 1.5 (master) which makes this PR obsolete. Please close this PR if you agree. If you don't agree, please tell us why you still want this PR, or ...

Issue #1175, opened 2024-12-19
Visual glitches with Fl_Scroll inside Fl_Tabs


abau:
**Describe the bug**

Putting a `Fl_Scroll` inside a `Fl_Tabs` leads to visual glitches when scrolling down and back up.

**To Reproduce**

```cpp
#include <FL/Fl.H>
#include <FL/Fl_Double_Window.H>
#include <FL/Fl_Scroll.H>
#include <FL/Fl_Tabs.H>
#include <FL/Fl_Group.H>
#include <FL/platform.H>
    
int main() {
    int x = 0, y = 0, w = 400, h = 300;
    auto *win = new Fl_Double_Window(w, h);
    auto *tabs = new Fl_Tabs(x, y, w, h);

    fl_open_display();
    tabs->client_area(x, y, w, h);

    auto *scroll = new Fl_Scroll(x, y, w, h, "Group 1");
    auto *content = new Fl_Group(0, 0, 600, 600);
    
    content->end();
    scroll->end();
    tabs->end();
    win->end();    
    win->show();
    return Fl::run();
}
```

1. Scroll down
2. Scroll back up

**Expected behavior**

No visual glitches (e.g., white horizontal lines) show up.

**Screenshots**

[Screencast_20241219_063058.webm](https://github.com/user-attachments/assets/c270a755-fd5a-4e1f-88ad-c48d919c0f96)

**FLTK Version**

https://github.com/fltk/fltk/commit/809dae489ddf4755d1c546253f48ca2cf6ec800b

**FLTK Configure / Build Options**

 - `cmake .. -D FLTK_BUILD_EXAMPLES=ON`

**Operating System / Platform:**

- Arch Linux using KDE Plasma 6.2.4 with Wayland.

Albrecht-S:
@abau Thanks for the report, but I can't reproduce this on my Debian 12/Bookworm running Gnome/Wayland. Are you using display scaling `!= 100%` in your setup? I tried with odd scaling factors and couldn't reproduce these artifacts.

@ManoloFLTK and others: can anybody test under KDE Plasma and report?

abau:
> Are you using display scaling != 100% in your setup?

No, not that I'm aware of.

polter-rnd:
I tried the example in OP on GNOME / Wayland without scaling, I wasn't able to exactly reproduce it, but still something strange is happening with that horizontal line:

[...]

PR #1393, opened 2026-04-04
docs: Explain restrictions on Fl_Scroll usage with Fl_Tabs

pydsigner:
Related to #1175, this ensures the Fl_Scroll/Fl_Tabs rendering issue is documented on both widgets rather than just on Fl_Tabs.

### DIFF
--- a/FL/Fl_Scroll.H
+++ b/FL/Fl_Scroll.H
@@ -94,6 +94,11 @@
   <I>You cannot use Fl_Window as a child of this since the
   clipping is not conveyed to it when drawn, and it will draw over the
   scrollbars and neighboring objects.</I>
+
+  <I>When the Fl_Scroll is a child of an Fl_Tabs widget, its contents must be
+  kept at least five pixels (as of FLTK 1.4.x) away from the edges of the 
+  Fl_Tabs to avoid a tearing effect when scrolling.</I>
+  <!-- See GitHub issue #1175. -->
 */
 class FL_EXPORT Fl_Scroll : public Fl_Group {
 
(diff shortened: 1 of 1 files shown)


pydsigner:
Not sure if I should be making this change against the 1.4 branch instead.

Albrecht-S:
Thanks for this PR, but I'm not sure if we should commit it (as-is), for at least two reasons:
1. The given restrictions apply to **_every child_** of `Fl_Tabs`, it can only be more obvious (and more annoying) in `Fl_Scroll` widgets. If we added these specific docs to `Fl_Scroll`, then we'd need to add it to every widget which is not useful.
2. The wording "_its contents must be kept at least five pixels ... away from the edges_" is not entirely correct (complete). Note that even if **all children** of `Fl_Tabs` are initially inset by 5 pixels, then scrolling once could move them up (or down) to the border. The correct solution is (hopefully?) described in the docs of `Fl_Tabs` (use an invisible box, or inset the Fl_Scroll widget alone and keep other children at the border).

I'd rather add a very short comment with a link (\\see ...) to (the docs of) `Fl_Tabs` if we wanted to make such an **_exception_** for `Fl_Scroll` widgets. I'd definitely not want different (and incomplete) wording in more than one place. Duplication of docs (or code) would likely diverge later.

> Not sure if I should be making this change against the 1.4 branch instead.

That's not necessary nor useful. FLTK 1.4 is in maintenance mode (only bug fixes). We may however also add some minor documentation fixes/updates, but it's not necessary to make a PR against `branch-1.4`.

pydsigner:
@Albrecht-S I've reduced the comment, let me know if this is more what you're thinking. Thanks!

Albrecht-S:
@pydsigner Thanks for the PR and for the later update.

I decided to squash and merge your PR as-is, sorry for the delay.

PR #1398, opened 2026-04-05
Bugfix: Fix Multi Browser callback with FL_WHEN_NOT_CHANGED when Enter/space/up/down is pressed

dannye:
**Bug:** When using a Fl_Multi_Browser with a `when_` of FL_WHEN_NOT_CHANGED, the callback was not being invoked correctly/consistently when Enter/space/up/down is pressed. (The callback would be invoked in several different scenarios when it shouldn't be, and sometimes called with the wrong `FL_REASON_*`.)

To summarize the changes:

* For space, up, and down: Pass `when() & ~(FL_WHEN_ENTER_KEY | FL_WHEN_NOT_CHANGED)` when `select()` is called.
   * The `when()` needs to include something other than FL_WHEN_ENTER_KEY or FL_WHEN_NOT_CHANGED for `select()` to consider doing the callback when a change occurs. (Remember for up/down, shift or ctrl must be held.)
   * This stops the callback from being invoked in cases where it shouldn't be.
* For Enter:
   * Also pass `when() & ~(FL_WHEN_ENTER_KEY | FL_WHEN_NOT_CHANGED)` when `select_only()` is called. This stops the callback from being invoked in cases where it shouldn't be.
   * Previously, the callback would be invoked with `FL_REASON_CHANGED` after `select_only()` is called, regardless of whether a change occurred (ie, even if `select_only()` returned 0). This must only be done if `select_only()` returned 1, so this is now wrapped in an `if` instead of coming immediately after the call to `select_only()`.
   * Additionally, there is now an `else` block for invoking the callback with `FL_REASON_RESELECTED` if `select_only()` returns 0, and only doing so if `when()` includes `FL_WHEN_ENTER_KEY_ALWAYS` (ie, FL_WHEN_ENTER_KEY and FL_WHEN_NOT_CHANGED).

A video of the before and after isn't very helpful here, since you wouldn't be able to see when I am pressing space or enter, or see when I am holding shift or ctrl.

To test:
* Run test/browser.exe
* Set the "when" dropdown to "FL_WHEN_NOT_CHANGED", "FL_WHEN_ENTER_KEY", or "FL_WHEN_ENTER_KEY_ALWAYS"
* Test all combinations of Enter, space, up, down
* See that the callback is invoked only exactly when expected

### DIFF
--- a/src/Fl_Browser_.cxx
+++ b/src/Fl_Browser_.cxx
@@ -750,20 +750,26 @@ int Fl_Browser_::handle(int event) {
           }
           return 1;
         }
-      } else  {
+      } else {
         switch (Fl::event_key()) {
         case FL_Enter:
         case FL_KP_Enter:
-          select_only(l, when() & ~FL_WHEN_ENTER_KEY);
-          if (wp.deleted()) return 1;
-          if (when() & FL_WHEN_ENTER_KEY) {
-            set_changed();
-            do_callback(FL_REASON_CHANGED);
+          if (select_only(l, when() & ~FL_WHEN_ENTER_KEY_ALWAYS)) {
+            if (wp.deleted()) return 1;
+            if (when() & FL_WHEN_ENTER_KEY) {
+              set_changed();
+              do_callback(FL_REASON_CHANGED);
+            }
+          } else {
+            if (wp.deleted()) return 1;
+            if ((when() & FL_WHEN_ENTER_KEY) && (when() & FL_WHEN_NOT_CHANGED)) {
+              do_callback(FL_REASON_RESELECTED);
+            }
           }
           return 1;
         case ' ':
           selection_ = l;
-          select(l, !item_selected(l), when() & ~FL_WHEN_ENTER_KEY);
+          select(l, !item_selected(l), when());
           return 1;
         case FL_Down:
           while ((l = item_next(l))) {


MatthiasWM:
Phew. I am not entirely sure if I agree with all changes. For Kp_Enter, that's ok. For Ctrl-Up//Down, not sure if I want to limit these callbacks. But to be perfectly honest, I am not too happy about the keyboard and mouse behavior of these widgets in general.

dannye:
@MatthiasWM Yeah, I know what you mean.

FWIW, this PR is basically the same as the already-merged #1396 except this PR is for "Multi" instead of "Hold".

The main issues that make this hard to reason about are:

1. I don't know \*why\* anyone would want to set a browser's "when" to FL_WHEN_NOT_CHANGED. (What practical reason could there be for \*only\* wanting to be notified when the selection does \*not\* change?)
   * Since it's hard to imagine why someone would do this, it's hard to imagine exactly what the behavior ought to be.
   * But, since test/browser.cxx offers FL_WHEN_NOT_CHANGED to test, then I think it should behave according to common sense (ie, don't invoke the callback when the selection \*does\* change, even if done with the up/down arrow keys).
2. Multi is a lot more complex than Hold, especially WRT to the possibilities of the "when".
   * For example, what if lines 1, 2, and 3 are already selected, then the mouse is pressed on line 1 then dragged to line 3 and then released? Did the selection change? From the user's point of view, no, not really; lines 1-3 were selected before and after the whole mouse interaction. But from FLTK's point of view, yes, because the selection changed multiple times as the mouse was dragged, even if the selection ended up being the same as when the mouse was pushed. This probably isn't worth "fixing" because it would be even more complicated and who would benefit from this extra nuance anyway?
   * That's why here I only attempt to fix keyboard interactions (which are much more instantaneous and therefore less complicated) and I do not touch mouse behavior.

I'm just aiming to make callback behavior both predictable and consistent, and so I still think this PR is a good idea. But I acknowledge that this is a possibly unimportant edge-case, so I'm okay with you rejecting some or all of this PR.

erco77:
Regarding (1), FL_WHEN_NOT_CHANGED.. just guessing: I imagine double clicking on something where its state stays the same but still wanting to get a callback to process the event for some special behavior, or clicking inside the widget but not on any item for some special behavior handled by the callback, like clearing all selections or maybe even posting a right-click menu.
Anyway, not sure offhand, but I'm guessing it's one of those things people like to do with callbacks. Arguably that kind of behavior would done with a subclass that overrides `handle()`.

MatthiasWM:
_tl;dr: history, thoughts, suggesting using the quick fix for now, hinting at the need for specs before (re)implementation._

`FL_WHEN_NOT_CHANGED` historically was an addition to `FL_WHEN_CHANGED`. Imagine a group of radio buttons. Original FLTK would send a callback if the user pressed a not selected radio button and change the selection. But then a developer wanted a callback, even when the user clicked on the already selected radio button, so `FL_WHEN_NOT_CHANGED` was added. There is little to no sense in setting `FL_WHEN_NOT_CHANGED` without setting `FL_WHEN_CHANGED`. There is actually very little sense in *not* setting `FL_WHEN_CHANGED` because setting a callback implies that all changes should trigger said callback.

There is a lot of ad-hoc code in the FLTK event and callback code because things were never documented, and if they are documented, this was done after the fact. A lot of that is my fault (callback reasons). 

I talked with @Albrecht-S at length that a lot of our fixes are the result of never defining things. `Fl::event_key()` is a perfect example. We never decided what exactly this returns, and with the original code written in 1990's X11, all other implementation just mimicked what the implementor thought it returned, mixed with what the host API has to offer.

I have started a table with all sane combination of FL_WHEN bits on the left, and all useful user action at the top (click, ctrl-click, kp_enter, shift-up, etc. etc.) and tried to fill every field with if a callback is called, what reason is given, and what the widget `changed()` flag should return. But just filling in what "Shift Click and Drag on a previously selected item" should do on a Multi Browser is not easy. And it's just one piece in a 3D puzzle.

Albrecht-S:
I'd like to clarify in some kind what @MatthiasWM wrote:
> FL_WHEN_NOT_CHANGED historically was an addition to FL_WHEN_CHANGED.

Generally `FL_WHEN_NOT_CHANGED` should be thought of as an **additional trigger** or a **modifier** of other when() conditions of a callback, i.e. it's one more bit in the bits of when() conditions which should always be or'ed with other `when()` conditions.

As an example, the [documentation in FL/Widget.H](https://github.com/fltk/fltk/blob/faf210a1d0db05d1469ee993b9dfe797c90cc65f/FL/Fl_Widget.H#L851) says: "FL_WHEN_ENTER_KEY|FL_WHEN_NOT_CHANGED: The Enter key will do the callback **_even if_ the text has not changed**. Useful for command fields." Some lines below: "Fl_Widget::when() is a **set of bitflags** used by subclasses of Fl_Widget to decide when to do the callback."

But even if this has been specified in the docs of Fl_Widget, every subclass needs to specify by what events the callbacks are triggered, and how the when() conditions (bits) are interpreted by this specific subclass.

As @MatthiasWM wrote, this has not always been done, and "common sense" does not always help to decide. :-(
What is "common"?

That said, @dannye wrote:
> I don't know *why* anyone would want to set a browser's "when" to FL_WHEN_NOT_CHANGED. ...

That's certainly not meant to be done by anybody, at least not as the only `when()` condition (bit).

OK, so far my attempt to clarify things, I hope this helps. Unfortunately, I cannot (currently) read/test/understand the proposed patch, so I'll leave that to @MatthiasWM, and I can't comment on the widget author's intentions and the implementation details.

dannye:
Thanks. Everything that everyone is saying makes perfect sense.

@Albrecht-S wrote:

> That said, @dannye wrote:
> 
> > I don't know _why_ anyone would want to set a browser's "when" to FL_WHEN_NOT_CHANGED. ...
> 
> That's certainly not meant to be done by anybody, at least not as the only `when()` condition (bit).

Well, I only considered it so strongly because the test/browser.cxx demo program offers it (ie, allows the user to set FL_WHEN_NOT_CHANGED as the only `when()` condition bit).

So I very much assumed that "only when not changed" is entirely expected/planned for.

...but Enumerations.H very clearly says:

```cpp
  FL_WHEN_NOT_CHANGED       =  2, ///< Do the callback whenever the user interacts with the widget
```

which pretty much totally negates the idea that the definition of "only FL_WHEN_NOT_CHANGED" is "_only_ when not changed", but rather it probably means "always, when changed or not", which would implicitly behave like `FL_WHEN_CHANGED | FL_WHEN_NOT_CHANGED` even if the FL_WHEN_CHANGED bit isn't actually set.

I can reduce this PR to only fix the "reason" problem with the Enter key (FL_REASON_CHANGED versus FL_REASON_RESELECTED) and remove the special handling for FL_WHEN_NOT_CHANGED.

I (or anyone else) can also revert PR #1396 which serves no good purpose anymore.

dannye:
Actually, even if we accept that FL_WHEN_NOT_CHANGED should never be the only flag and it should always be paired with _something_, the important question is: What if it is only paired with FL_WHEN_ENTER_KEY?

Well, that is equivalent to FL_WHEN_ENTER_KEY_ALWAYS (10), which is defined as:

```cpp
  FL_WHEN_ENTER_KEY_ALWAYS  = 10, ///< Do the callback when the user presses the ENTER key, even if the value doesn't change
```

So the mere presence of FL_WHEN_NOT_CHANGED does _not_ imply that the callback should pretty much always be invoked whether changed or not. It has to be paired with something _other than_ FL_WHEN_ENTER_KEY.

Now I think the important question for Multi Browsers specifically is: Should Up, Down, and Spacebar be treated _like_ Enter for the sake of convenience with regard to FL_WHEN_ENTER_KEY?

I kind of think yes, because it just seems more convenient and useful that way for a Multi Browser.

That's basically how Up and Down were already behaving. Now I'm thinking that Spacebar should have `& ~FL_WHEN_ENTER_KEY` removed so that it is treated more similarly to Up and Down (and Enter).

I updated the PR, let me know if you agree with the changes.

MatthiasWM:
Ok, I think I get it now. Just for the heck of it, I believe that the entire Browser_ event handling needs an overhaul. Furthermore, I think that Carthage must be destroyed.

PR #1399, opened 2026-04-05
Fix test/browser.cxx when_items list

dannye:
The display string says "**_FL_WHEN_ENTER_KEY_** + FL_WHEN_RELEASE_ALWAYS"
But the code says "int(**_FL_WHEN_ENTER_KEY_CHANGED_**)+int(FL_WHEN_RELEASE_ALWAYS)"

`FL_WHEN_ENTER_KEY_CHANGED` is not a single flag, but is actually `FL_WHEN_ENTER_KEY | FL_WHEN_NOT_CHANGED | FL_WHEN_CHANGED`.
(8 + 2 + 1 = 11)

`FL_WHEN_RELEASE_ALWAYS` is also not a single flag, but is `FL_WHEN_RELEASE | FL_WHEN_NOT_CHANGED`.
(4 + 2 = 6)

Both of these include `FL_WHEN_NOT_CHANGED` (2), and because the code uses `+` instead of `|`, this flag gets "double counted".
(11 + 6 = 17)

17 is equivalent to `FL_WHEN_CLOSED | FL_WHEN_CHANGED`, which is much different than the intended `FL_WHEN_ENTER_KEY + FL_WHEN_RELEASE_ALWAYS` (14).


MatthiasWM:
Thanks for the patch. You are indeed right.

PR #1403, opened 2026-04-05
Bugfix: Fl_Hold_Browser: Don't skip first line on down press if nothing is selected

dannye:
Pressing the down arrow key in a Fl_Hold_Browser while nothing is selected would skip straight from "line 0" to line 2, but in this scenario it makes much more sense for line 1 to become selected.

Before:

https://github.com/user-attachments/assets/0abfaa89-87c6-49b7-be76-54be46fcbe41

After:

https://github.com/user-attachments/assets/be7220a3-fdcf-4b1b-845e-7d195a1b6a81

This is my last bugfix for Fl_Browser_. Thanks for considering.

If #1396 is merged first, this PR will have merge conflicts, but I'll be happy to fix it.

### DIFF
--- a/src/Fl_Browser_.cxx
+++ b/src/Fl_Browser_.cxx
@@ -717,11 +717,13 @@ int Fl_Browser_::handle(int event) {
       if (type()==FL_HOLD_BROWSER) {
         switch (Fl::event_key()) {
         case FL_Down:
-          while ((l = item_next(l))) {
+          if (l1) l = item_next(l);
+          while (l) {
             if (item_height(l)>0) {
               select_only(l, when() & ~FL_WHEN_NOT_CHANGED);
               break;
             }
+            l = item_next(l);
           }
           return 1;
         case FL_Up:


dannye:
Another consequence of this bug is that when a list only contains 1 item and it is not yet selected, pressing the down arrow key does nothing. The item cannot be selected with just they keyboard.

dannye:
Fixed conflicts

MatthiasWM:
Thanks for the issue report and the proposed fix. I changed the original code a little bit for readability and added expected behavior for the Up button when nothing is selected.

dannye:
> and added expected behavior for the Up button when nothing is selected.

@MatthiasWM I didn't add behavior for the Up button when nothing is selected because it is not obvious to me what that behavior should be.

Your commit allows for "infinite wrapping" of the selected line while the Up button is held, but not when the Down button is held, which I do not think is desirable.

MatthiasWM:
Yeah, the wrapping was not intentional. I fixed that and I also mirrored the Down button behavior to select the top *visible* item. So Up without a previously selected item now select the bottom *visible* item. Not entirely sure if that is better than the bottom list item, but it's consistent with Down.

I am too tired for the callback/reason behavior at this point. I'll try to get back to that tomorrow.

dannye:
Thanks! It all looks good to me now. I agree selecting the bottom visible item is better.

PR #1406, opened 2026-04-12
Fluid: write code files only when the actually change (STR 1859)

MatthiasWM:
This prevents unnecessary file updates when content hasn't changed, reducing recompilation time in projects that depend on fluid-generated files. Uses std::ostringstrem before deciding if file differs.


fluid/io/Code_Writer.cxx
```diff
@@ -384,12 +384,25 @@ void Code_Writer::write_cc(const char *indent, int n, const char *c, const char
...
+
   va_list args;
   va_start(args, format);
-  vfprintf(header_file, format, args);
+  int n = vsnprintf(block_buffer_, block_buffer_size_, format, args);
+  if (n > block_buffer_size_) {
+    // Buffer was too small, reallocate and try again with the copy
+    block_buffer_size_ = n + 128;
+    if (block_buffer_) ::free(block_buffer_);
+    block_buffer_ = (char*)::malloc(block_buffer_size_+1);
```
RokerHRO, line 398: Why don't use realloc()?

MatthiasWM: Very true. It was a mindless copy and paste from existing code. I'll try to remember to update this.

Issue #1417, opened 2026-04-23
fl_add_symbol hand rolled hash system is buggy


MatthiasWM:
**Description of the Bug Report**
There is no need to implement hashing inside FLTK to store a fixed array of 100 or sy symbols by name.

**To Reproduce**
 - Currently fixed is incrementing the number of symbols, even if an existing symbol is reused.
 - Changing a symbol is rejected if the buffer is full, even if a symbol is merely replaced. 
 - If the array is relatively full, the hashing search function will prefer an empty entry over an existing entry with the same name.
 - The number of symbols is fixed.

**Operating System / Platform:**
Since 1.5 allows C++11, all the hashing code and all the search functionality can be replaced with `std::unordered_map` or `std::map`. See `static SYMBOL symbols[MAXSYMBOL];` in `fl_symbols.cxx:42`.

See the [https://cppreference.com/cpp/container/unordered_map](https://cppreference.com/cpp/container/unordered_map) example that implements exactly what we are doing here manually.

Since 1.4 is C++98 only, there can be no back port of this.

MatthiasWM:
This is now a Pull Request  #1418

PR #1418, opened 2026-04-23
Replace hand rolled symbol hashing with C++11 code (#1417)

MatthiasWM:


Albrecht-S:
@MatthiasWM News: see my commits (above) to your PR, as discussed. Details:

1. I added fl_remove_symbol().
2. I added another flag (`special`) in the table of symbols - which is NOT accessible by user code, i.e. it can't be set by fl_add_symbol(). It is only set by the table initialization.
3. I kept the special value `3` for `returnarrow`. This is an arbitrary value, we could change it if we like.
4. I did **not** replace `nullptr` in the table of drawing functions for `returnarrow` because the function  signature was different.
5. Removing symbols has not been tested by me. I just copied, pasted, and adjusted your code.
6. I did not yet `clang-format` the file intentionally. We should do that when we're satisfied with all changes.

MatthiasWM:
Thanks, Albrecht. I will do a few more changes by making the "returnarrow" special case a little less special.

Albrecht-S:
> I will do a few more changes by making the "returnarrow" special case a little less special.

Great, looking forward to your changes.

I suggest that we `clang-format` as the last operation before squashing and merging.

MatthiasWM:
`Symbol` previously used a `nullptr` function pointer as a special case to trigger pixel-level drawing (e.g. for the return arrow). That special case is now replaced by a proper call_with_rect flag and a dedicated `fl_add_symbol()` overload that accepts a rectangle-aware drawing function. The two drawing styles — fast pixel drawing (lines, text, bitmaps) and scalable vector drawing — are now symmetric, first-class options rather than a documented path and a hidden hack.

The Symbol constructors set `call_with_rect` automatically from the function pointer type, so there is no longer any way to register a symbol with a mismatched flag.

Albrecht-S:
Going to review and test ...

Albrecht-S:
@MatthiasWM Commit 53f6302ccc6ee0a2ba9136de55a71d02329adcf4 fixes typos and documentation issues.

From my POV this branch is technically OK so far.

~To do: we should still take the chance and **reformat the code with clang-format** before squashing and merging (rebasing). Please let me know if you want me to do the reformatting, squashing, and/or rebasing.~

**_EDIT 1:_** I had missed that you did already clang-format the source code. Everything good to go from me.

**_EDIT 2:_** The same applies to changes in .clang-format. :+1:

MatthiasWM:
I added a workaround for the wrong Doxygen link. This is a long standing bug (11 years) where the Doxygen parser bails if a function argument is a pointer to a function.

PR #1442, opened 2026-05-20
Paused: Add std::function callback support for Fl_Widget

MatthiasWM:
By automatically managing callback user data, I was now able to implement `std::function` based callbacks. This is great if the user wants to call a member function directly without writing a static callback trampoline. I also updated the documentation and added unit tests to `test/unittest`. 

As a side effect, I also added managed callback data to `Fl_Tree_Item` and `Fl_File_Chooser`. I might add `std::function` support to those two as well.

I invite anyone interested in doing a code review. I am sure this PR will be open for a few more days for testing before I merge it with master.

Example:

```
class MyButton : public Fl_Button {
  ...
  void beep(int pitch, double duration) { ... }
};
...
MyButton* btn = new MyButton(...);
btn->callback([this](Fl_Widget*) { beep(1000, 0.2); });
```

### DIFF
--- a/FL/Fl_File_Chooser.H
+++ b/FL/Fl_File_Chooser.H
@@ -1,7 +1,7 @@
 //
 // Fl_File_Chooser dialog for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2024 by Bill Spitzak and others.
+// Copyright 1998-2026 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
--- a/FL/Fl_File_Chooser.H
+++ b/FL/Fl_File_Chooser.H
@@ -54,14 +54,22 @@ public:
                    ///< effect for directories.
     DIRECTORY = 4  ///< Select a single, existing directory. Can be combined with CREATE.
   };
+protected:
+  /**
+   Values for flags_ field.
+  */
+  enum Flags {
+  AUTO_DELETE_USER_DATA = 1<<0, ///< automatically call `delete` on the user_data pointer when destroying this widget; if set, user_data must point to a class derived from the class Fl_Callback_User_Data
+};
 private:
   static Fl_Preferences *prefs_;
-  void (*callback_)(Fl_File_Chooser*, void *);
-  void *data_;
+  void (*callback_)(Fl_File_Chooser*, void *) = nullptr;
+  void *user_data_ = nullptr;
   char directory_[FL_PATH_MAX];
   char pattern_[FL_PATH_MAX];
   char preview_text_[2048];
   int type_;
+  unsigned int flags_ = 0;
   void favoritesButtonCB();
   void favoritesCB(Fl_Widget *w);
   void fileListCB();
@@ -135,7 +143,6 @@ private:
   static void cb_favOkButton(Fl_Return_Button*, void*);
 public:
   ~Fl_File_Chooser();
-  void callback(void (*cb)(Fl_File_Chooser *, void *), void *d = 0);
   void color(Fl_Color c);
   Fl_Color color();
   int count();
@@ -170,8 +177,11 @@ public:
   Fl_Fontsize textsize();
   void type(int t);
   int type();
+  void callback(void (*cb)(Fl_File_Chooser *, void *), void *d = 0);
+  void callback(void (*cb)(Fl_File_Chooser *, void *), Fl_Callback_User_Data* d, bool auto_free);
   void * user_data() const;
   void user_data(void *d);
+  void user_data(Fl_Callback_User_Data* d, bool auto_free);
   const char *value(int f = 1);
   void value(const char *filename);
   int visible();
@@ -249,6 +259,9 @@ public:
   Fl_Widget* add_extra(Fl_Widget* gr);
 protected:
   void show_error_box(int val);
+  unsigned int flags() const;
+  void set_flag(unsigned int c);
+  void clear_flag(unsigned int c);
 };
 FL_EXPORT char *fl_dir_chooser(const char *message,const char *fname,int relative=0);
 FL_EXPORT char *fl_file_chooser(const char *message,const char *pat,const char *fname,int relative=0);
--- a/FL/Fl_Widget.H
+++ b/FL/Fl_Widget.H
@@ -1,7 +1,7 @@
 //
 // Widget header file for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 1998-2025 by Bill Spitzak and others.
+// Copyright 1998-2026 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
@@ -93,9 +93,9 @@ struct FL_EXPORT Fl_Label {
  */
 class FL_EXPORT Fl_Callback_User_Data {
 protected:
-  Fl_Callback_User_Data() {} ///< Protected constructor
+  Fl_Callback_User_Data() = default; ///< Protected constructor
 public:
-  virtual ~Fl_Callback_User_Data() { } ///< Destructor
+  virtual ~Fl_Callback_User_Data() = default; ///< Destructor
   /// Clone method to allow copying of user data when needed.
   /// *Must* be overridden by derived classes that are used in Fl_Tree_Item.
   virtual Fl_Callback_User_Data* clone() const { return new Fl_Callback_User_Data(*this); }
@@ -107,6 +107,7 @@ protected:
   std::function<void(Fl_Widget*)> func; ///< The lambda function to be called
 public:
   Fl_Std_Function_User_Data(std::function<void(Fl_Widget*)> f) : func(std::move(f)) {}
+  ~Fl_Std_Function_User_Data() override = default; ///< Default destructor
   Fl_Callback_User_Data* clone() const override { return new Fl_Std_Function_User_Data(*this); } ///< Clone method for Lambda captures
   void call(Fl_Widget* widget) { func(widget); } ///< Call the stored lambda function
 };
--- a/FL/Fl_Widget.H
+++ b/FL/Fl_Widget.H
@@ -97,7 +97,7 @@ protected:
 public:
   virtual ~Fl_Callback_User_Data() { } ///< Destructor
   /// Clone method to allow copying of user data when needed.
-  /// *Must* be overridden by derived classes that are used in Fl_Menu_Item.
+  /// *Must* be overridden by derived classes that are used in Fl_Tree_Item.
   virtual Fl_Callback_User_Data* clone() const { return new Fl_Callback_User_Data(*this); }
 };
 
@@ -806,9 +806,9 @@ public:
   /** Sets the widget callback to a \c std::function.
 
     Accepts any callable that matches the callback signature, including
-    capturing lambdas and \c std::bind expressions. Unlike the classic
-    C-style callback, there is no user data field — captures provide
-    type-safe access to context data without casting.
+    capturing lambdas and \c std::bind expressions. The user data field
+    is used internally and must not be overwritten; use captures instead
+    for type-safe access to context data without casting.
 
     \param[in] cb new callback
   */
--- a/FL/Fl_Widget.H
+++ b/FL/Fl_Widget.H
@@ -94,6 +94,9 @@ protected:
   Fl_Callback_User_Data() {} ///< Protected constructor
 public:
   virtual ~Fl_Callback_User_Data() { } ///< Destructor
+  /// Clone method to allow copying of user data when needed.
+  /// *Must* be overridden by derived classes that are used in Fl_Menu_Item.
+  virtual Fl_Callback_User_Data* clone() const { return new Fl_Callback_User_Data(*this); }
 };
 
 
--- a/FL/Fl_Widget.H
+++ b/FL/Fl_Widget.H
@@ -101,15 +101,15 @@ public:
   virtual Fl_Callback_User_Data* clone() const { return new Fl_Callback_User_Data(*this); }
 };
 
-/** Callback user data to manage Lambda expressions with captures. */
+/** Callback user data to manage std::function based callbacks. */
 class FL_EXPORT Fl_Std_Function_User_Data : public Fl_Callback_User_Data {
 protected:
-  std::function<void(Fl_Widget*)> func; ///< The lambda function to be called
+  std::function<void(Fl_Widget*)> func; ///< The std::function to be called
 public:
   Fl_Std_Function_User_Data(std::function<void(Fl_Widget*)> f) : func(std::move(f)) {}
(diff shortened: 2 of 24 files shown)


src/Fl_File_Chooser.cxx
```diff
@@ -62,7 +62,7 @@ static const unsigned char idata_new[] =
 {0,0,120,0,132,0,2,1,1,254,1,128,49,128,49,128,253,128,253,128,49,128,49,
 128,1,128,1,128,255,255,0,0};
 static Fl_Image *image_new() {
-  static Fl_Image *image = NULL;
+  static Fl_Image *image = 0L;
```
RokerHRO: Why not `nullptr`?

FL/Fl_File_Chooser.H
```diff
@@ -54,14 +54,22 @@ public:
                    ///< effect for directories.
     DIRECTORY = 4  ///< Select a single, existing directory. Can be combined with CREATE.
   };
+protected:
+  /**
+   Values for flags_ field.
+  */
+  enum Flags {
+  AUTO_DELETE_USER_DATA = 1<<0, ///< automatically call `delete` on the user_data pointer when destroying this widget; if set, user_data must point to a class derived from the class Fl_Callback_User_Data
+};
```
RokerHRO, line 63: Instead of this flag I'd prefer just a user-given callback that is called in the File_Chooser's destructor.  That function can be
* a function or lambda that does your `delete static_cast<Fl_Callback_User_Data*>(user_data);`
* a no-op
* anything else the user wants/needs. (In most of the use case I'd assume it is just a lambda that can `delete` or `free()`  the `user_data` even if it is not derived from a special FLTK class, e.g. 3rd-party C data structs.)

src/Fl_File_Chooser.cxx
```diff
@@ -439,12 +434,33 @@ int Fl_File_Chooser::type() {
   return (type_);
 }
 
+void Fl_File_Chooser::callback(void (*cb)(Fl_File_Chooser *, void *), void *d ) {
+  callback_ = cb;
+  user_data(d);
```
RokerHRO, line 439: I thought `user_data` is a `void*` and not a function… how can you "call" it?

src/Fl_File_Chooser.cxx
```diff
@@ -439,12 +434,33 @@ int Fl_File_Chooser::type() {
...
 void * Fl_File_Chooser::user_data() const {
-  return (data_);
+  return user_data_;
 }
 
 void Fl_File_Chooser::user_data(void *d) {
-  data_ = d;
+  if ((flags_ & AUTO_DELETE_USER_DATA) && user_data_) {
+    clear_flag(AUTO_DELETE_USER_DATA);
+    delete static_cast<Fl_Callback_User_Data*>(user_data_);
```
RokerHRO, line 454: check for `nullptr`  is not necessary for `delete` because `delete nullptr` is a well-defined no-op.

src/Fl_Widget.cxx
```diff
@@ -440,6 +439,13 @@ void Fl_Widget::bind_deimage(Fl_Image* img) {
   bind_deimage( (img != NULL) );
 }
 
+void Fl_Widget::do_std_function_callback_(Fl_Widget* w, void* data) {
+  auto std_function_callback = static_cast<Fl_Std_Function_User_Data*>(w->user_data());
```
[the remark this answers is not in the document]
MatthiasWM: That is a difficult one. Yes, if `do_callback` is called with a different widget, this will crash. But if `do_callback` is called with different data, it will also crash. I will have to check `Fl_Menu_` based callbacks if they modify either of those two values.

FL/Fl_Widget.H
```diff
@@ -94,8 +96,20 @@ protected:
   Fl_Callback_User_Data() {} ///< Protected constructor
 public:
   virtual ~Fl_Callback_User_Data() { } ///< Destructor
+  /// Clone method to allow copying of user data when needed.
+  /// *Must* be overridden by derived classes that are used in Fl_Menu_Item.
+  virtual Fl_Callback_User_Data* clone() const { return new Fl_Callback_User_Data(*this); }
```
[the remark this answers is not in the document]
MatthiasWM: Ah yes. This is valid for `Fl_Tree_Item`, not `Fl_Menu_Item`. I fixed the comment. It's false to omit overriding `clone()`, but the error is limited to Tree Items. SO I don;t want to be too radical about it.

FL/Fl_Widget.H
```diff
@@ -787,6 +803,20 @@ public:
...
+    capturing lambdas and \c std::bind expressions. Unlike the classic
+    C-style callback, there is no user data field — captures provide
+    type-safe access to context data without casting.
+
+    \param[in] cb new callback
+  */
+  void callback(std::function<void(Fl_Widget*)> f) {
+    callback_ = do_std_function_callback_;
+    user_data(new Fl_Std_Function_User_Data(std::move(f)), true);
+  }
```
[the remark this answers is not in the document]
MatthiasWM: Thanks. I fixed that in the inline docs and the main docs.

test/unittest_core.cxx
```diff
@@ -165,7 +162,44 @@ TEST(Fl_Callback_Macros, FL_INLINE_CALLBACK) {
...
+TEST(std_function_callbacks, lambda) {
+  Fl_Group::current(NULL);
+  Fl_Button *btn = new Fl_Button(10, 10, 100, 100);
+  int capture_me = 36;
+  btn->callback([capture_me](Fl_Widget*) {
+    ok = capture_me;
+  });
+  btn->do_callback();
+  EXPECT_EQ(ok, 36);
+  delete btn;
```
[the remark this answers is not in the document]
MatthiasWM: Fixed

src/Fl_File_Chooser.cxx
MatthiasWM: It's generated form an .fl file. I did not see this and will change it into nullptr

src/Fl_File_Chooser.cxx
MatthiasWM: `user_data_` is the void pointer, `user_data` is the function to set `user_data_` (trailing underscore).

src/Fl_File_Chooser.cxx
MatthiasWM: Yes, I am aware of that. I wanted to avoid the static_cast, which is probably unnecessary, too. It's potentially more involved than reinterpret_cast though. I'd have to look at the generated assembler code if this saves anything/

FL/Fl_File_Chooser.H
MatthiasWM: I chose this implementation for Fl_Widget because it doesn't add another callback. Adding another callback reason would have required new `when` flags, so I needed up using this scheme for Fl_Widget.

I transferred the original idea to Fl_Fle_Chooser to stay consistent. 

Does that resolve it?

src/Fl_File_Chooser.cxx
MatthiasWM: OK, I looked it up. Static_cast only adds code for multiple inheritance. I will remove the `if` clauses

FL/Fl_File_Chooser.H
RokerHRO: I don't like that interface. The `enum Flag` with only one enumerator  `AUTO_DELETE_USER_DATA` looks just clumsy to me.

Moreover: This interface is too limited for real-world benefit: AFAIK for those "user data" existing data structs are used, e.g. coming from 3rd party software, often C libraries with special `freeFoo()` functions for their data types.

So with your approach the users are required to pack their user_data into yet another self-made class, derived from `Fl_Callback_User_Data`, which needs several lines of "boiler plate code", only to make the "auto delete" feature work. :-/

In my approach the `callback()` function just get a 2nd parameter for the deleter, which often can just be a lambda expression. And: That 2nd parameter can even be optional, if the default value does "the right thing".

That interface would remind on the c'tor of `std::unique_ptr<>`, so many C++ users should be familiar with this idiom.

See https://en.cppreference.com/cpp/memory/unique_ptr/unique_ptr  (overload 2 and 3).

And yes, we can discuss, what "the right thing" should be: `delete` or `do_nothing`. :-)

src/Fl_File_Chooser.cxx
RokerHRO: `delete` on a `void*` would not do virtual dispatch to find the right destructor of the derived class the user_data points to:

https://en.cppreference.com/cpp/language/delete#Notes

https://www.google.com/search?q=c%2B%2B+delete+on+void%2A

FL/Fl_Widget.H
```diff
@@ -97,7 +97,7 @@ protected:
 public:
   virtual ~Fl_Callback_User_Data() { } ///< Destructor
```
RokerHRO: Why empty d'tor and not default d'tor?

RokerHRO:
Some more aspects:
1) std::function is quite heavy. A "non-owning" functor would do the same and is much smaller and faster, consider something like https://github.com/TartanLlama/function_ref
What do you think?

2) What is the use-case of that "user data delete" feature? By using functors (no matter whether `std::function`, a `std::bind` or a lambda expression, we can "hold" arbitrary user data in the functor object itself and don't need `user_data` anymore.
And if that "user data" shall have the same live time as the widget, I'd say: just derive from that widget class and put your additional data into the derived class as a member and you're done. FLTK is C++, not C.  ;-)

FL/Fl_Widget.H
MatthiasWM: It was written before we decided to got from C++98 to C++11. I'll update that.

src/Fl_File_Chooser.cxx
RokerHRO: Oops... my fault. Sorry. I need new O^O

FL/Fl_File_Chooser.H
MatthiasWM: As for Fl_Callback_User_Data, this has been around since 1.4 even before we decided to go to C++11. I can't fundamentally change the API in 1.5. Adding a "delete user data" may be possible on top of the existing code, but it adds to Fl_Widget's size, so we have to discuss that.

The boilerplate code to derive a class from Fl_Callback_User_Data usually needs to be written only once per application. But yeah, it's an additional step.

As for the single flag in a 32 bit integer, I copied the basic idea from Fl_Widget for consistency and expandability while keeping the ABI intact. If I was building a library from scratch, I would do it differently (`int auto_delete_user_data :1;` probably, or bool).

RokerHRO: 1) the user has to define a class for _every different type_ of `user_data`.  That won't be hundreds of types, but more than one, I'd say. :-/

2) It is a good question how many users are really affected by such an API change. How often is that `Fl_Callback_User_Data` feature used? 

3) storing a `std::function<..>` as member in the widget costs more than a 2nd function pointer, because `sizeof(std::function<R(T)>)` is 32 on x86_64, and your `enum` also adds size to the widget class.

Btw, a lambda expression (without capture) would be converted automatically into a function pointer, so it can be used with the old API already very easily. Therefore the benefit of a more generic API with `std::function<>` might be smaller than it seems to be?

Maybe some example code would help to see the impact?

MatthiasWM:
Thanks for all the great suggestions! So to sum it up: 

- I must solve the do_callback situation that Copilot found when users call do_callback, but change the value of Fl_Widget and/or user_data which may lead to a crash
- I can add a delete-user-data callback to Fl_Widget to make all this more universal at the cost of adding a pointer to every widget, even if it's never used. Thinking... .
- change the Fl_File_Chooser class `flags` API to modern(er) C++ or stick with what Fl_Widget does?

@RokerHRO 
1: yeah, std::function is somewhat heavy, but it's standard. At least it's only allocated if it is really needed, and the old minimal callback system is still in place. The TartanLlama stuff seems to be very compiler dependent, which I really want to avoid.

2: is a result of 1. If I make std::function part of Fl_Widget, I add 32 bytes to Fl_Widget vs. 16 bytes currently. Again, not huge, but something to think about. The original API is from C++98 times, as I said before.

So I guess the bigger question is, throw out all the old callback code, replace the member variables with std::function, and rebuild the legacy API as well as possible? Or leave that to 1.6?

RokerHRO:
@MatthiasWM : Thank you for your summary :-)
You're right: Compatibility is always a problem and often forces one to stuck with a sub-optimal interface or implementation. :-/

I'd like to know the very problem this new callback aims to solve. What is not possible (or not as elegant?) with the old, existing interface?

I have some vague ideas in mind (e.g. std::variant<> holding alternatively
* a raw function pointer with `Fl_Widget*` argument
* a raw function pointer with 2 arguments: `Fl_Widget*` and `void*` 
* maybe a std::function<>?
* some "smart" Functor that automatically cleans-up the `user_data` when the widget ends its life
* … anything else? )
and see how to deal with that as the user (and how to implement) and how much `sizeof(Fl_Widget)` would increase. :-D

MatthiasWM:
By now, this PR has become much too big. I will try to split things into functional smaller commits later.

Albrecht-S:
@MatthiasWM Thanks for this PR. Although I'm not convinced yet that this is really necessary, particularly after reading the CI comments and @RokerHRO's comments. My first gut feeling is that we should rethink at least the auto-delete option which obviously makes things much more complicated. Why not leave deleting the callback stuff to the programmer?

I also like @RokerHRO's suggestion to subclass and add all relevant parameters to the class instance (of the user derived class) which would be entirely independent of the FLTK core.

These are only first thoughts, we really need some time (and discussion) to find out what we want and what the best API would be to extend the current callback system. I could also imagine to make an entire new callback feature (API) that does not depend on the current one (leaving the old one in place, of course). ... ...

MatthiasWM:
Thanks, @RokerHRO and @Albrecht-S . There are two concrete use cases I was trying to cover. The first one is simple: avoid stacked of callbacks away from where the widget is defined. Allowing lambdas with captures, writing a callback is a lot easier and keeps code together.

The second reason is secondary use of user_data. Years ago, when I wanted to use captures, but FLTK remained stuck in C++98, I simulated captures by implementing Fl_Callback_User_Data and a group of macros. Back then, it made things easier. Today, it's yet another legacy API to support. I am not even sure if anyone out there ever used this. Today, I would simply replace callback_/user_data_ with std::function and call it a day. 

There is a third unofficial use that I have seen for language bindings, where user_data stores the link back to construct that connects the C++ FLTK widget to a language representation.

Why don't people derive their own class, so they don't have to use user_data at all? Probably because it requires a lot of boilerplate code and generates a new and identical table for every derived widget. It makes sense for language bindings though, because all virtual function must be wrapped anyway.

I'll take a break form this for a day or two and rethink the concept.

src/Fl_File_Chooser.cxx
MatthiasWM: This fixed in Fluid, creating nullptr instead of 0L or NULL.

MatthiasWM:
This PR is superseded by "Add lambdas with captures as callback option." #1597

PR #1448, opened 2026-05-25
Improve Fluid formatting

MatthiasWM:
* change pointer type formatting
* change 0, NULL, and 0L into nullptr

### DIFF
--- a/FL/Fl_File_Chooser.H
+++ b/FL/Fl_File_Chooser.H
@@ -74,63 +74,63 @@ private:
 public:
   Fl_File_Chooser(const char *pathname, const char *pattern, int type_val, const char *title);
 private:
-  Fl_Double_Window *window;
+  Fl_Double_Window* window;
   inline void cb_window_i(Fl_Double_Window*, void*);
   static void cb_window(Fl_Double_Window*, void*);
-  Fl_Choice *showChoice;
+  Fl_Choice* showChoice;
   inline void cb_showChoice_i(Fl_Choice*, void*);
   static void cb_showChoice(Fl_Choice*, void*);
-  Fl_Menu_Button *favoritesButton;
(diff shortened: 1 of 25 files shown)


FL/Fl_File_Chooser.H
```diff
@@ -74,63 +74,63 @@ private:
...
+  Fl_Choice* showChoice;
   inline void cb_showChoice_i(Fl_Choice*, void*);
   static void cb_showChoice(Fl_Choice*, void*);
-  Fl_Menu_Button *favoritesButton;
+  Fl_Menu_Button* favoritesButton;
   inline void cb_favoritesButton_i(Fl_Menu_Button*, void*);
   static void cb_favoritesButton(Fl_Menu_Button*, void*);
 public:
-  Fl_Button *newButton;
+  Fl_Button* newButton;
```
erco77, line 87: I think I always put the asterisk (\*) next to the variable instead of the datatype in declarations, to remove that ambiguity in situations like `Fl_Button* a,b,c;`, where it might appear that a,b,c are pointers, but only 'a' would actually be.

So it seems better to put the asterisk with the variable, like `Fl_Button *a, *b, *c;` to avoid all that, should someone decide to add comma delimited extra variables later. I recall Doug Roble bringing that up in an old DD software meeting..!

MatthiasWM: I had a long discussion with Albrecht about this, and I also remember learning the "right star" style at DD.

"Modern C++" - whatever that is - seems to prefer, often insist on "left star". Reasoning that the type declaration should stay together, and the variable name stands by itself. 

The "char* a, b, c;" should never be an issue because those style guides also demand that every variable declaration gets its own line. Or even better, we don't use any raw pointers at all.

PR #1451, opened 2026-05-28
Tablet PR

ggarra13:
This PR should add tablet support for Wayland.  Code by ClaudeAI untested.  Should compile on Ubuntu 26.04 LTS.

### DIFF
--- a/CMake/options.cmake
+++ b/CMake/options.cmake
@@ -321,10 +321,12 @@ if(UNIX)
     if(EXISTS ${PROTOCOLS}/staging/cursor-shape/cursor-shape-v1.xml AND
         EXISTS ${PROTOCOLS}/stable/tablet/tablet-v2.xml)
       set(HAVE_CURSOR_SHAPE 1)
+      set(HAVE_TABLET 1)
       message(STATUS "Found dev files for Wayland protocols 'Cursor shape' and 'Tablet'")
       message(STATUS "  ==> option FLTK_USE_DBUS can be turned OFF if 'Cursor shape'-enabled wayland compositor is used.")
     else()
       set(HAVE_CURSOR_SHAPE 0)
+      set(HAVE_TABLET 0)
     endif()
     if(FLTK_BACKEND_X11)
       include(FindX11)
@@ -501,6 +503,7 @@ endif()
 #  - FLTK_HAVE_PEN_SUPPORT  : final result for building Pen/Tablet support,
 #                             also used to set config variable in config.h
 
+set(FLTK_HAVE_PEN_SUPPORT 0)
 if(FLTK_OPTION_PEN_SUPPORT)
   if(WIN32)
     try_compile(PEN_DRIVER_SUPPORTED
--- a/config.h.in
+++ b/config.h.in
@@ -357,7 +357,6 @@
 
 #cmakedefine01 HAVE_CURSOR_SHAPE
 
-
 /*
  * FLTK_HAVE_PEN_SUPPORT
  *
@@ -370,4 +369,3 @@
  *
  */
 
-#cmakedefine FLTK_HAVE_PEN_SUPPORT 1
--- a/fl_config.h.in
+++ b/fl_config.h.in
@@ -159,4 +159,15 @@
 #cmakedefine01 FLTK_USE_SVG
 
 
+/*
+ * FLTK_HAVE_PEN_SUPPORT
+ *
+ * Do we have pen/tablet support for the current platform?
+ * See CMake option FLTK_OPTION_PEN_SUPPORT, but note that a build test
+ * (try_compile()) is performed to test if the option is available
+ *
+ */
+
+#cmakedefine FLTK_HAVE_PEN_SUPPORT 1
+
 #endif /* _FL_fl_config_h_ */
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -869,7 +869,7 @@ if(UNIX AND FLTK_USE_WAYLAND)
     add_custom_command(
       OUTPUT tablet-protocol.c # tablet-client-protocol.h
       COMMAND wayland-scanner private-code  ${INFILE} tablet-protocol.c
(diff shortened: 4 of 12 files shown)


ggarra13:
@MatthiasWM I closed and now reopened the pull request as I had incorrectly changed Fl_x.cxx and had not properly tested X11 and Wayland separately (My workflow builds for both).

MatthiasWM:
Ok, so I was able to set up an Ubuntu 26.4 VM and hook up my Intuos PTH-450. So here are my spontaneous findings without doing any in-depth research.

1. The patch compiles without errors or warning.
2. It does not FLTK_HAVE_PEN_SUPPORT into account which could be set to "no" by the user - not hard to fix
3. The driver itself is logical. It's looks like generic code that comes from a Wayland reference implementation, mixed with what was taken from the Cocoa driver.
4. The driver does the first initialization well. It find the tablet and registers the first time it sees the pen.
5. The entire app gets stuck after the first pen-down event. After thinking about all the complicated reasons why that could be, the bug was elsewhere entirely. Turns out, calling Cairo_arc with a radius of 0 locks up test/penpal. Easy fix.
6. After that fix, drawing with the pen works really well. Pressure works, eraser works, buttons seem to work well enough.
7. FIXME: vertical tilt is flipped
8. FIXME: mouse emulation of pen events no longer works, tell Wayland if a pen event was not used

Edit: Wayland has no way to return unused events. This has to be hacked manually, even for the window decoration. This complicates things, because the current driver always installs a tablet listener, disabling pen-to-mouse for *all* FLTK apps. I'll have to think about this :-/ .

ggarra13:
> Ok, so I was able to set up an Ubuntu 26.4 VM and hook up my Intuos PTH-450. So here are my spontaneous findings without doing any in-depth research.

Thanks for working on this.

> 
> 1. The patch compiles without errors or warning.

Good. 

> 2. It does not FLTK_HAVE_PEN_SUPPORT into account which could be set to "no" by the user - not hard to fix

Yes.  I did not want to step on @Albrecht-S really tidy CMake work.  Also since my CMakeLists in my fork has the Vulkan port, I wanted to stay in.

> 3. The driver itself is logical. It's looks like generic code that comes from a Wayland reference implementation, mixed with what was taken from the Cocoa driver.

Yes.  My prompt to Claude was "port" your Cocoa.mm code to Wayland.  Previous to sending the prompt, I had set some "stubs" functions/variables in the appropiate Wayland files to help it along. 

> 4. The driver does the first initialization well. It find the tablet and registers the first time it sees the pen.

Good.

> 5. The entire app gets stuck after the first pen-down event. After thinking about all the complicated reasons why that could be, the bug was elsewhere entirely. Turns out, calling Cairo_arc with a radius of 0 locks up test/penpal. Easy fix.

I take it was a bug in the penpal.cxx application, right?

> 6. After that fix, drawing with the pen works really well. Pressure works, eraser works, buttons seem to work well enough.

Good.

> 7. FIXME: vertical tilt is flipped

Simple to fix, I imagine.

> 8. FIXME: mouse emulation of pen events no longer works, tell Wayland if a pen event was not used

Did you find a way around this one?

> Edit: Wayland has no way to return unused events. This has to be hacked manually, even for the window decoration. This complicates things, because the current driver always installs a tablet listener, disabling pen-to-mouse for _all_ FLTK apps. I'll have to think about this :-/ .

Not sure I follow you there completely.  Under Wayland, each FLTK application runs independently.  Installing a tablet listener for any application should not change that.

ggarra13:
Changes:

- Made it work properly under C++11.
- Moved FLTK_HAVE_PEN_SUPPORT to fl_config.h to make it public.
- Changed CMakeLists.txt and CMake/options.cmake to allow compilation without pen support.  Tested on Wayland only.
- Fixed penpal.cxx to not allow radius of less than 1.
- Made registration of tablet happen on first subscription instead of on Wayland initialization.
- Added ClaudeAI code to presumably handle normal FLTK events if pen events fail.  Matt, please test if these are now working properly.
- Inverted event tilt on Y.

MatthiasWM:
thanks for all the effort. The changes seem to fix exactly what I found. 

I can’t look at code at the moment or compile anything. I will be without a laptop for at least 10 days :-/ . I‘ll try to read my mail regularly though.

MatthiasWM:
Hi @ggarra13 . Any reason your PR is closed? I was just about to go through the changes and check if I can merge the PR. So far, everything seems great. I still have to look at the build system stuff, but the code itself;f seems good enough to merge. Smaller issues can be fixed later as they emerge.

MatthiasWM:
Note to self: check if the menubar changes form this commit: https://github.com/fltk/fltk/pull/1451/changes/a20589cba1f0f031c945c7e07f461869b7b49694 need to be back ported to Windows and macOS!

The mouse fallback should also allow for Right Mouse Button, maybe the upper barrel button? How is ist emulated in the Wayland fallback? Any other emulations (use `test/keyboard`).

For a later commit: Wayland has a pen wheel dataset. How can we add that to this driver? How can we add this to the API? Do other platforms have this option too?

ggarra13:
> Hi @ggarra13 . Any reason your PR is closed? I was just about to go through the changes and check if I can merge the PR. So far, everything seems great. I still have to look at the build system stuff, but the code itself;f seems good enough to merge. Smaller issues can be fixed later as they emerge.

Yes, I submitted a new PR (#1458) with both a Windows and a Wayland driver, both tested with my (cheaper) tablet than yours.

MatthiasWM:
> > Hi @ggarra13 . Any reason your PR is closed? I was just about to go through the changes and check if I can merge the PR. So far, everything seems great. I still have to look at the build system stuff, but the code itself;f seems good enough to merge. Smaller issues can be fixed later as they emerge.
> 
> Yes, I submitted a new PR (#1458) with both a Windows and a Wayland driver, both tested with my (cheaper) tablet than yours.

Oh, Ok, I see the other PR now. Thanks.

As a rule of thumb, I like to keep PRs to one issue and one platform, just to keep it comprehendible in its size. I understand the Wayland and Windows issues are related in this specific case. But it getting huge ;-)

PR #1452, opened 2026-05-31
Bugfix: Ensure tooltip window appears on the same screen as its widget

dannye:
Right now, tooltips suffer from "ambiguous coordinates" and therefore can appear on the wrong screen. Tooltip windows need to have their screen_num set using the widget's screen_num.

Before:

https://github.com/user-attachments/assets/b8964bfe-e5d7-4dec-88cb-e93176b4c8c2

After:

https://github.com/user-attachments/assets/c92afda7-0a3c-498f-b48c-d26319ee845f

### DIFF
--- a/src/Fl_Tooltip.cxx
+++ b/src/Fl_Tooltip.cxx
@@ -105,17 +105,22 @@ void Fl_TooltipBox::layout() {
       oy += p->y();
     }
   }
+  screen_num(Fl_Tooltip::current()->top_window()->screen_num());
   if (Fl::screen_driver()->screen_boundaries_known()) {
-    int scr_x, scr_y, scr_w, scr_h;
-    Fl::screen_xywh(scr_x, scr_y, scr_w, scr_h);
-    if (ox+ww > scr_x+scr_w) ox = scr_x+scr_w - ww;
-    if (ox < scr_x) ox = scr_x;
-    if (currentTooltipH > 30) {
-      if (oy+hh > scr_y+scr_h) oy -= 23+hh;
-    } else {
-      if (oy+hh > scr_y+scr_h) oy -= (4+hh+currentTooltipH);
+    int mx, my;
+    int nscreen = Fl::screen_driver()->get_mouse(mx, my);
+    if (screen_num() == nscreen) {
+      int scr_x, scr_y, scr_w, scr_h;
+      Fl::screen_driver()->screen_xywh(scr_x, scr_y, scr_w, scr_h, nscreen);
+      if (ox+ww > scr_x+scr_w) ox = scr_x+scr_w - ww;
+      if (ox < scr_x) ox = scr_x;
+      if (currentTooltipH > 30) {
+        if (oy+hh > scr_y+scr_h) oy -= 23+hh;
+      } else {
+        if (oy+hh > scr_y+scr_h) oy -= (4+hh+currentTooltipH);
+      }
+      if (oy < scr_y) oy = scr_y;
     }
-    if (oy < scr_y) oy = scr_y;
   }
 
   resize(ox, oy, ww, hh);
--- a/src/Fl_win32.cxx
+++ b/src/Fl_win32.cxx
@@ -2367,7 +2367,7 @@ void Fl_WinAPI_Window_Driver::makeWindow() {
     if (!force_position()) {
       xp = yp = CW_USEDEFAULT;
     } else {
-      if (!Fl::grab()) {
+      if (!Fl::grab() && !w->tooltip_window()) {
         xp = xwm;
         yp = ywm;
         x(int(round(xp / s)));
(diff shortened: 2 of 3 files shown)


dannye:
Actually, it would be even better if the tooltip appeared on the same screen as the widget even if the parent window is across two different screens and the widget is not on the window's "main" screen. This would make tooltip windows act more like "menutitle windows" rather than "menu windows". And this is how tooltip windows are already working on master but only when the screen's scales are equal (ie, when the ambiguous coordinates are not a problem).

src/Fl_Tooltip.cxx
```diff
@@ -105,17 +105,22 @@ void Fl_TooltipBox::layout() {
       oy += p->y();
     }
   }
+  screen_num(Fl_Tooltip::current()->top_window()->screen_num());
```
dannye, line 108: Issue # 1 is that tooltip windows did not have an explicit screen number, and so their coordinates were ambiguous. This could cause tooltip windows to appear on the wrong screen and with the wrong scale.

By setting the tooltip window's screen number equal to the current widget's screen number, the coordinates are no longer ambiguous and the tooltip window will always inherit the correct scale.

This is very similar to what is done for menu windows:

https://github.com/fltk/fltk/blob/722e2477a1ac8a82d3d75f726b50f9fb7ab0f680/src/Fl_Menu.cxx#L186-L187

src/Fl_Tooltip.cxx
```diff
@@ -105,17 +105,22 @@ void Fl_TooltipBox::layout() {
...
   if (Fl::screen_driver()->screen_boundaries_known()) {
-    int scr_x, scr_y, scr_w, scr_h;
-    Fl::screen_xywh(scr_x, scr_y, scr_w, scr_h);
```
dannye, line 110: Issue # 2: Previously, this code was attempting to clamp `ox` and `oy` to the _mouse's_ screen's area.

But the initial value of `ox` and `oy` are scaled coordinates of the _window's_ screen's scale. If the scales of the mouse's screen and of the window's screen are not equal, this "clamping" can move the tooltip window an unexpectedly large amount.

The fix (or, a good enough fix, I would argue) is to only attempt to clamp the tooltip window on the screen if the mouse's screen and the window's screen are the same screen. This keeps "current" behavior for all ordinary scenarios and improves the behavior for less ordinary scenarios (ie, when the window is mostly on screen A but the mouse (and the tooltip's widget) is on screen B then don't try to clamp the tooltip window; just let the tooltip window always be placed very near the mouse and always with the correct scale).

src/Fl_win32.cxx
```diff
@@ -2367,7 +2367,7 @@ void Fl_WinAPI_Window_Driver::makeWindow() {
     if (!force_position()) {
       xp = yp = CW_USEDEFAULT;
     } else {
-      if (!Fl::grab()) {
+      if (!Fl::grab() && !w->tooltip_window()) {
```
dannye, line 2370: Issue # 3 is that in `makeWindow()` for win32 this `if` block was (redundantly?) clamping the tooltip window onto its "assigned" screen. But the assigned screen is now the widget's window's screen, which is not necessarily the same as the mouse's screen (where the tooltip window will appear).

Since this clamping was mostly redundant anyway, simply don't perform this clamping for tooltip windows on win32.

dannye (commented):
I made the fixes that I mentioned in my previous comment. I'm very happy with the result.

@ManoloFLTK I would be very interested in your opinion on this if you are willing to take a look. Thank you.

Albrecht-S:
@dannye Thanks for reporting this issue and for suggesting a fix.

I can't follow the code changes right now (lack of time), hence here's only a heads-up:

Since the fix for this issue affects both platform specific (Windows) and platform independent code, the OP or any dev taking this issue (PR) should thoroughly check if other platforms are affected as well. Even the platform specific code might have been copied or worked on before the platform code split (in "drivers") so the potentially faulty code might be in the implementations of other platforms as well.

dannye:
> Since the fix for this issue affects both platform specific (Windows) and platform independent code, the OP or any dev taking this issue (PR) should thoroughly check if other platforms are affected as well.

@Albrecht-S Thanks, I added the equivalent fix to X11 (tested with and without Cairo) which now behaves the same as win32.

I tested on macOS and found no issues.

I tested on Wayland and found that tooltips do behave a bit oddly when a window is placed across two screens with different scales, but it's not any worse than how menu windows already behave. Since this PR is not about making any changes to menu windows, I would like to look into fixing the edge cases with Wayland separately in another PR, if that is acceptable.

MatthiasWM:
I can't stress hard enough that I hate the ambiguous coordinates problem.

dannye:
@Albrecht-S @ManoloFLTK I would be most appreciative if this fix could be merged before the release of FLTK 1.5.

PR #1457, opened 2026-06-13
Updated build.yml using Ninja and aarch64 architectures.

ggarra13:
As both linux aarch64 and windows aarch64 machines are soon to hit the streets, I created a pull request for them.

I also updated the build to use the Ninja generator which is usually faster than "make" or "MSBuild".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,24 @@ jobs:
     - uses: actions/checkout@v6
 
     - name: Install prerequisites
-      run: sudo apt-get update -y && sudo apt-get install -y libasound2-dev libglu1-mesa-dev libcairo2-dev libwayland-dev wayland-protocols libdbus-1-dev libxkbcommon-dev libegl-dev libopengl-dev libpango1.0-dev libdecor-0-dev libxfixes-dev libxcursor-dev libxinerama-dev ninja-build
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y \
+        libasound2-dev \
+        libglu1-mesa-dev \
+        libcairo2-dev \
+        libwayland-dev \
+        wayland-protocols \
+        libdbus-1-dev \
+        libxkbcommon-dev \
+        libegl-dev \
+        libopengl-dev \
+        libpango1.0-dev \
+        libdecor-0-dev \
+        libxfixes-dev \
+        libxcursor-dev \
+        libxinerama-dev \
+        ninja-build
 
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,18 @@ jobs:
     - uses: actions/checkout@v6
 
     - name: install prerequisites
-      run: sudo apt-get update -y && sudo apt-get install -y libasound2-dev libglew-dev libglu1-mesa-dev libcairo2-dev libx11-dev libxcursor-dev libxft-dev libxinerama-dev ninja-build
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y \
+        libasound2-dev \
+        libglew-dev \
+        libglu1-mesa-dev \
+        libcairo2-dev \
+        libx11-dev \
+        libxcursor-dev \
+        libxft-dev \
+        libxinerama-dev \
+        ninja-build
 
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
@@ -36,7 +47,17 @@ jobs:
       # Note the current convention is to use the -S and -B options here to specify source
       # and build directories, but this is only available with CMake 3.13 and higher.
       # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
-      run: cmake $GITHUB_WORKSPACE -G Ninja -D CMAKE_BUILD_TYPE=$BUILD_TYPE -D FLTK_BACKEND_WAYLAND=OFF -D CMAKE_CXX_STANDARD=11 -D CMAKE_CXX_EXTENSIONS=OFF -D FLTK_BUILD_FORMS=ON -D FLTK_OPTION_CAIRO_WINDOW=ON -D CMAKE_C_FLAGS_INIT="-Wall -Wunused" -D CMAKE_CXX_FLAGS_INIT="-Wall -Wunused"
+      run: |
+        cmake $GITHUB_WORKSPACE \
+        -G Ninja \
+        -D CMAKE_BUILD_TYPE=$BUILD_TYPE \
+        -D FLTK_BACKEND_WAYLAND=OFF \
+        -D CMAKE_CXX_STANDARD=11 \
+        -D CMAKE_CXX_EXTENSIONS=OFF \
+        -D FLTK_BUILD_FORMS=ON \
+        -D FLTK_OPTION_CAIRO_WINDOW=ON \
+        -D CMAKE_C_FLAGS_INIT="-Wall -Wunused" \
+        -D CMAKE_CXX_FLAGS_INIT="-Wall -Wunused"
 
     - name: Build
       working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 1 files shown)


MatthiasWM:
Looking good as far as the code is concerned, and the Ninja build is a plus. 

Question: are arm64 build results helpful for every commit? The macOS builds are already arm64, so we know I there is a processor specific failure. Would it make sense to keep those builds as Actions that can be manually started in the release process, or should they be built on every commit?

Question: in line 106/136, you no longer set `DFLTK_BUILD_SHARED_LIBS:BOOL=ON`, but you do set `-D FLTK_OPTION_CAIRO_WINDOW=ON`. Any particular reason for that change?

ggarra13:
> Question: are arm64 build results helpful for every commit? The macOS builds are already arm64, so we know I there is a processor specific failure. Would it make sense to keep those builds as Actions that can be manually started in the release process, or should they be built on every commit?

They probably should be rebuilt on every commit.  While macOS builds are arm64, the Windows builds use an aarch64 MSVC compiler which might see things differently.  That's how I stumbled on the fact that fluid as it stands today, would not compile the tests properly with the FLTK_HAVE_PEN_SUPPORT was on.
The Linux aarch64 builds run faster than any other CPU so it is a non issue to have them run every time.
For the macOS arm64 builds to match Windows, we would need to compile it with clang-cl.  But most people compiling on Windows use probably MSVC.

ggarra13:
> Question: in line 106/136, you no longer set `DFLTK_BUILD_SHARED_LIBS:BOOL=ON`, but you do set `-D FLTK_OPTION_CAIRO_WINDOW=ON`. Any particular reason for that change?

AFAICT, there was no change from the original workflow.  I have now separated the cmake and apt lines into multiple lines, to make them easier to edit and check parameters.  Feel free to make all builds use the same flags, if you see fit.

Albrecht-S:
@ggarra13 Thanks for this PR. Using `Ninja` on Linux/Unix and macOS seems appropriate and saves some build time.

However, we definitely don't want to use `Ninja` under Windows because this would compile with GNU tools (`MSYS2/MinGW-w64`) rather than Visual Studio (tested in my own fork to be sure).

I also agree with @MatthiasWM that building all (now 6 instead of 4) builds would not be necessary for every commit. I admit that some compilers may issue different diagnostics (or even fail) whereas others work w/o warnings, but we shouldn't waste too much of GitHub's CI build times. It's also annoying to have to wait for more than 4 builds to see if one's recent commit was OK. ;-)

OTOH, using some **additional** build configurations once per day could be helpful. In this case we **could** use `Ninja` (i.e. `MSYS2/MinGW-w64`) to build on Windows (X86 + ARM) and maybe more. I'm also considering to build the docs once per day on GitHub so users and devs can download the docs from GitHub build artifacts if desired.

**_Technical question:_** why did you use backticks ("`") for continuation lines in commit b710212e497df0cc9b059770b4589d27340b8041 instead of backslash ("\\") ? Is there a specific reason, or is this equivalent? I'd prefer using backslash everywhere for consistency (if there's no other reason).

You also wrote:
> While macOS builds are arm64, the Windows builds use an aarch64 MSVC compiler which might see things differently. That's how I stumbled on the fact that fluid as it stands today, would not compile the tests properly with the FLTK_HAVE_PEN_SUPPORT was on.

As I noted above, the Windows builds in your PR use(d) MSYS2/MinGW-w64 (GNU) compilers. I'm aware that some older versions had issues with the (newer) Pen support functions, but we should definitely use MSVC compilers in our "standard" CI builds and - optionally - the GNU compilers/tools in a separate (maybe daily) build config.

ggarra13:
>  It's also annoying to have to wait for more than 4 builds to see if one's recent commit was OK. ;-)

But with Ninja with all platforms you are now waiting LESS not more.  The previous runs where taking in total. about 8 mins for the whole suite.  Now it takes 6 minutes.


> **_Technical question:_** why did you use backticks ("`") for continuation lines in commit [b710212]

Since you are building with PowerShell instead of Msys2, the continuation for it is backticks not backslashes like in Unix.



> As I noted above, the Windows builds in your PR use(d) MSYS2/MinGW-w64 (GNU) compilers.

That may be a bug in my part.  We can keep Ninja and just pass -D CMAKE_CXX_COMPILER and -D CMAKE_C_COMPILER to it.  I'll have a verified fix later today.

ggarra13:
> As I noted above, the Windows builds in your PR use(d) MSYS2/MinGW-w64 (GNU) compilers.

You were indeed correct.  I placed a new action that places cl.exe first in the %PATH%.

This action is well maintained and supports both architectures.  You can then keep using cl.exe with Ninja.

Albrecht-S:
> I placed a new action that places cl.exe first in the %PATH%.
> This action is well maintained and supports both architectures. You can then keep using cl.exe with Ninja.

That's good to know, thanks. But I really don't want to use "`cl` with `Ninja`" only to make it **faster**.
**_Reasoning:_** the main point of our CI builds is to do what "normal Windows users" (no offense intended) would do. That is to use `MSBUILD` (or whatever) and to let the VS IDE files be generated by CMake. I also doubt that `Ninja` and `MSYS2/MinGW-w64` etc. would be installed on standard Windows systems (although they **are** on the GitHub CI platforms).

That said, I'm open for changes, particularly to use Ninja on Linux/Unix and maybe macOS as well. As you say, this will speed up the build w/o sacrificing any tests (using `make` instead of `ninja`doesn't add any advantages, IMHO).

I'd like to optimize our standard builds for speed, but then we could probably add any "other" platforms in a separate CI run (as has been said already, likely once per day).

MatthiasWM:
With macOS, I'd say it's similar to Windows: most users will probably use Xcode as it is pretty much the default on macOS. So should the CI build use Xcode as well? I switched to VS Code and Ninja a while ago on macOS, but I don't know what our user base does.

Albrecht-S:
Regarding macOS and Xcode: running Xcode (or `xcodebuild` or whatever it's called) from the commandline (by using `cmake --build ...` issues way too many useless diagnostic messages. I never tried to read/understand this mess. Therefore, using Xcode in a CI build doesn't seem useful. `make` or `ninja` works well though and outputs only useful messages (compiler warnings).

Albrecht-S:
**_Progress report:_** I'm working on an update by cherry-picking the improvements of this PR. The new CI script will enable us to use any build type on each `push` event or only by manual trigger or schedule like once per day or week. If necessary, I will split the script into two or more parts and let you know how it works ...

Albrecht-S:
@ggarra13 Commit 1e0de4fbddea3c1783fe438dcc465ba1477413d0 implements most of your proposed changes, but some minor pieces may still be missing. This modified workflow allows to edit the jobs by modifying a single "if: ..." line for each job to decide on which triggers the respective jobs are executed.

Thank you, I tried to cherry-pick all your good suggestions.

Right now I left the basic jobs be triggered on each push, whereas additional builds for the arm platform(s) are executed only once per day or when triggered by an FLTK dev. These additional builds can be extended later to build the FLTK docs as well (or we may add another workflow script).

Unfortunately I had to make some decisions that are not included in your PR which makes your PR almost obsolete. I'm sorry for that, but I hope that we're on a good way.

Anyway, I believe the new workflow file can be modified easily to include other build configurations, dependent on any trigger types.

Comments appreciated.

ggarra13:
> Unfortunately I had to make some decisions that are not included in your PR which makes your PR almost obsolete. I'm sorry for that, but I hope that we're on a good way.
> 
> Anyway, I believe the new workflow file can be modified easily to include other build configurations, dependent on any trigger types.
> 
> Comments appreciated.

Any comment on why you decided not to use Ninja on Windows builds?  I think that's a pity.

ggarra13:
> Comments appreciated.

I liked your commented code, but you left out one critical bug I fixed in #1462, in Fl_Widget.cxx.

```
Fl_Widget::~Fl_Widget() {
  // Removed incorrect Fl::Pen::unsubscribe(this); 
  // which would make it crash on aarch64 builds 
  // *AND* AMD64 Debug ones.
  Fl::clear_widget_pointer(this);
  if (flags() & COPIED_LABEL) free((void *)(label_.value));
  if (flags() & COPIED_TOOLTIP) free((void *)(tooltip_));
  image(NULL);
  deimage(NULL);
  // remove from parent group
  if (parent_) parent_->remove(this);
```

Albrecht-S:
> Any comment on why you decided not to use Ninja on Windows builds? I think that's a pity.

Please see above: https://github.com/fltk/fltk/pull/1457#issuecomment-4710342338 .

I'm open to use `Ninja` for Windows with or without `cl` (VS) or `clang` as one of the additional builds in daily or maybe weekly builds, but the "standard" CI builds should IMHO definitely use VS (i.e. MSBUILD) directly for above reasons - even if it takes longer to build. It would also be useful to add MSYS2/MinGW-w64 builds daily/weekly CI builds.

> I liked your commented code, but you left out one critical bug I fixed in https://github.com/fltk/fltk/pull/1462, in Fl_Widget.cxx.

Hmm, this looks like OT **here**, did you mean to post it anywhere else?
I didn't notice any build failures with arm/aarch64 builds as proposed by you, but I'm not sure. (?)

ggarra13:
> > I liked your commented code, but you left out one critical bug I fixed in #1462, in Fl_Widget.cxx.
> 
> Hmm, this looks like OT **here**, did you mean to post it anywhere else?

> I didn't notice any build failures with arm/aarch64 builds as proposed by you, but I'm not sure. (?)

You missed it because my old and original suggestion would build with Msys2 which did not crash.  And now you don't notice as you have it conditional.

With your current build fluid crashes due to the Fl_Widget.cxx bug:

```
$ git switch master
$ git switch -C test_master

edit: .github/workflows/build.yml
         remove conditional compile on ARM64 builds.
         
 $ git push origin test_master
```
 
When you build with MSVC ARM64, you will get fluid crashing like:

```
  Generating valuators.cxx, valuators.h
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(254,5): error MSB8066: Custom build for 'C:\a\fltk\fltk\test\valuators.fl;C:\a\fltk\fltk\test\CMakeLists.txt' exited with code -1073741819. [C:\a\fltk\fltk\build\test\valuators.vcxproj]
```

ggarra13:
Also I noticed the Linux-aarch64 builds are also not set to build on push.  The real question I have for you guys:  do you not get aarch64 runners on Github?  I do get them myself on any push.

PR #1458, opened 2026-06-14
PR: Tablet API driver (actually events as they were named on Cocoa) implementation for Wayland and WinAPI

ggarra13:
Tested to compile and work on Win64, Wayland and Cocoa with my XP Deco v02 tablet.
Tested to compile on Win64 ARM64 and Linux ARM64.  See PR #1457 to see how to add this to .github runs.

Disclosure:  done with AI.  Claude did the main port, Grok suggested a change for Windows behavior. Gemini suggested the change to fix it.  Neither worked and I manually went and fix it in penpal application.  Windows needs a boolean flag to make tablet work, as it synthetizes a dummy FL_PUSH after Fl::Pen::TOUCH is sent.  That FL_PUSH is needed and should not be filtered as you would have to re-subscribe all widgets, which is impossible.  I added a comment to both penpal.cxx and Fl_win32.cxx for how to get tablet working on Windows.


src/drivers/WinAPI/fl_WinAPI_platform_init.cxx
```diff
@@ -11,7 +11,7 @@
 //
 // Please see the following page on how to report bugs and issues:
 //
-//     https://www.fltk.org/bugs.php
+//     https://www.fltk.org/bugs.phps
```
MatthiasWM, line 14: //     https://www.fltk.org/bugs.php

no .phps

MatthiasWM:
Ok, phew. You renamed teh Window driver, so I can't get a good diff on that file, and I have to assume that everything was copied and no additions break things. What I *can* read looks good for a first merge. Bugs can be squashed as they emerge. I am OK with a "Squash and Merge", and if it is only to avoid making this PR even bigger.

If I don;t get a "no" from @Albrecht-S or @ManoloFLTK by Monday 6pm CET, I will give it a shot and hang around to quick-fix issues if they occur.

ggarra13:
@MatthiasWM I removed your WIP Fl_WinAPI_Window_Driver.cxx as it was not working properly.  It was just doing an FL_PUSH with no pressure.  I also did not know what to follow, as you named your Cocoa code as Fl_Cocoa_Pen_Events.mm while most other code is usually named as *_Driver.

MatthiasWM:
> @MatthiasWM I removed your WIP Fl_WinAPI_Window_Driver.cxx as it was not working properly. It was just doing an FL_PUSH with no pressure.

Yeah, thanks for looking into that. I kind of got disrupted half way in.

> I also did not know what to follow, as you named your Cocoa code as Fl_Cocoa_Pen_Events.mm while most other code is usually named as *_Driver.

True. We may want to rename all of them to Fl_XXX_Pen_Driver.. But let's merge first tomorrow. And if everything compiles, we can rename later.

ggarra13:
Here's the problems I found so far with the drivers on all platforms.  With my tablet, none of the OS fully conforms to the Wayland more advanced behavior.

- All platforms: I do not seem to see the barrel button (one of the buttons on the actual pen) respond to Fl::event_state().  I'll do some more testing later today on Wayland to recheck.  The pen API on all platforms seems to miss the WHEEL event (I guess you can have a pen with a mousewheel now?)
- On Windows, it synthesizes a FL_PUSH after a FL::Pen::TOUCH. This makes it easy to work with menus and closing the window without needing to subscribe EVERY widget, but if you have an application that needs FL_PUSH to do something else (like you do on penpal.cxx that draws with the full radius), you need to use a bool variable to stop it.  See penpal.cxx for an example.
- On Wayland, libdecor (used on the titlebar) is unaware of the tablet so you cannot close the window (and you cannot of course subscribe it).  Any ideas or way around that @ManoloFLTK?
- Cocoa seems to accept the barrel button, but on my tablet, it does not set the right BUTTON0/1 state for it, it only changes the state to 1 from 2.

ManoloFLTK:
@MatthiasWM I have no access to any tablet device under Wayland-FLTK. I therefore cannot approve or disapprove this PR.

ManoloFLTK:
> On Wayland, libdecor (used on the titlebar) is unaware of the tablet so you cannot close the window (and you cannot of course subscribe it).

libdecor implements the `wl_touch` interface documented to "represent a touchscreen associated with a seat."
Isn't the pen able to generate a touch event on the window closing button?

Moreover, I'm not sure that FLTK has no access to pen events in the titlebar because it has access to mouse events there. For example function `pointer_enter` of `Fl_Wayland_Screen_Driver.cxx` runs each time the mouse enters the titlebar and FLTK calls function `Fl_Wayland_Window_Driver::surface_to_window(surface)` which returns NULL to recognize the mouse is not yet in the window's client surface. The code of  `pointer_enter` also shows how to make sure the entered surface is the titlebar of a GTK titlebar and FLTK can then use the `GTK shell protocol` and handle so called "gestures" such as middle button clicks.

ggarra13:
> libdecor implements the `wl_touch` interface documented to "represent a touchscreen associated with a seat."

I think you are mistaken Manolo.  Looking at the libdecor code that is bundled with FLTK, it does not support neither `wl_touch`  nor `zwp_tablet_seat_v2` and similar tablet protocols.  Same with libdecor-gtk so far.
Maybe you can point to me where exactly those are handled?

> Isn't the pen able to generate a touch event on the window closing button?

Is there a way in Wayland to synthetize a new event programatically?  With its security conscious approach I find it doubtful.  Can you show me brief code segment, for example, to move the mouse programatically?

ManoloFLTK:
File `libdecor-gtk.c` implements a `wl_touch_listener`. Function `touch_up` can close or maximize a window following touch events. At least, that's what the source code suggests, but I never could see it work for lack of touch device.

ggarra13:
> Moreover, I'm not sure that FLTK has no access to pen events in the titlebar because it has access to mouse events there. For example function `pointer_enter` of `Fl_Wayland_Screen_Driver.cxx` runs each time the mouse enters the titlebar and FLTK calls function `Fl_Wayland_Window_Driver::surface_to_window(surface)` which returns NULL to recognize the mouse is not yet in the window's client surface. The code of `pointer_enter` also shows how to make sure the entered surface is the titlebar of a GTK titlebar and FLTK can then use the `GTK shell protocol` and handle so called "gestures" such as middle button clicks.

Ok.  I looked at Fl_Wayland_Screen_Driver.cxx code and I think I know how to get it to work.  The AI code is encapsulated and does not use Fl_Wayland_Screen_Driver::seat as data passed to the function.  It should still be possible to get it through Fl::screen_driver().
The one thing missing is access to the static wl_surface* gtk_shell_surface, which seems needed to send the events (I am assuming the titlebar is just anothe surface but not in control of FLTK) and the button logic as is done in pointer_button can then be added to tool_cb_down.
Anyway, I'll be doing some refactoring.

ggarra13:
There was a debug printout left on the Win32 driver that got introduced.  It is in:

Fl_win32.cxx

line 1247.  Please do another commit to remove it together with the superflous:
`
#include <iostream> `

at the beginning.

ggarra13:
> Anyway, I'll be doing some refactoring.

@ManoloFLTK I did try and I failed.  The reason the mouse pointer works in libdecor is that it subscribes its wl_surface (hidden to FLTK) to the pointer protocol.  libdecor will need to be changed and add support for tablet-v2.

ManoloFLTK:
File `config.h.in` now ends like this
```
/*
 * FLTK_HAVE_PEN_SUPPORT
 *
 * Do we have pen/tablet support for the current platform?
 * See CMake option FLTK_OPTION_PEN_SUPPORT, but note that a build test
 * (try_compile()) is performed to test if the option is available
 *
 * Note: this option is "hidden" in 'config.h', i.e. it's not (yet)
 * publicly accessibe. Move this to 'fl_config.h' to make it public.
 *
 */
```
that is, there's no line pre-defining `FLTK_HAVE_PEN_SUPPORT`. Is this really wanted?

ggarra13:
> File `config.h.in` now ends like this

That comment should probably be removed.  FLTK_HAVE_PEN_SUPPORT lies in fl_config.h.in as part of the public API.

ManoloFLTK:
> That comment should probably be removed. FLTK_HAVE_PEN_SUPPORT lies in fl_config.h.in as part of the public API.

OK. I've removed this comment and committed that (e0f7ade).

ManoloFLTK:
FLTK's Wayland platform is available from Debian 11 bullseye and Ubuntu 20.04 and above (see `README.Wayland.txt`).
The wayland protocol file used here (`stable/tablet/tablet-v2.xml`) is available since Debian Trixie (or bookworm-backports) and Ubuntu questing 25.10. Therefore I propose to add the small attached patch to file `CMake/options.cmake` that checks for availability of this file for the Wayland platform and turns `FLTK_HAVE_PEN_SUPPORT` off is it's not available on the build platform.

[pen_support_wld.patch.txt](https://github.com/user-attachments/files/29103294/pen_support_wld.patch.txt)

ggarra13:
@ManoloFLTK Your patch sounds reasonable.

ManoloFLTK:
> Therefore I propose to add the small attached patch to file CMake/options.cmake that checks for availability of this file for the Wayland platform and turns FLTK_HAVE_PEN_SUPPORT off is it's not available on the build platform.

Committed at 5f0987d.

ManoloFLTK:
With the present PR, variables `Fl::Private::e_x_down` and `e_y_down` are declared 7 times. I propose the attached patch to declare them once in `src/Fl_Private.H` and allocate and initialize them once in `Fl.cxx`.

These variables are mostly used to support the pen device and as such only when `FLTK_HAVE_PEN_SUPPORT=1`, but not entirely: under macOS they are also used if `FLTK_HAVE_PEN_SUPPORT=0`. For this reason the patch declares those 2 variables independently from the value of `FLTK_HAVE_PEN_SUPPORT`.

[e_xy_down.patch](https://github.com/user-attachments/files/29195532/e_xy_down.patch)

Issue #1427, opened 2026-05-01
X11/Wayland: Fl_Text_Editor's shortcuts and Fl::event_key don't work as expected with certain keyboard layout settings


savannstm:
**Description of the Bug Report**
Standard shortcuts (Ctrl+Z, Ctrl+C, etc.) in `Fl_Text_Editor` stop working when a non-Latin keyboard layout is set as the primary layout in the system.

Snippet from `Fl_Text_Editor`:

```cpp
// These are the default key bindings every widget should start with
static struct {
  int key;
  int state;
  Fl_Text_Editor::Key_Func func;
} default_key_bindings[] = {
    ...
    { 'z',          FL_CTRL,                  Fl_Text_Editor::kf_undo       },
    { 'z',          FL_CTRL|FL_SHIFT,         Fl_Text_Editor::kf_redo       }, // Windows screen driver also defines Ctrl-Y
    { '/',          FL_CTRL,                  Fl_Text_Editor::kf_undo       }, // Emacs
    { '?',          FL_CTRL,                  Fl_Text_Editor::kf_redo       }, // Emacs
    { 'x',          FL_CTRL,                  Fl_Text_Editor::kf_cut        },
    { FL_Delete,    FL_SHIFT,                 Fl_Text_Editor::kf_cut        },
    { 'c',          FL_CTRL,                  Fl_Text_Editor::kf_copy       },
    { FL_Insert,    FL_CTRL,                  Fl_Text_Editor::kf_copy       },
    { 'v',          FL_CTRL,                  Fl_Text_Editor::kf_paste      },
    { FL_Insert,    FL_SHIFT,                 Fl_Text_Editor::kf_paste      },
    { 'a',          FL_CTRL,                  Fl_Text_Editor::kf_select_all },
    { 0,            0,                        0                             }
};
```

At `Fl_Text_Editor::handle_key`, FLTK just checks whether `Fl::event_key` is in the list, using `Fl_Text_Editor::bound_key_function` function:

```cpp
Fl_Text_Editor::Key_Func Fl_Text_Editor::bound_key_function(int key, int state, Key_Binding* list) const {
    Key_Binding* cur;
    for (cur = list; cur; cur = cur->next)
        if (cur->key == key)
        if (cur->state == FL_TEXT_EDITOR_ANY_STATE || cur->state == state)
            break;
    if (!cur) return 0;
    return cur->function;
}
```

The problem is that the key code, returned by `Fl::event_key` depends on set system layouts.

[...]

PR #1461, opened 2026-06-15
Linux: fix support of non-Latin or multiple keyboard layouts (#1427).

ManoloFLTK:
This commit mainly affects X11-FLTK, but also Wayland-FLTK to a lower extent.

Major change regarding X11-FLTK:
Use the XKB X extension. This allows to inform the client at startup of what keyboard layout is active and to ask the X server to send XKB events to its client when the selected layout changes. Altogether, the FLTK app is now always aware of what is the current active keyboard layout. Previously X11-FLTK was wrong as soon as the active layout was not the first in list. This change corresponds to the new preprocessor variable HAVE_X11_XKB being defined to 1. When there's no support for the XKB extension at build time, HAVE_X11_XKB = 0 and FLTK falls back to its code before the present commit.

Other change regarding both X11- and Wayland-FLTK when the active layout is non-Latin and when the pressed key is letter-bound:
The value FLTK puts in Fl::event_key() is changed. Previously, it was the X11-defined non-Latin keysym of the pressed key. This rendered very difficult for an FLTK client to define keyboard shortcuts associated to Ctrl-Letter, for any of the 26 Latin letters. This also prevented FLTK from detecting all standard widget shortcuts (e.g., Ctrl-C or Ctrl-V in the Fl_Input widget) as soon as the selected layout is non-Latin. This also made Linux-FLTK to behave differently from Windows-FLTK and macOS-FLTK. The fix has Linux-FLTK put the Latin letter associated to the key in Fl::event_key() when such key is pressed. This makes Linux-FLTK behave as Windows-FLTK and macOS-FLTK and respond to all FLTK-defined widget shortcuts. It also allows FLTK client apps to easily and in a platform-independent way define Letter-bound shortcuts for menu items or custom widgets that work also with non-Latin layouts.

Unfortunately, it was not found how to obtain this result in a generic way. The fix relies on 2 assumptions found to be true after having surveyed a large number of Latin and non-Latin keyboard layouts. One assumption determines how FLTK decides whether the active layout is non-Latin. The 2nd assumption determines how FLTK determines the pressed key is associated to a Latin letter, and if so to what letter. The source code that depends on these assumptions is limited to a 6-line if() clause. It's therefore easily removed if a more generic method is found.


ManoloFLTK:
TLDR; This PR aims at fixing a number of problems in X11-FLTK and Wayland-FLTK with the data returned by `Fl::event_key()` after a keypress event when 2 or more keyboard layouts are defined. These problems cover, but are not limited to the support of non-Latin keyboard layouts (issue #1427). The proposed fix lets `Fl::event_key()` return the same values under Linux as under Windows and macOS when pressing a letter-bound key of the keyboard.

1) When more than one keyboard layout is defined, FLTK under X11 puts erroneous data in `Fl::event_key()` in many situations listed below. Notice that what FLTK puts in `Fl::event_text()` is correct, so that typing text works OK. What is affected is `event_key()`-dependent stuff, noticeably menu and widget shortcuts.

1a) Consider only Latin layouts. X11-FLTK fills `event_key()` according to the layout located first in the list of usable layouts, even if the selected layout is not the first of the list. Example: suppose the layout list contains, in this order, French, German, English (US). Run `test/handle_keys` and press the top-left letter key of the keyboard (labelled Q in the US layout). X11-FLTK puts 'a' in `event_key()` and 'a', 'q', or 'q' in `event_text()` when the active layout is set to French, German, English, respectively. With English layout selected, run `test/editor`, put some text in the window and type Ctrl-Q. That should trigger the File/Quit editor menu item, but in fact it triggers the "Select all" feature of the `Fl_Text_Editor` widget. Do the same operation with gnome's "Text Editor" app: typing Ctrl-Q does close the app. Thus, `event_text()` is always correct, but `event_key()` is wrong() except when the French layout (first in list) is selected.
For another example, put the German layout first in the list of layouts and run `test/handle_keys` typing the 'Y' key (this key is labelled 'Z' in German keyboards). X11-FLTK puts 'z' in `event_key()` and 'z', 'y', or 'y' in `event_text()` when the active layout is set to German, French, English, respectively. Run `test/editor `while the US layout is selected (German remains 1st in list), type some text and Ctrl-Y, this incorrectly triggers the Undo operation. Run gnome's "Text Editor" app and type Ctrl-Y, no undo operation occurs. It's necessary to type Ctrl-Z (I always name keys with their US name here) to trigger undo, as expected because the US layout is active.

Conclusion: X11-FLTK without the present fix fails to recognize at least some Ctrl-Letter shortcuts if the active keyboard layout is not the first in the list, even with Latin layouts only.

1b) Add also non-Latin keyboard layouts to the list of known layouts. The behavior of X11-FLTK becomes very chaotic as for what it puts in `event_key()` (notice it doesn't prevent `event_text()` from remaining always correct). Consider this order of layouts: English, German, French, Arabic, Russian, Greek, Japanese. Select English (1st in list) and run `test/handle_keys`. Type 'Q' (the top-left letter key), take note of the values of `event_key()` and `event_text()`, select the next layout in list and repeat until the Japanese layout. Here is the outcome:
```
Layout      event_key      event_text   keysym meaning according to X11/keysymdef.h
English     q              q
German      q              q
French      q              a
Arabic      0x5d6          ض			x5d6 = ARABIC LETTER DAD
Russian     0x5d6          й
Greek       0x5d6          ;
Japanese    0x4c0          タ			0x4c0 = KATAKANA LETTER TA
```
Conclusions:
- `event_text` is always correct
- these outcomes don't depend on the order of selections made in the fixed list of layouts: they are unchanged if `handle_keys` is run starting with the Japanese layout and going up the list of layouts or any other order I could try.
- when the 4th layout is active, `event_key` contains this layout's keysym values
- when the 5th or 6th layouts are selected, `event_key` contains keysym values of the 4th layout. I believe so because if I put, say, Russian in position 4, `event_key` values reported for the 5th or 6th layout are Cyrillic letters.
- when the 7th layout is selected, `event_key` contains this layout's keysym values.
- these observations may not allow to predict what X11-FLTK puts in `event_key` for all situations, but for sure, X11-FLTK is often wrong with non-Latin layouts as for `event_key` data.

1c) What happens with non-Latin layouts when one attempts to trigger a menu or widget shortcut with a Ctrl-"Latin letter" keystroke.
When a non-Latin layout is selected, Ctrl-Q should trigger the "File/Quit editor" menu item of `test/editor` because it does with gnome's "Text Editor" and it does with FLTK apps under Windows and macOS (where Cmd replaces Ctrl). Without fix, X11-FLTK does that only if the 1st layout in the list is the US layout. Any other setting prevents at least a few Ctrl-Letter shortcuts from being correctly recognized.

2) The proposed fix - part 1.
The XKB X11 extension allows to fix the incorrect behavior of X11-FLTK when the active layout is not the first in list. With it, the client app is informed of the active layout at startup and it's possible to ask the X11 server to send an XKB event when the active layout changes to inform the client of the nature of the new layout. This is a generic solution documented in the [libxkbcommon library](https://xkbcommon.org/doc/current/group__x11.html#details) that's presented as follows:
> The xkbcommon-x11 module provides a means for creating an xkb_keymap corresponding to the currently active keymap on the X server. …… It can be used as a replacement for Xlib’s keyboard handling. 
  
The fix requires Debian packages `libx11-xcb-dev` and `libxkbcommon-x11-dev` or Fedora package `libxkbcommon-x11-devel` to be installed on the build machine. Without them, FLTK falls back to its pre-fix code.
  
With this fix, FLTK is informed of the active layout at startup and each time the layout is changed.

With this fix and Latin layouts, FLTK behaves under X11 or Wayland as under Windows or macOS. The issue described above in 1a) with Latin layouts is fixed: `event_key` values correctly reflect the active layout and menu shortcuts work as expected.

The issue described above in 1b) with non-Latin layouts is also fixed: FLTK detects each layout change and uses corresponding keysyms. But the issue 1c) above remains, that is, Ctrl-letter keystrokes when a non-Latin layout is active are not recognized in an FLTK app, although they are recognized by gnome's "Text Editor".

3) Use of non-Latin keyboard layouts under Wayland
Without fix, Wayland-FLTK used with non-Latin layouts puts non-Latin data in event_key (and always the correct text in event_text). This prevents the shortcut-handling code of FLTK from identifying all Ctrl-Letter shortcuts. That's a bug that deserves being fixed in FLTK 1.5.

4) The proposed fix - part 2.
Before discussing a fix, it's necessary to define what should we expect FLTK to put in `event_text` when a non-Latin layout is active and Ctrl-"latin letter" is pressed.

4a) What Windows-FLTK does when typing Ctrl-Letter while a non-Latin layout is selected.
The Windows-FLTK source code shows that pressing Ctrl+Letter where Letter is one of the 26 letter keys of the US layout puts in `event_key` what Windows calls the "virtual key". The web site http://www.kbdlayout.info/ reports the behavior of more than 200 keyboard layouts under Windows. This web site shows all virtual keys of a given layout. For all layouts I could explore using this web site (I explored many), "virtual keys" of a non-Latin layout always reproduce the US layout as for the 26 latin letters. In contrast, "virtual keys" of Latin layouts can shuffle somewhat (German and French layouts) or extensively (Dvorak layouts, Turkish F layout) the position of the 26 letters. In all cases, typing Ctrl-X where X is the layout's virtual key in an FLTK app puts the ascii code of X in `event_key` and triggers the Ctrl-X shortcut (X can be here any of the 26 letters). These keypresses trigger the corresponding operations also with non-FLTK NotePad++ app.

4b) What macOS-FLTK does when typing Cmd-Letter while a non-Latin layout is selected.
macOS-FLTK puts in `event_key` the product of the `charactersIgnoringModifiers` Objective-C message which is a Latin letter for 26 keyboard keys, even with non-Latin layouts. The "onscreen keyboard" allows to visualize these letters because, for non-Latin layouts where keys have initially non-Latin values, the onscreen keyboard shows Latin letters when the Cmd key alone is pressed. As with Windows "virtual keys", for all non-Latin layouts I could explore, pressing Cmd makes the 26 Latin letters appear at their locations in the US layout. Conversely, pressing Cmd does not modify the possibly shuffled location of the 26 letters in a Latin layout. Both FLTK apps and Apple's "TextEdit" reply to Cmd-letter shortcuts according to what the onscreen keyboard displays. 

4c) What Linux expects when pressing Ctrl-Letter while a non-Latin layout is selected.
Using gnome's "Text Editor", it's possible to see which Ctrl-Letter keystrokes trigger copy, paste, cut, undo, select all, open, save, find, print, …. In all tested non-Latin cases the Ctrl-USletter keystroke triggers the expected operation. With the "fix - part 1" only, FLTK doesn't recognize such shortcuts because it puts a non-Latin keysym in `event_key` while the shortcut detection code needs a Latin letter there to trigger the shortcut.

To fix this under Linux, one needs to relate a key to a Latin letter just like FLTK does exploiting information transmitted it by the OS under Windows (considering virtual keys) or under macOS (sending the `charactersIgnoringModifiers` Objective-C message, macOS-FLTK puts a Latin letter in `event_key` also with non-Latin layouts and puts non-Latin text only in `event_text`). Unfortunately, nothing in XKB seems to relate a key to a Latin letter when a non-Latin layout is active (see the XKB API at https://xkbcommon.org/doc/current/topics.html). That's reflected in the product of "Show keyboard layout": there's no way to have the displayed layout show a Latin letter associated to a key.

Therefore, the present PR proposes to fix the remaining part of the issue as follows:
- Find a way to determine whether the active layout is Latin or non-Latin.
- Don't change further the behavior of FLTK for Latin layouts. It's OK and matches what FLTK does under Windows and macOS.
- In case of a non-Latin layout, and when the pressed key is one of the 26 letter keys of the US layout, put in event_key the corresponding Latin letter. This makes Linux-FLTK behave as Windows-FLTK and macOS-FLTK.

This fix makes use of 2 assumptions empirically found to be satisfied by all non-Latin layouts surveyed:

A1) Non latin keyboard layouts associate a non-latin keysym to the 'A' key of the 
US layout, whereas Latin layouts associate an ascii letter, possibly not A, to this key.

A2) Non latin layouts require to press Ctrl and what is the 'A' key in the US layout,
to generate a Ctrl-A event, and so on for all 26 latin letters (replace Ctrl by Cmd 
under macOS). That is what FLTK does under Windows and macOS. In contrast, latin layouts 
use the 26 ascii letters but may shuffle a lot their position and may also assign Latin 
letters to what are non-letter keys in the US layout. 

I'm fully aware it's better not to rely on empirical observations and to build a generic solution independent of such observations. The present PR proposes a working solution that could be committed to FLTK 1.5. It would not prevent implementation of a generic solution, if one is ever found, because a single if() block with this pseudocode embodies the empirical fix :
```
  if (the active layout is non-Latin) {
    if (the pressed key is one of the 26 letter keys of the US layout) {
       put this letter's ascii value in Fl::event_key()
    }
  }
```
This 6-line block is present once in `Fl_x.cxx` and once in `Fl_Wayland_Screen_Driver.cxx`.

Expected benefit of the fix
The condition `Fl::event_key() == 'c' && (Fl::event_state() & FL_COMMAND)` can be used across platforms and across keyboard layouts to test for typing Ctrl-C (Cmd-C under macOS). Consequently, FLTK menu and widget shortcut code recognizes Ctrl-C (and so on for all 26 latin alphabet letters) also with non-Latin keyboard layouts and also if the active layout is not the first in list.

Expected impact on existing FLTK apps:
- No impact on any app using any layout as for anything derived from `Fl::event_text()`.
- No impact either on any use of `Fl::event_key()` for non-letter keys (e.g., arrow keys).
- The only impact is when using `Fl::event_key()` with a letter key with a non-Latin layout or when the active layout is not the first in list. It's been checked that FLTK widgets use `Fl::event_key()` only for menu item shortcuts and for widget-defined shortcuts (e.g., support of Ctrl-C/Ctrl-V by Fl_Input). All widget-defined shortcuts include the FL_COMMAND modifier. Therefore the impact of the fix is restricted to how apps defined their menu shortcuts and to specialized apps that would directly use `Fl::event_key()`.
Without the fix, X11-FLTK would correctly recognize menu shortcuts of the form Ctrl-Letter with a non-Latin layout in this setting: 
- 1st layout in list: English (US)
- 2nd layout in list: Russian (or any other non-Latin layout)
- activate the 2nd layout.
Conversely, X11-FLTK would fail to recognize at least some of the Ctrl-Letter menu shortcuts in other settings, for example if the non-Latin layout was 1st in list or if the 1st in list was not US-like. The value X11-FLTK without fix gives to `event_key` when pressing Ctrl-Letter with a non-Latin layout that's first in the list makes detection of this event as a shortcut very impractical because this value is not what FLTK does under other platforms and is not the Unicode value of that key when pressed alone. It's a keysym value as defined by the X protocol and listed in file X11/keysymdef.h (example with the Russian layout, pressing Q puts 0x06ca = CYRILLIC SMALL LETTER SHORT I in `event_key`, distinct from this letter's Unicode point U+0439). One can speculate that users of FLTK with non-Latin layouts would have given up using Ctrl+Letter shortcuts or used the setting where FLTK recognized OK both text and menu shortcuts, rather than crafted custom shortcut-recognizing code that would necessarily have been platform, layout and layout-order specific. Furthermore, this ignores situations with a long list of layouts where FLTK's behavior was quite chaotic.

Miscellaneous
About the shortcuts of the International menu of `test/menubar`:
I didn't find under Linux a keyboard layout that produces as direct keystrokes the Korean characters used as shortcuts in this menu. All other shortcuts work after the fix under X11-FLTK and all but Ctrl+€ under Wayland-FLTK. Before the fix: all non Korean but Ctrl+€ worked.

MatthiasWM:
I finally had the time to read through the PR. I don't understand everything X11 - in particular, I don't know the libraries that you link - but the code looks sound, and I agree with your assumptions. 

I like very much that CMake checks for the files at compile time. AM I right that no additional resources are needed at run time?

I am ok with applying the PR and asking our non-latin keyboard friend to test this out.

Historically, I am so accustomed to the U.S. keyboard that even when I type German, I don't switch layouts :-)

ManoloFLTK:
@MatthiasWM Many thanks for having looked at this PR.

> I like very much that CMake checks for the files at compile time. 

I'm not completely sure of what you're alluding to here. The PR adds 2 new Debian packages (`libx11-xcb-dev` and `libxkbcommon-x11-dev`) required at build-time so the resulting library defines `HAVE_X11_XKB` to 1 and therefore supports changing keyboard layouts. These 2 packages will be added to the list of requirements present in `README.Wayland.txt`. In their absence, CMake will not fail and will fall back to an X11-only build.

> Am I right that no additional resources are needed at run time?

Yes. No additional library is needed at run time for the hybrid Wayland+X11 library. That's visible with the `--ldflags` command as follows
```
./fltk-config --ldflags
-L/home/mgouy/fltk/build/lib -lfltk -lm -lX11 -lXext -lpthread -lXinerama -lXfixes -lXcursor -L/usr/lib/aarch64-linux-gnu -lpangocairo-1.0 -lpango-1.0 -lcairo -lharfbuzz -lgobject-2.0 -lglib-2.0 -lXrender -lwayland-cursor -lwayland-client -lxkbcommon -ldbus-1 -ldecor-0 -lfontconfig -ldl
```
that's unchanged relatively to the pre-PR situation. But, a specially-built libfltk that would be X11 only (no Wayland) would gain dependency to `libxkbcommon`, the key library to handle properly changing keyboard layouts while an app runs.  That dependency was already here for Wayland support. In absence of `libxkbcommon`, FLTK would fallback to the pre-PR code where changing layouts are not handled properly. 

> …and asking our non-latin keyboard friend to test this out.

The OP of #1427 has already returned his advice on that issue's page: "No complaints about this PR. Brilliant reasoning and observations. For me personally, everything looks more than good."

PR #1464, opened 2026-06-20
New `Fl::set_font_by_family` font loading system

savannstm:
Text of this PR was written by hand, then fed into Claude and asked to make it a little bit more concise, eliminate repeated points and make clearer.

## Changes

- New `Fl::set_font_by_family` and `Fl::get_font_name2` public functions.
- Driver internals now take explicit `weight` and `style` arguments instead of inferring them from the font name string.
- `Fl::set_font` now pre-parses weight and slant from the passed font name.
- Fixed a buffer overflow in `Fl_Cairo_Graphics_Driver::get_font_name` (`strcpy` changed to `strlcpy`).
- `get_font_name` no longer appends `bold`/`italic` to `f->fontname`. See rationale below.
- Code that blindly used `f->name + 1` (stripping the first character on platforms where FLTK font names carry an `I`/`B`/`P`/space prefix) now actually checks whether the first character is one of those before stripping. This may not cover all such call sites - flagging in case there are others.

## Found bug

`Fl_Xlib_Graphics_Driver::get_font_sizes` calls `XftListFonts` and unconditionally skips the first character of the font name, regardless of whether it's actually a prefix character. Compare with `fontopen`, which only strips `I`/`B`/`P`/space and leaves everything else alone. `get_font_sizes` looks wrong here, and this wrong behavior is present in the GDI driver, for example.

I fixed this bug by checking the first character against `B`/`I`/`P`/` ` first, but I may be wrong about it being a bug. Maybe it's somehow how it supposed to be.

## Testing

- **Linux (Pango / Xft / plain X11):** all three build and produce consistent `Fl::set_font` behavior.
- **macOS:** untested - no Mac, no Obj-C experience. `ct_weight_from_opentype` and `create_font` (replacing the `CTFontCreateWithName` calls) are AI-generated and need review from someone who knows CoreText.
- **Windows:** tested with mingw-w64 + wine. Builds and produces consistent `Fl::set_font` behavior.

Testing/troubleshooting help welcome, especially for macOS and Windows.

## Design rationale

**Naming:** `set_font_by_family` over `set_font_by_name` - the OS resolves a font using family + weight + slant, not a bare name, so `_by_family` is the accurate description.

**Not appending bold/italic in `get_font_name`:** the attributes are already returned via the optional integer out-param, so synthesizing them into the name string is redundant and, as noted in the existing source comments, unreliable - the font name may already contain "bold"/"italic" as part of its actual name. It's up to the user to indicate whether a certain font bold or italic, or anything else.

**`Fl_Fontdesc` shouldn't store the user-passed name:** currently it does, and `get_font_name` dynamically re-strips the prefix character from it on every call. I don't see what the stored raw name is for - the caller already knows what they passed in. What `Fl_Fontdesc` actually needs is the family (+ weight/slant) that the OS *resolved*, not the request string. Practically: the prefix-stripping and weight/slant parsing should happen once, in `Fl::set_font`/`Fl::set_font_by_family`, and the result stored directly as `int weight`/`int slant` fields on `Fl_Fontdesc` - not re-derived from a stashed string on every `get_font_name` call.

This ties into font-loading success propagation: right now fonts aren't actually loaded until used, so there's no trivial way to report load success/failure at `set_font` time. If `set_font`/`set_font_by_family` loaded the font immediately via the OS API, failure could be propagated to the caller, and `get_font_name` could return the actually-resolved family - which it currently can't do, since both `Fl::font_name` and `Fl::get_font_name` just echo back what the user passed in. I haven't attempted this in the current PR, this is possibly breaking.

`Fl::get_font` has the same issue conceptually - its docs imply the string differs per face, but two distinct `Fl_Font` indices (e.g. `FL_FREE_FONT` and `FL_FREE_FONT + 1`) can carry identical names, so the int index is the only reliable identity. `get_font_name`'s contract makes no consistency guarantee either way, so changing what it returns shouldn't break anything.

## Documentation gap

`Fl::set_font`'s docs don't cover the XLFD and comma handling when Xft is used: if XLFD name is passed and Xft is active, XLFD name actually gets parsed right. If commas are in the parsed font name and Xft is used, the name will be split by comma into multiple font names. This behavior is not true on any other backend, including X11+Pango.

## Future work / out of scope for this PR

- **Drop legacy XLFD handling and `Fl::set_fonts`** in the next major release. XLFD loading only fires when Xft/Pango are both disabled, or when Xft is enabled and an XLFD name is explicitly passed - fontconfig has been present on Linux for 20+ years, and `Fl::set_fonts`'s own docs already say it's not relevant anymore.
- Once immediate font loading lands, `Fl_Fontdesc.name` could be dropped entirely in favor of just the resolved family buffer.
- `Fl_Fontdesc.fontname` needs to be either changed to be bigger than 128 bytes, because in some cases it may not fit the font name if it's bigger, or the best way is to make this buffer dynamically-allocated, fitting the font name exactly.

## TL;DR

Adds `Fl::set_font_by_family`, moves weight/slant parsing into `Fl::set_font` instead of leaving it to be re-derived later, and fixes a buffer overflow + a prefix-stripping bug along the way. Bigger structural issue (fonts not loaded until use, so success can't be propagated) is flagged but not addressed here. It needs further discussion.

### DIFF
--- a/FL/Enumerations.H
+++ b/FL/Enumerations.H
@@ -989,6 +989,27 @@ const Fl_Font FL_BOLD                   = 1;    ///< add this to Helvetica, Cour
 const Fl_Font FL_ITALIC                 = 2;    ///< add this to Helvetica, Courier, or Times
 const Fl_Font FL_BOLD_ITALIC            = 3;    ///< add this to Helvetica, Courier, or Times
 
+enum Fl_Font_Weight {
+    FL_WEIGHT_THIN = 100,
+    FL_WEIGHT_EXTRALIGHT = 200,
+    FL_WEIGHT_LIGHT = 300,
+    FL_WEIGHT_SEMILIGHT = 350,
+    FL_WEIGHT_BOOK = 380,
+    FL_WEIGHT_NORMAL = 400,
+    FL_WEIGHT_MEDIUM = 500,
+    FL_WEIGHT_SEMIBOLD = 600,
+    FL_WEIGHT_BOLD = 700,
+    FL_WEIGHT_EXTRABOLD = 800,
+    FL_WEIGHT_HEAVY = 900,
+    FL_WEIGHT_ULTRAHEAVY = 1000,
+};
+
+enum Fl_Font_Style {
+    FL_STYLE_NORMAL = 0,
+    FL_STYLE_ITALIC = 1,
+    FL_STYLE_OBLIQUE = 2,
+};
+
 /**@}*/
 
 /** Size of a font in pixels.
--- a/FL/Fl.H
+++ b/FL/Fl.H
@@ -619,6 +619,17 @@ FL_EXPORT extern const char* get_font(Fl_Font);
 FL_EXPORT extern const char* get_font_name(Fl_Font, int* attributes = 0);
 
 /**
+  Get a human-readable string describing the family of this face.  This
+  is useful if you are presenting a choice to the user.  There is no
+  guarantee that each face has a different name.  The return value points
+  to a static buffer that is overwritten each call.
+
+  The integer pointed to by \p weight (if not null) is set to a OpenType font weight from 100 to 900.
+  The integer pointed to by \p style (if not null) is set to 0, if no style, 1 if italic, and 2 if oblique.
+*/
+FL_EXPORT extern const char* get_font_name2(Fl_Font, int* weight = 0, int* style = 0);
+
+/**
   Return an array of sizes in \p sizep.  The return value is the
   length of this array.  The sizes are sorted from smallest to largest
   and indicate what sizes can be given to fl_font() that will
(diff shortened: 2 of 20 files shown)


MatthiasWM:
This is quite an awesome PR. Thank you very much. I like that the API is unified for all platforms, and more precise. I can't test everything right here right now, but everything looks very thought out.

I have one tiny suggestion: can we get an alternative to `get_font_name2()`? We have tried really hard to avoid trailing numerals, because the user never knows if this refers to an FLTK version number, or if there is an `get_font_name3()`. Maybe we can come up with something more descriptive, I don't know, `get_descriptive_font_name2()` or `get_undecorated_font_name2()` or something more descriptive than `2`.

Other than that, I am looking forward to when you mark your PR "Ready for review"

savannstm:
@MatthiasWM I changed `get_font_name2` to `get_font_family`, so it corresponds to `set_font_by_family`.

I guess this is ready for review, however someone really has to test the behavior on MacOS. And there's some points in the PR body that need to be considered for the future development, like font loading success propagation.

Otherwise, I think PR is ready for review.

PR #1472, opened 2026-06-22
Tablet support for titlebar buttons for both Cairo and GDK3 plugins

ggarra13:
This PR implements, roughly based on Matt's and my own work, Wayland support for the buttons of both Cairo and GDK3 backends.

As libdecor does not provide a method for knowing where the buttons are, this PR implements a function that accepts offsets from the right to control the button placement.


ManoloFLTK:
- On the very positive side: this PR does not modify files `libdecor-gtk.c` nor `libdecor-cairo.c `which means it should be compatible with the distribution package form of `libdecor`.
- On the negative side: I believe `libdecor` will in the future delegate all graphics to a separate process which will hinder access to pen events inside the titlebar without cooperation from `libdecor` itself.

ggarra13:
> * On the very positive side: this PR does not modify files `libdecor-gtk.c` nor `libdecor-cairo.c `which means it should be compatible with the distribution package form of `libdecor`.

Critizing my own work.  Albeit it does not modify libdecor, it does add a dependency on libdecor's memory layout for its titlebar surface in functions for:

 fl_is_surface_from_GTK_titlebar (which you Manolo wrote)
 fl_is_surface_from_cairo_titlebar (which I wrote).

ggarra13:
> * On the negative side: I believe `libdecor` will in the future delegate all graphics to a separate process which will hinder access to pen events inside the titlebar without cooperation from `libdecor` itself.

If libdecor moves to its own process, it would not be hindered by the limitation on tablet-v2's only tablet manager per process, so it would be able to implement the tablet-v2 protocol.

MatthiasWM:
Sorry for my slow reply. So you basically rewrote the libdecor decoding of mouse event to handle pen events? 

I am no fan of code duplication, but if this is the way it must be under Wayland, I'll accept it.

@ggarra13 @ManoloFLTK : I can't really judge the code. I know what's going on, but I don't know the details. Do you both agree that the code shall be merged as is? If so, I'll happily do that and fix the patch to match my renaming of files ;-)

ggarra13:
@MatthiasWM Yes, so far it is the only way I found.  The position of the buttons on the title bar are "guessed".   There's one thing it does not do: highlight the titlebar buttons.  Other than that, the code was heavily battle tested.  libdecor is not likely to change due to Wayland's security, albeit as @ManoloFLTK libdecor may become a new process.  If it does, it will be up to libdecor itself to handle the tablet, as it does with touch screens.  I believe Manolo may be in touch with libdecor developer(s) and may provide more feedback.

ManoloFLTK:
> I am no fan of code duplication, but if this is the way it must be under Wayland, I'll accept it.

I don't think this is the way it must be done under Wayland.

One reason is that the position of titlebar widgets is guessed. Gnome allows to customize the content of window titlebars with the "tweaks" application. Titlebar widgets can be placed to the left instead of the right; the minimize and the maximize widgets can be removed, independently of the window's capabilities that influence what is found in the titlebar menu. This would invalidate all guessing made in the PR.
Another reason is that libdecor plans to relocate all GTK operations to another process.
The way to go is the same as what was done to support touch events: process them in the libdecor-gtk plugin. Touch events support was initially a PR to libdecor, that was later accepted and committed to master. The same approach should be followed as for support of tablet events. I can't do it having no access to adequate hardware.

> Do you both agree that the code shall be merged as is?

-1 to merge this PR.

ggarra13:
@ManoloFLTK @MatthiasWM Thanks for your feedback.  Based on your feedback, I agree that GTK should not be handled like this then.  What about the Cairo backend?  Same thing?
 
Let me propose two things then:

1) Keep this patch on my Vulkan branch for the time being.
2) Modify this PR that has just the fixed subwindow code, removing the GTK (and Cairo?) button guessing, as that's a current bug in the Wayland driver otherwise.

ManoloFLTK:
> the fixed subwindow code

What's the remaining purpose of this PR?

ggarra13:
Fix a bug: Currently, subwindows, like those created with allow_expand_outside_parent() are handled in any order.  With this patch, the front-most window is used for tablet events.

Albrecht-S:
I agree that guessing is no solution, and thanks to Manolo for adding even more info.

@ggarra13 Could you please open a separate PR for your intended bug fix (only) so we can close this PR as-is?

ggarra13:
> @ggarra13 Could you please open a separate PR for your intended bug fix (only) so we can close this PR as-is?

@Albrecht-S @ManoloFLTK @MatthiasWM You can find a clean PR #1562 for the improved find_below_pen subwindow/groups function and a fix for tooltip windows.

Albrecht-S:
@ggarra13 Thanks for opening #1562 which seems to be the PR requested above. Please correct me if I'm wrong.

Can we close [this one](https://github.com/fltk/fltk/pull/1472) now?

ggarra13:
@Albrecht-S ye

ggarra13:
@Albrecht-S That's correct.   Cllose this one.

PR #1477, opened 2026-06-29
Wayland Driver: keep track of decimal portion of past scroll events

InventorXtreme:
I noticed that under certain Wayland compositors, touchpad scroll events are sent such that slow scrolling results in only fractional (i.e. 0.5) values being passed into fltk. The current Wayland driver simply chops off the decimal portion, meaning that slow scrolling with a touchpad is impossible. Naturally, this is very annoying to use and prevents precise scrolling in large text displays or list groups. 

To remedy this, I have the function pointer_axis keep a sum of the decimal portion of past scrolling values, and whenever the sum crosses beyond 1 in the positive or negative direction, it will add 1 unit in the proper direction to the delta. In this patch this is implemented with a static variable in the function, which I do not think is very clean code, but I wanted to ask the more experienced developers here where I should add that state to. A couple of minutes of browsing through the code leads me to think putting it in the wayland seat struct would make sense, but I do not have the familiarity with the codebase to be sure. 

Apologies if I have missed some important convention to contributing, I am not a big open source contributor but want to make sure other people can take advantage of my fix. The patch has been tested on Gentoo Linux running Hyprland.

### DIFF
--- a/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
@@ -409,13 +409,13 @@ static void pointer_axis(void *data, struct wl_pointer *wl_pointer,
   int delta = (int) delta_whole;
   
   // add the fractional part to carryover
-  carryover = carryover + delta_fractional;
+  carryover +=  delta_fractional;
 
   // if carryover has reached |1|, add it to the total delta
-  delta += (int) (carryover / 1.0);
+  delta += (int) carryover;
 
   // keep any remaining fractional portion
-  carryover = std::fmod(carryover, (double) 1);
+  carryover -= (int) carryover;
 
   if (abs(delta) >= 10) delta /= 10;
   // fprintf(stderr, "FL_MOUSEWHEEL: %c delta=%d\n", axis==WL_POINTER_AXIS_HORIZONTAL_SCROLL?'H':'V', delta);
--- a/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
@@ -393,7 +393,26 @@ static void pointer_axis(void *data, struct wl_pointer *wl_pointer,
   Fl_Window *win = Fl_Wayland_Window_Driver::surface_to_window(seat->pointer_focus);
   if (!win) return;
   wld_event_time = time;
+
+
+  // This variable will persist across function calls,
+  // keeping track of any scroll inputs that havent quite crossed the
+  // threshold to scroll by one unit
+  static int carryover = 0;
+
+  // the integer part of the value
   int delta = wl_fixed_to_int(value);
+
+
+  // add the decimal part to carryover
+  carryover +=  (!(value & 0x80000000)) ? (value & 0xFF) : -(value & 0xFF);
+
+  // if carryover has reached |1|, add it to the total delta
+  delta += carryover / 256;
+
+  // keep any remaining decimal region
+  carryover = carryover % 256;
+
   if (abs(delta) >= 10) delta /= 10;
   // fprintf(stderr, "FL_MOUSEWHEEL: %c delta=%d\n", axis==WL_POINTER_AXIS_HORIZONTAL_SCROLL?'H':'V', delta);
   // allow both horizontal and vertical movements to be processed by the widget
--- a/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
@@ -398,20 +398,24 @@ static void pointer_axis(void *data, struct wl_pointer *wl_pointer,
   // This variable will persist across function calls,
   // keeping track of any scroll inputs that havent quite crossed the
   // threshold to scroll by one unit
-  static int carryover = 0;
+  static double carryover = 0;
 
-  // the integer part of the value
-  int delta = wl_fixed_to_int(value);
+  double delta_full = wl_fixed_to_double(value);
 
+  double delta_whole = 0;
+  double delta_fractional = std::modf(delta_full, &delta_whole);
 
-  // add the decimal part to carryover
-  carryover +=  (!(value & 0x80000000)) ? (value & 0xFF) : -(value & 0xFF);
+  // the whole part of the value
+  int delta = (int) delta_whole;
+  
+  // add the fractional part to carryover
+  carryover = carryover + delta_fractional;
 
   // if carryover has reached |1|, add it to the total delta
-  delta += carryover / 256;
+  delta += (int) (carryover / 1.0);
 
-  // keep any remaining decimal region
-  carryover = carryover % 256;
+  // keep any remaining fractional portion
+  carryover = std::fmod(carryover, (double) 1);
 
   if (abs(delta) >= 10) delta /= 10;
   // fprintf(stderr, "FL_MOUSEWHEEL: %c delta=%d\n", axis==WL_POINTER_AXIS_HORIZONTAL_SCROLL?'H':'V', delta);


Albrecht-S:
Wow, thanks for this PR. I understand what you want to achieve but I'm not our Wayland specialist which is @ManoloFLTK. I remember that Windows has a similar mechanism to save partial scrolling deltas, hence this looks plausible although I don't know the Wayland protocol enough to say it's correct.

To your patch proposal: I don't really grok your "arithmetic", but I seem to understand that you're "calculating" with the Wayland specific type `wl_fixed_t`. I'd wish to see a more "well-defined" function like `wl_fixed_to_int()` to retrieve the fractional part (or is there one that converts `wl_fixed_t` to float or double? That would be more appropriate than bit masking (although `wl_fixed_t` should be well defined by itself).

InventorXtreme:
Thank you for the response. There is a function that can convert it into a double, but I (for some reason) thought that the decision to use fixed point decimals was an FLTK design decision, not a Wayland decision. Because Wayland does not provide a function for extracting only the fractional part, I improvised. Since I now realize that I am cool to use floating point numbers, I will refactor and use that method instead. I will also look at the windows implementation of touchpad scrolling for inspiration.

InventorXtreme:
I changed the bit twiddling into double manipulation, which should make it more clear. The Windows implementation also uses a static variable to keep track of past partial scrolling deltas, so I figure that it is actually fine to use here as well.

src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
```diff
@@ -393,7 +393,30 @@ static void pointer_axis(void *data, struct wl_pointer *wl_pointer,
...
+
+  // the whole part of the value
+  int delta = (int) delta_whole;
+  
+  // add the fractional part to carryover
+  carryover = carryover + delta_fractional;
+
+  // if carryover has reached |1|, add it to the total delta
+  delta += (int) (carryover / 1.0);
+
```
Albrecht-S, line 416: Why not just
```
delta += (int)carryover;
```

... or do we need something like floor() or ... ? All these nasty details ;-(

src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx
```diff
@@ -393,7 +393,30 @@ static void pointer_axis(void *data, struct wl_pointer *wl_pointer,
...
+  // keep any remaining fractional portion
+  carryover = std::fmod(carryover, (double) 1);
+
```
Albrecht-S, line 419: I think it would be clearer to replace this with
```C++
carryover -= (int)carryover;
```
i.e. to subtract **the same value** that was added to `delta`.

InventorXtreme:
I cleaned up both of those operations. I had just done a literal translation of the previous commit to work with floats and hadn't noticed that both the division and mod were pretty much nops.

Albrecht-S:
OK, latest version looks pretty good, but we need to test and evaluate. Maybe we want to switch back to integer arithmetic for speed considerations, but then we know what to do now. Thanks anyway.

Waiting for @ManoloFLTK to chime in and to comment on this...

Albrecht-S:
FTR: I tested the original (master) and the patched version, and I confirm that it's indeed **impossible** to scroll **very slowly** using the touchpad under Wayland (Debian 12 / Bookworm). If you move your fingers slow enough, **_absolutely nothing happens_** (as expected). With the patch, the same slow scrolling operation works fine, i.e. it scrolls very slowly as it should do.

Testing with test/scroll, using either current master of this PR's commit 212e3a2601ba77b64bf63359918ebed245f1d92d.

MatthiasWM:
This pull request is a great idea. Thank you very much.

Why do we have a line `if (abs(delta) > 10) delay /= 10;` ?? Not saying it is wrong, but I don't understand.

Also, you use one static value, but scrolling has at least two axes: horizontal and vertical, and they used to be calculated independently: `if (axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL) ...`

Lastly (sorry, I am too pedantic), I am not sure if `int delta = (int) delta_whole;` is correct because I believe it's always rounding *down*, but for negative offsets, we should round towards 0, right? Not sure, no time to test right now.

InventorXtreme:
The `if (abs(delta) > 10) delay /= 10;` was added by a previous commit, I think because some Wayland compositors will send much higher values than other for some reason, but I am not entirely sure. I think the `int delta = (int) delta_whole` is correct because iirc double to int conversion simply truncates the fractional portion (i.e. -3.75 -> -3). I will try to add another axis carryover later. Finally, if we end up using the integer code instead of the floating point code, I think there may be a bug in the negative case as it simply grabs the last 8 bits of the int and I am do not think it accounts for the fact the number is stored in two's complement. I will look into this more if we end up using it.

Albrecht-S:
@MatthiasWM Good points.

My tests showed that scrolling down and scrolling up behaved as expected which seems to indicate that casting to `(int)` does the right thing (truncating towards 0). I always forget such details. Otherwise we might need floor() ?

> Why do we have a line if (abs(delta) > 10) delay /= 10; ?? Not saying it is wrong, but I don't understand.

Neither do I, but that was not part of the patch, it existed before that in master. @ManoloFLTK ?

Having two axes is always a valid observation. We should consider this as well.

I'm also not quite happy with the `static` variable as such, even if the Windows driver does similar things. Maybe it doesn't matter, but my gut feeling says that such a variable would have to be reset to 0 at some time. But when? OTOH, if it only affects the very first part of scrolling in one direction, this may be negligible. Maybe.

InventorXtreme:
I agree that the static variable does leave something to be desired. I do think that it would be acceptable to not reset the value to 0, as in a worst case scenario when, for example, the user wants to scroll down but the carryover value is -0.999..., at worst the user needs to scroll ever so slightly more to cause it to move in the correct direction. I do not think this is too big of a deal. However, I do think that we could implement a reset functionality by having it reset to 0 whenever non-scrolling pointer motion is detected, which I think would be a one line change to pointer_motion, assuming we store the variable in the Wayland seat structure. I my gut feeling is that would be a pretty good approximation for when the user has moved on to another element or has stopped scrolling. Once again, even if this heuristic fails occasionally, the failure case is simply that the user has to scroll almost imperceptibly more or less than normal, so I do not think it is a major issue one way or the other.

ManoloFLTK:
> Why do we have a line if (abs(delta) > 10) delay /= 10; ?

That's because when I run wayland, I see function `pointer_axis()` receiving parameter `wl_fixed_t value` such that `wl_fixed_to_double(value)` is either +10.0 or -10.0. I didn't try to understand why this is so.

Albrecht-S:
A short excerpt of a much longer AI reply:

_Historically, before the introduction of the axis_discrete event in Wayland protocol version 5, compositors often used a hardcoded value (commonly 10) to represent a single logical scroll step (a "detent" or click of a mouse wheel). In this legacy context, a value of 10 meant "scroll one line," not "move 10 pixels"._

ManoloFLTK:
> Historically, before the introduction of the axis_discrete event in Wayland protocol version 5, ……

I see that FLTK uses now the pointer protocol at version 3. But it should be possible to ask for a more advanced version, provided FLTK implements its additional requirements.

ManoloFLTK:
I've tried to ask for a more advanced version of the `seat` protocol. This doesn't change the fact that function `pointer_axis()` receives parameter `wl_fixed_t value` such that `wl_fixed_to_double(value)` is either +10.0 or -10.0.  My conclusion is therefore that this division by 10 statement can remain as it is.

Albrecht-S:
FTR: There is a new, superior PR (#1581, created by @MatthiasWM) with a full fix for all supported platforms. It
- keeps track of the fractional part of past scroll events (as this one does for Wayland only)
- adds two `float` event variables to give full access to "fractional scrolling" in x/y direction.

See also #1571.

I'm leaving this PR open for now. Once we decide to merge #1581, this PR will become obsolete.

@InventorXtreme Thanks for your work on this anyway, it's very much appreciated.

Albrecht-S:
Closing this PR because PR #1581 has been merged.

@InventorXtreme Thanks for your efforts, I hope you like this solution.

PR #1480, opened 2026-07-12
Fl_Native_File_Chooser_Zenity: reduce count call strlen() for micro-op

GermanAizek:
Each time a part of the command was added, strlen(command) function was called to find the current end of the string This led to unnecessary line passes (calculating the length for O(n)) at each iteration of the loop Commit transfers management to the cumulative index tracking method (variable l)


ManoloFLTK:
@GermanAizek Please, explain the meaning of "micro-op".

GermanAizek:
> @GermanAizek Please, explain the meaning of "micro-op".

Hi, micro optimization

MatthiasWM:
As the comment in this function already states, this should be solved using std::string. The PR does not improve the code and just moves strlen() calls around.

PR #1481, opened 2026-07-12
Fl_Menu: replace SAFE_STRCAT and reduce strlen() calls

GermanAizek:
Changes replaces an unsafe SAFE_STRCAT macro with a more reliable and correct string length management mechanism to prevent buffer overflows and reduce strlen() calls

### DIFF
--- a/src/Fl_Menu_.cxx
+++ b/src/Fl_Menu_.cxx
@@ -27,8 +27,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


Albrecht-S:
Thanks for your PR. Please fix indenting to 2 spaces per level, otherwise your patch(es) can't be accepted.

Your patch includes this comment:
> // BUG FIX: Pass original namelen, not current len, to recursive call

Please elaborate what the bug is (was) and what its effect was.

Albrecht-S:
1. No reply to a request for more than 2 weeks
2. This PR won't be applied. The solution to the issue (if any) will be different.

PR #1482, opened 2026-07-12
settings_panel: remove excess calls c_str()

GermanAizek:
meaningless cast string -> const char* -> string


MatthiasWM:
Fixed in settings_panel.fl

PR #1484, opened 2026-07-12
fltk: use default constructors and destructors for micro-op

GermanAizek:
References:
- https://stackoverflow.com/questions/59261490/why-is-there-performance-variation-using-default-constructor-instead-of
- https://stackoverflow.com/questions/56272431/how-can-a-compiler-generated-default-constructor-be-more-efficient-than-a-self-w


MatthiasWM:
Reminds me that eventually we will have to write out rule of five constructors, move and copy. But that's for another day,

PR #1486, opened 2026-07-12
Fl: realloc always need check on valid ptr maybe out of memory

GermanAizek:
References:
- https://stackoverflow.com/questions/21006707/proper-usage-of-realloc
- https://stackoverflow.com/questions/1986538/how-to-handle-realloc-when-it-fails-due-to-memory


MatthiasWM:
This is due to use C++11 containers. For now this code is hiding one error condition (an unallocated object) with another (an unwatched widget). The correct way would probably be to throw an out of memory exception (which the C++ version will do).

PR #1494, opened 2026-07-12
print_panel: prevent buffer overflow limit sscanf 1024 bytes

GermanAizek:
Reference: https://stackoverflow.com/questions/71835933/avoiding-char-buffer-overflow-more-efficiently


MatthiasWM:
Compacted similar PRs into commit e6ae02ebaf19c12a8a4184734e7c006a89d9fe35

PR #1496, opened 2026-07-12
Fix truncation to 32bit (memsize -> 32bit -> memsize)

GermanAizek:


MatthiasWM:
AI slop. Both issues are technically there (if you assume that a file name can be longer than 4'294'967'295 bytes), but even then, the suggested fix is wrong because it does not reflect the intention.

GermanAizek:
@MatthiasWM,
this is already static analysis, I use LLM only if I need to write a lot of code.

MatthiasWM:
Compacted similar PRs into commit e6ae02ebaf19c12a8a4184734e7c006a89d9fe35

PR #1500, opened 2026-07-12
Fl_File_Input: make errorcolor() as static

GermanAizek:


MatthiasWM:
Methods were removed. No longer applicable.

PR #1502, opened 2026-07-12
Fl_GIF_Image: GIF_FRAME make data pointer as const read-only

GermanAizek:
Using const where the data should not change improve code readablity, make use safe API ABI and allows C/C++ compiler to minor optimize code better.

Reference:
- https://stackoverflow.com/questions/68526590/const-pointer-vs-pointer-c
- https://www.reddit.com/r/C_Programming/comments/ghoku0/what_are_the_benefits_to_use_const_pointers_or/

### DIFF
--- a/FL/Fl_GIF_Image.H
+++ b/FL/Fl_GIF_Image.H
@@ -66,8 +66,8 @@ protected:
     const struct CPAL {
       uchar r, g, b;
     } *cpal;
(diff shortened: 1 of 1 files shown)


MatthiasWM:
Bundles a batch of small, independent const-correctness fixes into one commit rather than ~50 near-identical one-line diffs: 19366c3f720d00d9af9b9db616df5bd6179c1633

PR #1507, opened 2026-07-12
Fl_Graphics_Driver: make members as const for micro-op

GermanAizek:
Reference: https://stackoverflow.com/questions/5827799/is-it-good-practice-to-add-const-at-end-of-member-functions-where-appropriate


MatthiasWM:
Bundles a batch of small, independent const-correctness fixes into one commit rather than ~50 near-identical one-line diffs: 19366c3f720d00d9af9b9db616df5bd6179c1633

Issue #1407, opened 2026-04-13
Add -scale as a command line argument


MatthiasWM:
It would be nice if users can launch apps at a scale, so adding "-scale" as a command line argument to `int Fl::arg(int argc, char **argv, int &i)` makes sense to me. See `-geometry WxH+X+Y` and other command line arguments.

Five possible extra points:

Not all apps in test check the command line args though, but it would be nice if `test/demo` was scaled by the user, to set the environment variable "FLTK_SCALING_FACTOR" locally, so processes launched by `demo` start with the same scale.

erco77:
I haven't seen a commit yet, so curious: does this `-scale` flag set/overwrite the `FLTK_SCALING_FACTOR` environment variable, or does it set the scale via the API?

And which ever way that works, it should _probably_ (almost surely) override the env variable, and should be documented as such.

I say it should override the env variable because env variables are typically used for defaults, and command line flags take precedence to override them within the local app only. A command line flag would *not* affect other FLTK apps that this app might start, as those would simply inherit the environment settings and use those as the default.

ManoloFLTK:
I'd like to point out a possible ambiguity in what is the scale of an FLTK app. Windows and Wayland use a scale notion (independently from FLTK of course) when they allow to scale each display to account for the varied resolution of today's displays. FLTK apps use that scale when they start for the specific display where a window maps. Variable FLTK_SCALING_FACTOR, if defined, is used multiplicatively to this initial scaling. Finally, the user can type `ctrl-+` and multiply once more the scale by 110%, 150%,…
I'm afraid a user of the new `-scale` command line argument could confuse it with what Windows and Wayland call a display's scale factor.

Albrecht-S:
I **think** a good algorithm would be like this:

1. Always read the scaling factor from the system. This is what the user would see if they ran the program directly.
2. Read env variable `FLTK_SCALING_FACTOR` ( -> `s1`, default = 1.0) **and** the value used by `-scale` ( -> `s2`, default = 1.0) on the commandline. Multiply these two values and use the result ( `s3 = s1 * s2` ) as the factor to apply to the system scaling factor. The result would be that a user would get a predictable behavior when they use `-scale` because it's applied on top of the "other" factors that would be used if `-scale` was not applied.
3. Set the value `s3` as the new env variable `FLTK_SCALING_FACTOR` (in demo.cxx) before other programs are executed.
[...]

PR #1522, opened 2026-07-13
Add -scaling_factor as cmdline arg and env FLTK_SCALING_FACTOR

GermanAizek:
Reference: https://github.com/fltk/fltk/issues/1407

### DIFF
--- a/src/Fl_arg.cxx
+++ b/src/Fl_arg.cxx
@@ -24,6 +24,8 @@
 #include "Fl_System_Driver.H"
 #include "Fl_Screen_Driver.H"
 #include <ctype.h>
+#include <stdlib.h>
+#include <FL/fl_utf8.h>
 
 static int fl_match(const char *a, const char *s, int atleast = 1) {
   const char *b = s;
@@ -97,6 +99,11 @@ extern const char *fl_bg2;
   \li -scheme string
   <br>
   Sets the widget scheme using Fl::scheme().
+
+  \li -scaling-factor number
+  <br>
+  Sets the UI scaling factor. This overrides the "FLTK_SCALING_FACTOR"
+  environment variable.
 
   \li -title string
   <br>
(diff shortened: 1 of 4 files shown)


src/Fl_arg.cxx
```diff
@@ -209,6 +215,20 @@ int Fl::arg(int argc, char **argv, int &i) {
...
 
+  } else if (fl_match(s, "scaling-factor", 7)) {
+    float f = 1.0f;
+    if (sscanf(v, "%f", &f) == 1) {
+      Fl::screen_scale(0, f);
+#ifdef WIN32
+      // MSWindows only: use putenv() because setenv() is not available with all compilers
+      char env_buf[64];
+      snprintf(env_buf, 64, "FLTK_SCALING_FACTOR=%f", f);
+      putenv(env_buf);
```
MatthiasWM: putenv does not necessarily copy its buffer. Once env_buffer disappears off the stack, FLTK_SCALING_FACTOR becomes a random value. Hence the introduction of setenv

test/demo.cxx
```diff
@@ -411,6 +411,14 @@ void dobut(Fl_Widget *, long arg) {
 
 #endif
 
+  // append scaling factor to the command line
+  float f = Fl::screen_scale(0);
+  if (f != 1.0f) {
+    char s[40];
+    snprintf(s, sizeof(s), " -scaling-factor %g", f);
+    strncat(command, s, sizeof(command) - strlen(command) - 1);
```
MatthiasWM: strncat is bad because it my truncate without a NUL byte. Use fl_strlcat instead.

src/Fl_arg.cxx
Albrecht-S: ... and that's why fl_putenv() exists.

Albrecht-S: Also: `#ifdef WIN32` is not allowed inside the library. `fl_putenv()` would solve this too.

GermanAizek: Many thanks for quick code review im fix it in https://github.com/fltk/fltk/pull/1522/commits/b049dfcd7847c3ca64f674297122a2835ac08c3a

test/demo.cxx
GermanAizek: Many thanks for quick code review im fix it in https://github.com/fltk/fltk/pull/1522/commits/b049dfcd7847c3ca64f674297122a2835ac08c3a

GermanAizek:
@MatthiasWM @Albrecht-S done I moved fl_strlcat to fl_string_functions.h for use in demo.cxx

Issue #1276, opened 2025-07-06
Fix usage of isspace(), ispunct(), and more ...


Albrecht-S:
**Description of the Bug Report**
These functions are not unicode aware and their behavior can be locale dependent. This leads to all kinds of failures, although the wrong usage is often benign.
For more details please refer to [STR 3436](https://www.fltk.org/str.php?L3436).

**To Reproduce**
**Simple example:** A file name in Cyrillic letters like `Тест.fl` (would be `test.fl` in ASCII) produces incorrect output in the header file because the letters of the file name are misinterpreted.

1. Save a file named `Тест.fl` with any fluid contents (copy-paste the file name to be sure it's all Cyrillic).
2. Run `fluid -c Тест.fl`
3. See output below
4. Note the header guards `#ifndef __________h` etc.

```
$ cat Тест.fl
# data file for the Fltk User Interface Designer (fluid)
version 1.0400
header_name {.h}
code_name {.cxx}
comment {Minimal fluid example} {in_source in_header
}

decl {\#include <FL/Fl_Box.H>} {public global
}
$ head Тест.h
// Minimal fluid example
// generated by Fast Light User Interface Designer (fluid) version 1.0500

#ifndef __________h
#define __________h
#include <FL/Fl.H>
#include <FL/Fl_Box.H>
#endif
```

**FLTK Version**
 - current git branch `master` as of today
 - commit 2a1677802fe68a9745e9ab6b2dcbeb515e50886c

**FLTK Configure / Build Options**
Default CMake build: this doesn't matter.

**Operating System / Platform:**
All platforms

**Additional context**
The `fluid` issue above is just an example, all similar functions must be checked. See below for a list of known functions that have similar issues.

Albrecht-S:
For a (more up-to-date) list of known functions with such issues in Unicode context please see comment https://github.com/fltk/fltk/issues/1276#issuecomment-4925248520 below.

(List of functions removed here).

The check marks in the list ~above~ below should be checked when all occurrences of a particular function have been fixed. If possible, add the commit of the fix.

Note that it is not sufficient to mask single characters like `c & 255` or similar (see also comments on [STR 3436](https://www.fltk.org/str.php?L3436))

~If possible, the more important changes should be backported to 1.4.~
**_EDIT (Jul 14, 2026):_** we do no longer backport to 1.4 (unless the fix is really critical).

MatthiasWM:
[...]

PR #1523, opened 2026-07-13
added impl ASCII tolower and toupper and replace all

GermanAizek:
Reference: https://github.com/fltk/fltk/issues/1276


MatthiasWM (commented):
tests and examples can't include files in src, the WIN32 test for keyboard events could benefit from using the locale based toupper and tolower. Other than that, the PR seems fine!

PR #1527, opened 2026-07-15
JPEG_Image: migrate lstat from heap to stack

GermanAizek:
@MatthiasWM 

In the old code, the programmer had to manually call delete lstat in every place where the function could terminate ahead of time (for example, if there was an error opening the ERR_FILE_ACCESS file or an error in the ERR_FORMAT format). If at least one delete was skipped (or the code did not reach it), a memory leak occurred.

Removing all delete lstat calls has made the function cleaner and more reliable. Now the logic of exiting the function (via return) has become safe, since the stack is cleared automatically by C/C++ compiler.

Allocating memory on the stack is significantly faster than allocating memory on the heap via new, as it boils down to simply changing the CPU stack pointer rather than searching for a free block on the heap.

Reference with benchmarks:

### DIFF
--- a/src/Fl_JPEG_Image.cxx
+++ b/src/Fl_JPEG_Image.cxx
@@ -242,7 +242,7 @@ void Fl_JPEG_Image::load_jpg_(const char *filename, const char *sharename, const
   fl_jpeg_error_mgr       jerr;     // Error handler info
   JSAMPROW                row;      // Sample row pointer
 
-  struct load_stat *lstat = new load_stat();
+  struct load_stat lstat;
 
   // Clear data...
   alloc_array = 0;
@@ -250,15 +250,13 @@ void Fl_JPEG_Image::load_jpg_(const char *filename, const char *sharename, const
 
   // Open the image file if we read from the file system
   if (filename) {
-    if ((lstat->fp = fl_fopen(filename, "rb")) == NULL) {
+    if ((lstat.fp = fl_fopen(filename, "rb")) == NULL) {
       ld(ERR_FILE_ACCESS);
-      delete lstat;
       return;
     }
   } else {
     if (data==0L) {
       ld(ERR_FILE_ACCESS);
-      delete lstat;
       return;
     }
   }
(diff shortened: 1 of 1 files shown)


src/Fl_JPEG_Image.cxx
```diff
@@ -242,23 +242,21 @@ void Fl_JPEG_Image::load_jpg_(const char *filename, const char *sharename, const
   fl_jpeg_error_mgr       jerr;     // Error handler info
   JSAMPROW                row;      // Sample row pointer
 
-  struct load_stat *lstat = new load_stat();
+  struct load_stat lstat;
```
MatthiasWM, line 245: This change adds a bug by no longer calling the special constructor.

MatthiasWM:
Rejected. The PR is buggy.

erco77:
Hmm, I was genuinely curious about this; the following test code build/run seems to show the ctor is called in either case, as in c++ struct and class seems similar in this respect:

```
$ cat foo.cxx
#include <stdio.h>

// A copy of the definition of load_stat from Fl_JPEG_Image.cxx
struct load_stat {
  FILE *fp;                             // file pointer if reading from disk
  int max_finish_decompress_err;        // count errors and give up after a while
  int max_destroy_decompress_err;       // ... to avoid recursion and deadlock

  load_stat() {         // c'tor
    printf("CTOR\n");
    fp = nullptr;
    max_finish_decompress_err  = 5;
    max_destroy_decompress_err = 5;
  }
  ~load_stat() {        // d'tor
    if (fp)
      fclose(fp);
  }
};

int main() {

    struct load_stat lstat1;   // test automatic instancing 
    struct load_stat *lstat2 = new load_stat(); // test dynamic instancing
    printf("1:");
    printf("         fp: %p\n", lstat1.fp);
    printf(" max_finish: %d\n", lstat1.max_finish_decompress_err);
    printf("max_destroy: %d\n", lstat1.max_destroy_decompress_err);
    printf("2:");
    printf("         fp: %p\n", lstat2->fp);
    printf(" max_finish: %d\n", lstat2->max_finish_decompress_err);
    printf("max_destroy: %d\n", lstat2->max_destroy_decompress_err);
    return 0;
}

$ g++ foo.cxx -o foo

$ ./foo
CTOR            <-- automatic
CTOR            <-- dynamic
1:         fp: (nil)
 max_finish: 5
max_destroy: 5
2:         fp: (nil)
 max_finish: 5
max_destroy: 5
```

src/Fl_JPEG_Image.cxx
MatthiasWM: Never mind, I was wrong.

MatthiasWM:
Apologies. I misread the constructor for a function call. I am so used to class names being capitalized that I did not see `load_stat:: load_stat()` for what it was.

MatthiasWM:
I did commit the PR.

MatthiasWM:
@GermanAizek My apologies for crushing all your PRs into two compressed commits. It's just so much easier for us to handle few commits with the same theme vs. many commits that change only a single line.

PR #1528, opened 2026-07-15
Graphics_Driver: migrate rgb from heap to stack

GermanAizek:
Parent PR for more info:

Reference with benchmarks:


MatthiasWM:
Rejected. Petty patch without true benefit in working and proven code.

PR #1531, opened 2026-07-15
Window_Driver: migrate rgba from heap to stack

GermanAizek:
Parent PR for more info:

Reference with benchmarks:


MatthiasWM (commented):
Acceptable, will be merged into single commit.

MatthiasWM:
Changed my mind: Rejected. Petty patch without true benefit in working and proven code.

PR #1532, opened 2026-07-15
PostScript_image: migrate rgb from heap to stack

GermanAizek:
Parent PR for more info:

Reference with benchmarks:

### DIFF
--- a/src/drivers/PostScript/Fl_PostScript_image.cxx
+++ b/src/drivers/PostScript/Fl_PostScript_image.cxx
@@ -77,9 +77,8 @@ static void destroy_BGRA(void *data) {
 
 
 void Fl_PostScript_Graphics_Driver::draw_pixmap(Fl_Pixmap *pxm,int XP, int YP, int WP, int HP, int cx, int cy) {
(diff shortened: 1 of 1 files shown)


src/drivers/PostScript/Fl_PostScript_image.cxx
```diff
@@ -77,9 +77,8 @@ static void destroy_BGRA(void *data) {
 
 
 void Fl_PostScript_Graphics_Driver::draw_pixmap(Fl_Pixmap *pxm,int XP, int YP, int WP, int HP, int cx, int cy) {
-  Fl_RGB_Image *rgb =  new Fl_RGB_Image(pxm);
-  draw_rgb_bitmap_(rgb, XP, YP, WP, HP, cx, cy);
-  delete rgb;
+  Fl_RGB_Image rgb(pxm);
```
MatthiasWM, line 80: Acceptable, will be merged into single commit.

MatthiasWM:
Changed my mind: Rejected. Petty patch without true benefit in working and proven code.

PR #1534, opened 2026-07-15
Cairo_Graphics_Driver: migrate rgb and ddashes from heap to stack

GermanAizek:
Reference with benchmarks:

### DIFF
--- a/src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
+++ b/src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
@@ -340,13 +340,12 @@ void Fl_Cairo_Graphics_Driver::line_style(int style, int width, char* dashes) {
   else c_join = CAIRO_LINE_JOIN_MITER;
   cairo_set_line_join(cairo_, c_join);
 
-  double *ddashes = NULL;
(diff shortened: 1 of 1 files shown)


src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
```diff
@@ -340,13 +340,12 @@ void Fl_Cairo_Graphics_Driver::line_style(int style, int width, char* dashes) {
   else c_join = CAIRO_LINE_JOIN_MITER;
   cairo_set_line_join(cairo_, c_join);
 
-  double *ddashes = NULL;
-  int l = 0;
+  double ddashes_array[6]; // Max 6 elements for dashes_flat/cap
```
Albrecht-S, line 343: This allocation is buggy.

Albrecht-S (changes requested):

Albrecht-S:
The accepted answer in your link to a 12-year old SO question includes a summary:
> Note: these costs are often insignificant - not even worth a look or second thought unless you're writing something where latency or throughput are enormously important.

This is not the case for a gui library like FLTK.

PR #1535, opened 2026-07-15
PostScript: migrate off, modal, psize from heap to stack

GermanAizek:
Reference with benchmarks:

### DIFF
--- a/FL/fl_callback_macros.H
+++ b/FL/fl_callback_macros.H
@@ -568,7 +568,7 @@
       _FL_CBD_CLASS_NAME(TYPE0 p0, TYPE1 p1) \
       : p0_(p0), p1_(p1) { } \
     }; \
-    WIDGET->callback(_FL_CBD_CLASS_NAME::cb, new _FL_CBD_CLASS_NAME(VALUE0, VALUE1), true); \
+    WIDGET.callback(_FL_CBD_CLASS_NAME::cb, new _FL_CBD_CLASS_NAME(VALUE0, VALUE1), true); \
   } while(0)
 
 #define FL_INLINE_CALLBACK_1(WIDGET, TYPE0, NAME0, VALUE0, LAMBDA) \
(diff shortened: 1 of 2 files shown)


FL/fl_callback_macros.H
```diff
@@ -568,7 +568,7 @@
       _FL_CBD_CLASS_NAME(TYPE0 p0, TYPE1 p1) \
       : p0_(p0), p1_(p1) { } \
     }; \
-    WIDGET->callback(_FL_CBD_CLASS_NAME::cb, new _FL_CBD_CLASS_NAME(VALUE0, VALUE1), true); \
```
MatthiasWM, line 571: No! This is obviously untested and false!

src/drivers/PostScript/Fl_PostScript.cxx
```diff
@@ -1137,8 +1137,8 @@ void Fl_PostScript_Graphics_Driver::transformed_draw_extra(const char* str, int
   // create an offscreen image of the string
   Fl_Color text_color = Fl_Graphics_Driver::color();
   Fl_Color bg_color = fl_contrast(FL_WHITE, text_color);
-  Fl_Image_Surface *off = new Fl_Image_Surface(w_scaled, (int)(h+3*scale), 1);
-  Fl_Surface_Device::push_current(off);
+  Fl_Image_Surface off(w_scaled, (int)(h+3*scale), 1);
```
MatthiasWM, line 1140: Acceptabel

MatthiasWM: Will be combined with other PRs

MatthiasWM:
Rejected. The first patch is plain wrong. The second patch if petty. The third patch would make sense if the dialog was auto-deleted through a unique_ptr. Having an entire dialog box hierarchy on the stack can be harmful.

PR #1550, opened 2026-08-01
all: fix Clang-Tidy warnings not all members are init inside constructor

GermanAizek:


FL/Fl_Check_Browser.H
```diff
@@ -64,7 +64,7 @@ public: // IRIX 5.3 C++ compiler doesn't support private structures...
 private:
   cb_item *first;
   cb_item *last;
-  cb_item *cache;
+  cb_item *cache{};
```
Albrecht-S: This is a "fix", but it's the wrong way to do it.
- If it's a safety measure, then all variables or at least all pointers should be zero-initialized.
- If it's not a safety measure, then the initialization could also be done in the c'tor.

Anyway, "fixing" things just because a compiler flags the code is not good advice.

FL/Fl_Graphics_Driver.H
```diff
@@ -165,14 +165,14 @@ protected:
   Fl_Color color_; ///< current color
   int sptr;///< For internal use by FLTK
   static const int matrix_stack_size = FL_MATRIX_STACK_SIZE; ///< For internal use by FLTK
-  matrix stack[FL_MATRIX_STACK_SIZE]; ///< For internal use by FLTK
-  matrix m; ///< current transformation matrix
+  matrix stack[FL_MATRIX_STACK_SIZE]{}; ///< For internal use by FLTK
```
Albrecht-S: 1. Initializing this entire array is wasting CPU time.
2. A better way is to use a vector or any other C++11 object which is already planned.
3. This array is used as a stack, and every entry is properly initialized before use, therefore this initialization is counter-productive, see 1.

Not acceptable.

FL/Fl_Graphics_Driver.H
```diff
@@ -165,14 +165,14 @@ protected:
...
+  matrix m{}; ///< current transformation matrix
```
Albrecht-S: Could be done, but is useless.

src/drivers/X11/Fl_X11_Screen_Driver.H
```diff
@@ -44,8 +44,8 @@ protected:
     float scale;
 #endif
   } FLScreenInfo;
-  FLScreenInfo screens[MAX_SCREENS];
-  float dpi[MAX_SCREENS][2];
+  FLScreenInfo screens[MAX_SCREENS]{};
+  float dpi[MAX_SCREENS][2]{};
```
Albrecht-S: Once again, initializing large arrays is wasting CPU time.

src/fl_show_colormap.cxx
```diff
@@ -29,8 +29,8 @@
 */
 class ColorMenu : public Fl_Window {
   Fl_Color initial;
-  Fl_Color which, previous;
   int done;
+  Fl_Color which, previous{};
```
Albrecht-S: Why was this statement moved down, and why are other Fl_Color variables not initialized?

Again, this is a bad solution w/o solving the real problem.

Albrecht-S (changes requested):
All over: not acceptable although some minor parts could be done.

src/fl_show_colormap.cxx
GermanAizek: yes you're right `Fl_color which` skipped initialization

src/drivers/X11/Fl_X11_Screen_Driver.H
GermanAizek: Are you sure there won't be a case of access uninitialized data `dpi` and `screens`?

Albrecht-S: If you find one, please let us know.

src/fl_show_colormap.cxx
Albrecht-S: ... and `initial` as well.

FL/Fl_Check_Browser.H
GermanAizek: fix it here https://github.com/fltk/fltk/pull/1550/commits/2b52fccf86f30658810c3370e326b5c05dee80c8

FL/Fl_Graphics_Driver.H
GermanAizek: fix it here https://github.com/fltk/fltk/pull/1550/commits/2b52fccf86f30658810c3370e326b5c05dee80c8

FL/Fl_Graphics_Driver.H
GermanAizek: fix it here https://github.com/fltk/fltk/pull/1550/commits/2b52fccf86f30658810c3370e326b5c05dee80c8

src/drivers/X11/Fl_X11_Screen_Driver.H
GermanAizek: fix it here https://github.com/fltk/fltk/pull/1550/commits/2b52fccf86f30658810c3370e326b5c05dee80c8

src/fl_show_colormap.cxx
GermanAizek: fix it here https://github.com/fltk/fltk/pull/1550/commits/2b52fccf86f30658810c3370e326b5c05dee80c8

src/fl_show_colormap.cxx
```diff
@@ -28,9 +28,9 @@
  Pretty much unchanged from Forms.
 */
 class ColorMenu : public Fl_Window {
-  Fl_Color initial;
-  Fl_Color which, previous;
+  Fl_Color initial{};
   int done;
+  Fl_Color which{}, previous{};
```
Albrecht-S, line 32: You still didn't answer the question why you reordered the statements

GermanAizek: fix it in commit https://github.com/fltk/fltk/pull/1550/commits/7c9d04de09e678c9e6d8873b99b2d50aa9b726b2

MatthiasWM:
Implemented in a more classic FLTK style.

PR #1552, opened 2026-08-01
Actions CI: add new Github workflow for all OS and cpu arch

GermanAizek:
> wouldn't you like Actions CI to build every PR request and check the tests? I can do this because it will be easier for you to conduct a code review and generally that PR is going to be successful on different OS platforms.


Albrecht-S:
No.

Please don't post unsolicited PR's. We like to discuss such changes before so we don't have to review PR's we don't want.

Issue #1311, opened 2025-10-07
New Wayland subwindow code creates issues with resizes of opengl windows inside a group


ggarra13:
**Description of the Bug Report**
The new and simplified subwindow code for dragging it added a new bug that makes the OpenGL/Vulkan subwindows inside a group that is resized, not refresh properly.

**To Reproduce**
N/A

**Expected behavior**
No redraw issues and subwindows to appear properly.

**Screenshots**

In this video, the subwindows move properly, but the timeline with the thumbnails has refresh issues and leaves garbage behind (part of the timeline window):

https://github.com/user-attachments/assets/8285d3d9-2579-4728-b638-a888b7a33aaa

**FLTK Version**
1.5 GH

**Operating System / Platform:**
Ubuntu 25.04

**Linux/Unix Runtime, if applicable:**
 - Wayland

ggarra13:
Find attached two of the driver files that changed.

[Fl_Wayland_Window_Driver.ok.cxx.txt](https://github.com/user-attachments/files/22738436/Fl_Wayland_Window_Driver.ok.cxx.txt)

[Fl_Wayland_Window_Driver.bad.cxx.txt](https://github.com/user-attachments/files/22734350/Fl_Wayland_Window_Driver.bad.cxx.txt)

ManoloFLTK:
Please, describe what is wrong in the video and exactly what resize/move operation of what widget/window is being done and fails.

A short program to reproduce the issue would be useful.

ggarra13:
> Please, describe what is wrong in the video and exactly what resize/move operation of what widget/window is being done and fails.

In the video, first, we check moving of the subwindows.  All works fine there.   

Then the timeline is expanded with the Edit button to show the thumbnails.  The first time it partially works, leaving some thumbnails without creation.  The second time it does not work at all, leaving only traces of the thumbnails in the middle of the screen (with the size of the unexpanded timeline).  Repeating the operation seems to sometimes work and sometimes not work.  Dragging the timeline separator seems to refresh properly in this video, but I've seen it with problems too.

All operations used to work fine with Fl_Wayland_Window_Driver.ok.cxx.

ManoloFLTK:
Sorry. I need either a program to reproduce the issue or much more detail about what is being resized and fails: for what widget is Fl_Widget::resize() called? what subwindows does this widget contain? are these subwindows Fl_Gl_Window's? are other operations such as playing a video simultaneously performed?

[...]

PR #1562, opened 2026-08-03
Wayland patch with new find_pen_below and better tooltip handling.

ggarra13:
This patch addresses two bugs in the current implementation:

- It avoids tablet events (native or synthesized) to incorrectly be sent to any tooltip window, which could make tooltip windows flicker.
- It improves upon find_pen_below to find the widgets recursively, allowing child widgets and subwindows to find the correct widget/window.  Groups (and Windows) are searched in drawing order, so that frontmost subwindows take precedence over those below.  This is needed for allow_expand_outside_parent() Windows to work correctly, for example.

### DIFF
--- a/src/drivers/Wayland/Fl_Wayland_Pen_Driver.cxx
+++ b/src/drivers/Wayland/Fl_Wayland_Pen_Driver.cxx
@@ -118,10 +118,6 @@ struct TabletTool {
   struct zwp_tablet_tool_v2      *wl_tool;
   enum zwp_tablet_tool_v2_type    type;
 
-  double         decor_sx { 0 };      // surface-local x when over a decoration surface
-  double         decor_sy { 0 };      // surface-local y when over a decoration surface
-  struct libdecor_frame *focus_frame { nullptr }; // owning frame when over a decoration
-
   uint64_t                        hardware_serial;
   int                             pen_id;       // int-sized pen identity
   Trait                           capabilities; // reported by capability events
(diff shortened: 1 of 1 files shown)


ManoloFLTK:
> This is needed for allow_expand_outside_parent() Windows to work correctly, for example.

Subwindows with `allow_expand_outside_parent()` were introduced (by me) as a workaround for the lack of possibility under Wayland to drag a window above another window and to detect their relative positions. All of this will be made pointless when the real support under Wayland of docking widgets will be merged (discussion #1257), hopefully into FLTK 1.5. That new approach will also fix pending issue #1311.

ggarra13:
@ManoloFLTK I hope you are not planning to deprecate the allow_expand_outside_parent() API and its currently correct working code.  Discussion #1257, as I pointed out, still needs to support multiple windows docking into the same docking target.
Issue #1311 may be unrelated, as we never got around to find the actual culprit of the OpenGL problem.  I am currently patching it in my Vulkan branch (but you did not accept the patch for not being able to reproduce it in your code).

ManoloFLTK:
> Discussion https://github.com/fltk/fltk/discussions/1257, as I pointed out, still needs to support multiple windows docking into the same docking target.

As I pointed out, the docking/undocking widget supports multiple docking windows to the same target.

MatthiasWM:
Thanks. Tested on my HP and seems to work fine. I can't say anything about the other issues that were brought up in the discussion.

Issue #1546, opened 2026-07-30
Crash when using copied SVG image


L0PiTaL:
**Description of the Bug Report**
Crash when using SVG image on Windows

**To Reproduce**
The "stack trace" would look like the following:

```
Fl_SVG_Image::draw
  resize()                         // original is rasterized
  Fl_RGB_Image::draw
    Fl_GDI_Graphics_Driver::draw_rgb
      Fl_Graphics_Driver::draw_rgb   // when alpha blending is unavailable
        img2 = img->copy(...)      // virtual call: Fl_SVG_Image::copy()
                                                 // img2 is a new lazy SVG, img2->array == NULL
        cache(img2)                     // Calls Fl_GDI_Graphics_Driver::cache with img2->array = NULL
          fl_draw_image(img2->array)  // Calls fl_draw_image with a NULL pointer
            scan_cb()
              memcpy(..., NULL, ...)
```

**Expected behavior**
Not crash

**FLTK Version**
1.4.4, but based on source code it seems it is still present in master

**Operating System / Platform:**
Windows 10

**Additional context**

A solution would be similar to this one: https://github.com/fltk/fltk/issues/1363
Just adding `normalize()` after the copy. Diff is from 1.4.4, but same for master:

```
diff --git a/src/Fl_Graphics_Driver.cxx b/src/Fl_Graphics_Driver.cxx
index 494d6333c..9785ed60e 100644
--- a/src/Fl_Graphics_Driver.cxx
+++ b/src/Fl_Graphics_Driver.cxx
@@ -316,6 +316,7 @@ void Fl_Graphics_Driver::draw_rgb(Fl_RGB_Image *img, int XP, int YP, int WP, int
     Fl_RGB_Scaling keep = Fl_Image::RGB_scaling();
     Fl_Image::RGB_scaling(Fl_Image::scaling_algorithm());
     Fl_RGB_Image *img2 = (Fl_RGB_Image*)img->copy(w2, h2);
+    img2->normalize();
     Fl_Image::RGB_scaling(keep);
     cache(img2);
     draw_fixed(img2, XP, YP, WP, HP, cx, cy);
```

ManoloFLTK:
[...]

PR #1569, opened 2026-08-09
Fix UAF in Fl_Cairo_Graphics_Driver::clean_utf8()

rodarima:
The buffer returned by realloc can start at a different address, causing all previous pointers to become stale. Instead of using the pointer q to the current location in the reallocated buffer, use just an offset so we always point to the correct place with utf8_buffer + q.

Similarly, we get rid of retval and keep track of the private buffer usage and first call using the explicit flag use_priv_buffer.

We also avoid increasing the buffer by N + 1000 (doubling it each time) and only do increments of 1000 bytes.

---

Found while fuzzing Dillo, reproducer:

```c++
/* 
 * Build with:
 * clang++ `fltk-config --cflags` -o badfltk -fsanitize=address `fltk-config --ldflags` badfltk.cc
 */

#include <FL/Fl.H>
#include <FL/Fl_Window.H>
#include <FL/fl_draw.H>

static const unsigned char badtext[] = {
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80,
	0x80, 0x80, 0x80, 0x80, 0x80, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xe2, 0x80,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	0xff, 0xff, 0xff, 0xff
};

class BadWindow : public Fl_Window
{
public:
    BadWindow() : Fl_Window(400, 100) {}

    void draw() override {
        Fl_Window::draw();

        fl_font(FL_HELVETICA, 16);

        fl_draw(
            reinterpret_cast<const char *>(badtext),
            sizeof(badtext),
            10, 50
        );
    }
};

int main() {
    BadWindow win;
    win.show();
    return Fl::run();
}
```

### DIFF
--- a/src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
+++ b/src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
@@ -1,7 +1,7 @@
 //
 // Support for Cairo graphics for the Fast Light Tool Kit (FLTK).
 //
-// Copyright 2021-2024 by Bill Spitzak and others.
+// Copyright 2021-2026 by Bill Spitzak and others.
 //
 // This library is free software. Distribution and use rights are outlined in
 // the file "COPYING" which should have been included with this file.  If this
@@ -1285,32 +1285,52 @@ void Fl_Cairo_Graphics_Driver::font(Fl_Font fnum, Fl_Fontsize s) {
 const char *Fl_Cairo_Graphics_Driver::clean_utf8(const char* str, int &n) {
   static char *utf8_buffer = NULL;
   static int utf8_buffer_len = 0;
-  char *q = utf8_buffer;
+  int q = 0;
   const char *p = str;
-  const char *retval = str;
   int len, len2;
   const char *end = str + n;
   char buf4[4];
+  int use_priv_buffer = 0;
   while (p < end) {
     unsigned codepoint = fl_utf8decode(p, end, &len);
-    if (retval != str || (len == 1 &&  *(uchar*)p >= 0x80)) { // switch to using utf8_buffer
-      len2 = fl_utf8encode(codepoint, buf4);
-      if (!utf8_buffer_len || utf8_buffer_len < (q - utf8_buffer) + len2) {
-        utf8_buffer_len += (q - utf8_buffer) + len2 + 1000;
+
+    // Switch to using utf8_buffer if needed
+    if (!use_priv_buffer && (len == 1 && *(uchar*)p >= 0x80)) {
+      // Make room if needed
+      if (utf8_buffer_len < p - str) {
+        utf8_buffer_len = (p - str) + 1000;
         utf8_buffer = (char *)realloc(utf8_buffer, utf8_buffer_len);
       }
-      if (retval == str) {
-        retval = utf8_buffer;
-        q = utf8_buffer;
-        if (p > str) { memcpy(q, str, p - str); q += (p - str); }
+      use_priv_buffer = 1;
+      // Fill private buffer with data so far
+      if (p > str) {
+        memcpy(utf8_buffer, str, p - str);
+        q = (p - str);
       }
-      memcpy(q, buf4, len2);
+    }
+
+    if (use_priv_buffer) {
+      len2 = fl_utf8encode(codepoint, buf4);
+
+      // Increase buffer if needed
+      if (utf8_buffer_len < q + len2) {
+        utf8_buffer_len = q + len2 + 1000;
+        utf8_buffer = (char *)realloc(utf8_buffer, utf8_buffer_len);
+      }
+
+      // Copy decoded codepoint
+      memcpy(utf8_buffer + q, buf4, len2);
       q += len2;
     }
     p += len;
   }
-  if (retval != str) n = q - retval;
-  return retval;
+
+  if (use_priv_buffer) {
+    n = q;
+    return utf8_buffer;
+  }
+
+  return str;
 }
 
 


Albrecht-S:
Thank you for this PR which looks solid at a first glance. Just to clarify: UAF == Use After Free.

I don't want to merge it though yet, because ...

There are some minor issues (at least one) I would have flagged in a review. One issue I have in mind (a redundant `if` condition) is not your fault, it existed already in the old code.

That said, we're talking about FLTK 1.5 (no more backports to 1.4), and this enables us to use `std::string` for `utf8_buffer` which would, AFAICT, simplify the code substantially and make it less error-prone. This is one of the reasons we decided to use C++11, BTW: to get rid of such self-allocated and extended buffers.

While I wrote the paragraph above I noticed that Dillo doesn't use 1.5 yet (officially, although I built it successfully with 1.5 IIRC). This bug really has crash potential, which would justify a backport -- and that would preclude using `std::string` because it's not allowed in 1.4...

Well, these are my thoughts for now, we need to evaluate the situation. Maybe the final solution could be a fix like yours for 1.4 (thoroughly reviewed and tested) and a rewrite using `std::string` for 1.5?

OK, I just wanted to let everyone know of my thoughts. Comments, anybody?

@rodarima What is the current official development status of Dillo WRT FLTK (1.3, 1.4, 1.5, ...) ?

My next steps will be to test with the given demo program.

Albrecht-S:
Bug confirmed with the demo program using ASAN (just to be sure), although it was obvious, given the description. But anyway, such a reproducer is helpful to confirm the **fix** as well.

@rodarima Thank you for this excellent bug report and demo program.

I'll continue reviewing your fix later, and I'm also thinking about an alternative approach with `std::string` for 1.5 only. Let's see how this works out...

rodarima:
Also, realloc can fail returning NULL, but is another problem other than the UAF. It will crash at the memcpy so it won't go too far, but not sure how you want to cover that one. I suggest solving that in another commit/PR.

> One issue I have in mind (a redundant if condition) is not your fault, it existed already in the old code.

I'm not sure if there is a redundant `if`, I think they are all needed. Which one you see?

> What is the current official development status of Dillo WRT FLTK (1.3, 1.4, 1.5, ...) ?

Dillo supports platforms that don't have support for C++11, so I don't think we can fully switch to 1.5. Probably I will try to keep support both with 1.3 and newer releases. FLTK 1.4 support is covered here: https://bug.dillo-browser.org/439/

There are several problems in different parts (specially Wayland) and I will try fix it or report what I find.

src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
```diff
@@ -1285,32 +1285,50 @@ void Fl_Cairo_Graphics_Driver::font(Fl_Font fnum, Fl_Fontsize s) {
...
+    if (!use_priv_buffer && (len == 1 &&  *(uchar*)p >= 0x80))
+      use_priv_buffer = 1;
+
+    if (use_priv_buffer) {
       len2 = fl_utf8encode(codepoint, buf4);
-      if (!utf8_buffer_len || utf8_buffer_len < (q - utf8_buffer) + len2) {
-        utf8_buffer_len += (q - utf8_buffer) + len2 + 1000;
+
+      // Increase buffer if needed
+      if (!utf8_buffer_len || utf8_buffer_len < q + len2) {
```
Albrecht-S: The first part of this `if` appears to be redundant. It could be written differently:
```C++
  if (utf8_buffer_len != 0 || utf8_buffer_len < q + len2) {
```
Assuming `utf8_buffer_len` always `>= 0` (must be non-negative, although it's declared as `int`).

If `utf8_buffer_len > 0`, then the second term is irrelevant, the entire `if` is `true`.

If `utf8_buffer_len == 0` the second term applies, but then it can be rewritten to `0 < q + len2` or `q + len2 > 0` which should always be true, i.e. the second term is redundant (always true).

But that would mean that the entire `if` is always true, and the buffer is **always** (re)allocated (for each byte in the while loop). Weird. This is really a good task for AI ;-) or a static code analyzer.

Now I'm really confused. Where is my mistake, what did I miss? Or was this code always flawed? Any help (correction) would be appreciated.

This is one more reason to use `std::string` for the buffer, isn't it?

Note: I'm now going to test my theory with the old code, before I apply the suggested fix.

Albrecht-S:
> FLTK 1.4 support is covered here: https://bug.dillo-browser.org/439/

Thanks, I bookmarked this page. Maybe I (or we, the FLTK Team) can help with some issues, for instance by providing more configuration support in `FL/fl_config.h` (available since 1.4). However, please don't expect such additions in 1.4, this would probably be done exclusively in 1.5. I suggest that you open an issue, or even better a discussion ([in section Development](https://github.com/fltk/fltk/discussions/categories/development)) to tell us what configure items you'd want to see in that file (Xft, Pango, ...?). Please don't continue this particular topic here (it's OT).

Albrecht-S:
> I'm not sure if there is a redundant if, I think they are all needed. Which one you see?

@rodarima Please see the review I started (above, not finished yet) for my analysis. It's really confusing, maybe I made a mistake, but I'm now going to test (verify?) my theory.

If my analysis turned out to be correct, then the entire `if` would be redundant (always `true`).

rodarima:
> Please see the review I started

Is not public unless you do either "Start review"+"Submit review" or "Comment" for single comments. Probably you want the latter.

Albrecht-S (changes requested):
Preliminary review, still work in progress.

Albrecht-S:
> not public unless you do either "Start review"+"Submit review" or "Comment" for single comments.

Thanks, I didn't know that, of course I see my own comments. Always learning something new on GitHub. ;-)
Now it should be public.

Albrecht-S:
BTW: testing the old code reveals that the realloc happens
- once at the start of the loop
- after several bytes have been processed (not for each byte)

The crash (UAF) happens directly after the realloc, which was to be expected.

src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
Albrecht-S: OK, I found my mistake in the analysis. `!utf8_buffer_len` is equivalent to `utf8_buffer_len == 0` (not `utf8_buffer_len != 0`). The rest of my code analysis was correct, except the conclusion.

I can remove the first part of the `if`, and the code works fine, i.e. the first condition of the `if` is redundant. It may be a shortcut for the very first round of the loop, and that only once because the buffer is `static`. Thus it's not useful. In all following iterations it's an unnecessary condition and uses a tiny bit more CPU. Not worth any changes, but I wondered what this was about when I saw it for the first time.

@rodarima Do you agree?

Albrecht-S:
OK, testing the new code revealed: it works fine, and ASAN doesn't find any bugs. I finally tested with this slightly modified code fragment:
```C++
      // Increase buffer if needed
      if (/* !utf8_buffer_len || */ utf8_buffer_len < q + len2) {
        int new_len = (q + len2 + 255) & ~127 ;
        fprintf(stderr, "realloc(utf8_buffer, %6d), old = %6d\n", new_len, utf8_buffer_len);
        fflush(stderr);
        utf8_buffer_len = new_len; // q + len2 + 1000;
        utf8_buffer = (char *)realloc(utf8_buffer, utf8_buffer_len);
      }
```

#### Changes:
1. First part of the `if` statement commented out
2. define `new_len` in 128 byte increments (power of 2, just for demonstration)
3. log realloc's
4. use new_len to reallocate the buffer.

#### Output:
```
$ bin/test/pr-1596_fix-uaf 
realloc(utf8_buffer,    256), old =      0
realloc(utf8_buffer,    512), old =    256
realloc(utf8_buffer,    768), old =    512
realloc(utf8_buffer,   1024), old =    768
```

The output looks good, and the code is "flexible" (using a different increment than 1000), but could also be used with 1000 or 1024 (or whatever). That's just for demonstration, not intended for real usage.

That's not yet a complete review, but it confirms the fix.

Albrecht-S:
Yet another code update, just in case someone is interested in testing:
```C++
      // Increase buffer if needed
      if (utf8_buffer_len < q + len2) {
        utf8_buffer_len = (q + len2 + 0x1ff) & ~0x1ff ; // round up to 512 byte chunks (power of 2)
        // fprintf(stderr, "Fl_Cairo_Graphics_Driver::clean_utf8: [re]alloc(utf8_buffer,%5d)\n", utf8_buffer_len);
        // fflush(stderr);
        utf8_buffer = (char *)realloc(utf8_buffer, utf8_buffer_len);
      }
```

The full diff (including copyright year and a tiny spelling fix):
[pr-1596_fix-uaf.diff.txt](https://github.com/user-attachments/files/30907792/pr-1596_fix-uaf.diff.txt)

I also checked that `branch-1.4` is **_compatible_** so this patch could be backported (as of today).

src/drivers/Cairo/Fl_Cairo_Graphics_Driver.cxx
```diff
@@ -1285,32 +1285,50 @@ void Fl_Cairo_Graphics_Driver::font(Fl_Font fnum, Fl_Fontsize s) {
...
-      if (retval == str) {
-        retval = utf8_buffer;
-        q = utf8_buffer;
-        if (p > str) { memcpy(q, str, p - str); q += (p - str); }
+
+      // Fill private buffer with data so far on first call
+      if (use_priv_buffer == 1) {
+        if (p > str) {
+          memcpy(utf8_buffer, str, p - str);
+          q += (p - str);
```
Albrecht-S: I propose to change this to
```C++
          q = p - str;
```
At this point `q` **must** be equal to `p - str`, and indeed, at this time `q == 0` because `use_priv_buffer == 1` (`q` has never been updated so far). This is why the current code works.

However, a direct assignment `q = p - str;` makes this clear.

Albrecht-S:
I also notice:
- the original buffer **doesn't need** to be `nul` terminated (but it can be)
- the copied UTF-8-clean buffer is **never** `nul` terminated.

This may have been intended (or the latter is not required) but it should be mentioned in the documentation (comment) of `Fl_Cairo_Graphics_Driver::clean_utf8()`. This is not something concerning this patch, it's probably been this way before. Just mentioning it here.

Also, FTR: this method is defined in class `Fl_Cairo_Graphics_Driver` (static, public) but it is used in several other graphics (and PostScript) drivers. I think we could find a better place in a base class of `Fl_Cairo_Graphics_Driver` or anywhere else.

Potential plan to deal with this PR:
1. Test, debug, confirm, apply to 1.5 (including any more appropriate fixes and docs etc.).
2. Backport to 1.4 because it's high priority with crash potential. This should include only the relevant patch (once done) w/o further changes.
3. FLTK 1.5 only:
  - move to a better place
  - rewrite using `std::string` for the buffer

Point 3. is optional and not relevant for a release, but I strongly recommend to fix this before 1.5 will be released.

Albrecht-S:
@rodarima Once we're done with testing I'd like to push my updates to your branch if you don't mind and then squash and merge your branch with my additions. Is this OK for you?

Albrecht-S:
1. :-1: Bad news: I found a serious bug in the patch.
2. :+1: Good news: I also found a (preliminary) fix. :-)
3. :-1: Bad news: I need further tests, and it's too late for me to do it today. I need a break.
4. :+1: Good news: I'll get back to you tomorrow with news, but this may be late (afternoon, evening).

Please don't hold your breath.

Hint: you need a text with a very long line that consists of many [1] valid char's before the first invalid character that triggers copying the buffer to the local buffer. The demo code made the mistake to have invalid chars right at the beginning. But don't waste your time, I have a fix.

-----
[1] larger than the first allocation size of the local buffer

ManoloFLTK:
@Albrecht-S If the future fix of this issue gets backported to 1.4, I'd recommend to also backport the one-line fix of #1546 that does produce a crash under Windows (albeit in some not entirely clarified situations). Also, I would suggest to move the `clean_utf8` member function to class `Fl_Graphics_Driver`.

rodarima:
Oh, I saw the other problem not exposed by my test case. I made the logic simpler by calling realloc on two paths, first time when switching to the private buffer and when the space is low. That should get rid of that problem.

I also addressed your other two comments. Let me know if you see something else.

Edit: Prevent reducing the already allocated buffer in first realloc.

Albrecht-S:
@ManoloFLTK wrote:
> If the future fix of this issue gets backported to 1.4, I'd recommend to also backport the one-line fix of https://github.com/fltk/fltk/issues/1546 that does produce a crash under Windows ...

Noted, I'll consider this.

> Also, I would suggest to move the clean_utf8 member function to class Fl_Graphics_Driver.

Yes, that was my plan - although only for 1.5, together with the rewrite using `std::string` I'm planning to do.

Albrecht-S:
> I saw the other problem not exposed by my test case. I made the logic simpler by calling realloc on two paths, first time when switching to the private buffer and when the space is low.

@rodarima Thanks for updating your PR. Looks good now and passes my tests. I'll make some final tests, but I'm confident that the latest version is fine.

Here's my plan:

1. Final tests and checks by AI (done).
2. Merge the existing PR to `master` to resolve the potential crash in 1.5. This will close this PR.
3. Backport to `branch-1.4` to resolve the potential crash in 1.4.
4. Later: refactor using `std::string` and move this method to class `Fl_Graphics_Driver`. I'll do this shortly, when time permits. No assistance needed - unless I ask. ;-)

OK, I asked Claude to check the code. It replied: "_it does what it's meant to (strip invalid UTF-8 bytes from a string), but there are several real issues..._". Besides the known and intended issues (realloc not checked, never freeing the static buffer, not reentrant) one point raised my attention:

> Re-encoding valid codepoints unnecessarily. Once use_priv_buffer is set, every subsequent codepoint — including perfectly valid ones — gets round-tripped through fl_utf8decode + fl_utf8encode instead of just being memcpy'd from the source.

When I read this I also noticed that `buf4` was only 4 bytes, although (invalid!) code-points could store up to 6 bytes. This is a known issue, although it _shouldn't_ happen in practice.

So yes, I'm now going to merge this PR (in `master`), then I'll make a small modification, and finally I'll backport it to 1.4.

Albrecht-S:
OK, I merged this PR (4b961017423b686f3774b7fc8367a1f6d37518ad) and added my own updates in another commit (f46d2ab028807dfe44f206fd747acc89cfc6d72c).

Feedback welcome.

Next step: backport to 1.4 - after a break...

Albrecht-S:
Done. Both `master` and `branch-1.4` have been updated.

PR #1579, opened 2026-08-13
Breaking change: Use floats for Fl::e_dx and Fl::e_dy

### LINKED ISSUE #1571: Smoother touch pad scrolling
(Tested on e11495ae8566c995e14736d98083f0e3d5879bb2)

When using the touch pad of my laptop on Wayland, the data returned by `wl_fixed_to_double(value)` is a double that ranges from 0 to around 10 or 15 (can reach higher values like 50-60). When using the mouse wheel, it always returns -15 or +15 for each wheel tick.

In `src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx`:

https://github.com/fltk/fltk/blob/e11495ae8566c995e14736d98083f0e3d5879bb2/src/drivers/Wayland/Fl_Wayland_Screen_Driver.cxx#L391-L398

You are currently dividing the delta only when it passes above 10, so you are causing the mapping:

```
[0..10) -> [0..10)
[10..N) -> [1..N/10)
```
This causes a sudden reduction of speed when passing from 10 to above. You can see the data below with ddelta being the raw double reported by wayland and delta the value after the mapping:

<details>
<summary>Input data</summary>

```
# From trackpad
FL_MOUSEWHEEL: V ddelta=-1.527344 delta=-1
FL_MOUSEWHEEL: V ddelta=-6.277344 delta=-6
FL_MOUSEWHEEL: V ddelta=-10.011719 delta=-1 <-- divided by 10
FL_MOUSEWHEEL: V ddelta=-11.371094 delta=-1
FL_MOUSEWHEEL: V ddelta=-16.292969 delta=-1
FL_MOUSEWHEEL: V ddelta=-10.351562 delta=-1
FL_MOUSEWHEEL: V ddelta=-11.031250 delta=-1
FL_MOUSEWHEEL: V ddelta=-9.332031 delta=-9
FL_MOUSEWHEEL: V ddelta=-13.066406 delta=-1
FL_MOUSEWHEEL: V ddelta=-11.371094 delta=-1
FL_MOUSEWHEEL: V ddelta=-11.031250 delta=-1
FL_MOUSEWHEEL: V ddelta=-10.691406 delta=-1
FL_MOUSEWHEEL: V ddelta=-6.957031 delta=-6
FL_MOUSEWHEEL: V ddelta=-7.976562 delta=-7
FL_MOUSEWHEEL: V ddelta=-6.109375 delta=-6
FL_MOUSEWHEEL: V ddelta=-5.601562 delta=-5
FL_MOUSEWHEEL: V ddelta=-3.902344 delta=-3
FL_MOUSEWHEEL: V ddelta=-2.546875 delta=-2
FL_MOUSEWHEEL: V ddelta=-2.375000 delta=-2
FL_MOUSEWHEEL: V ddelta=-1.695312 delta=-1
FL_MOUSEWHEEL: V ddelta=-1.187500 delta=-1
FL_MOUSEWHEEL: V ddelta=-0.507812 delta=0
FL_MOUSEWHEEL: V ddelta=-0.507812 delta=0

# Mouse wheel
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1 <-- also divided
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=15.000000 delta=1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
FL_MOUSEWHEEL: V ddelta=-15.000000 delta=-1
```
</details>

Both cases are sent to Dillo using the same FL_MOUSEWHEEL event type, but they mean different things.

This causes that if I scroll with the touch pad it advances at much **higher** speed than intended if the value is below 10, or goes very **slow** if it increases above 10 (is very annoying).

Given that you are already returning +1/-1, we don't have room to return fractional displacements for the touch pad. If you return a float this would be easy, but not all machines have a float unit and I think you can still use integers efficiently.

My suggestion is to return the delta multiplied by (1024.0 / 15.0) instead of just 1 or -1, and divide it later on by 1024 after scaling it by the linesize.

This has plenty of precision to compute a fractional displacement using only integers and then the compiler can optimize away the division of 1024 with just a bit shift (notice the `sar eax, 10` instruction for the 1024 division):

https://godbolt.org/z/z1nxb11Yd

Here is an example patch that can be tested with the fonts test program (provided you have plenty of fonts to use the scrollbar).

[0001-Use-fractional-delta-for-Fl-e_dy.patch](https://github.com/user-attachments/files/30949756/0001-Use-fractional-delta-for-Fl-e_dy.patch)

This solution preserves the displacement as int and the event type as FL_MOUSEWHEEL for both input types, so I think is not very invasive. The same should be done for the X axis and propagated to all other backends.

I tested this in Dillo and it produces a good experience, similar to other browsers. I also tested that I can scroll pixel by pixel if I do it slowly, so we have plenty of precision with integers.

Another problem that we may want to consider is if we want to expose that the two events come from different inputs so I can have different sensitivities in Dillo.

rodarima:
Allow floating point displacements for smooth scrolling.

**NOTE**: Only tested in Wayland with a touch pad and mouse wheel (works fine).
TODO: Test in other platforms.

Fixes #1571

### DIFF
--- a/FL/core/events.H
+++ b/FL/core/events.H
@@ -45,8 +45,8 @@ FL_EXPORT extern int                e_x;                ///< Mouse X position (w
 FL_EXPORT extern int                e_y;                ///< Mouse Y position (window relative)
 FL_EXPORT extern int                e_x_root;           ///< Mouse X position (screen absolute)
 FL_EXPORT extern int                e_y_root;           ///< Mouse Y position (screen absolute)
-FL_EXPORT extern int                e_dx;               ///< Mouse wheel horizontal delta
-FL_EXPORT extern int                e_dy;               ///< Mouse wheel vertical delta
+FL_EXPORT extern float              e_dx;               ///< Mouse wheel horizontal delta
+FL_EXPORT extern float              e_dy;               ///< Mouse wheel vertical delta
 
 // Mouse click handling
 FL_EXPORT extern int                e_clicks;           ///< Number of consecutive clicks
@@ -132,13 +132,13 @@ FL_EXPORT inline int event_y_root()             { return e_y_root; }
   Returns the current horizontal mouse scrolling associated with the
   FL_MOUSEWHEEL event. Right is positive.
 */
-FL_EXPORT inline int event_dx()                 { return e_dx; }
+FL_EXPORT inline float event_dx()               { return e_dx; }
 
 /**
   Returns the current vertical mouse scrolling associated with the
   FL_MOUSEWHEEL event. Down is positive.
 */
-FL_EXPORT inline int event_dy()                 { return e_dy; }
+FL_EXPORT inline float event_dy()               { return e_dy; }
 
 //
 // Mouse Query Functions
--- a/src/Fl.cxx
+++ b/src/Fl.cxx
@@ -52,8 +52,6 @@ int             Fl::damage_,
                 Fl::e_y,
                 Fl::e_x_root,
                 Fl::e_y_root,
-                Fl::e_dx,
-                Fl::e_dy,
                 Fl::e_state,
                 Fl::e_clicks,
                 Fl::e_is_click,
@@ -61,6 +59,10 @@ int             Fl::damage_,
                 Fl::e_original_keysym,
                 Fl::scrollbar_size_ = 16,
                 Fl::menu_linespacing_ = 4;      // 4: was a local macro in Fl_Menu.cxx called "LEADING"
+
+// Displacement in linesize units
+float           Fl::e_dx,
+                Fl::e_dy;
 
 char            *Fl::e_text = (char *)"";
 int             Fl::e_length;
--- a/src/Fl_cocoa.mm
+++ b/src/Fl_cocoa.mm
@@ -962,11 +962,8 @@ static void cocoaMouseWheelHandler(NSEvent *theEvent)
   float s = Fl::screen_driver()->scale(0);
   float edx = [theEvent deltaX];
   float edy = [theEvent deltaY];
-  int dx = roundf(edx / s);
-  int dy = roundf(edy / s);
-  // make sure that even small wheel movements count at least as one unit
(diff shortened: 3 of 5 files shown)


MatthiasWM:
I would not change the type of Fl::event_dx() please. Apart from new compiler warnings, older software that relies on integer values would completely ignore small scroll increments, even if there are many of them. I suggested a solution that I beleive works here: https://github.com/fltk/fltk/issues/1571#issuecomment-5303334706

rodarima:
Yes, this is an intended breaking change. I wanted to test if this would even work (I only tested in Wayland) and how it would simplify the code in FLTK. I may open another PR using a new variable instead so it is backwards compatible.

Albrecht-S:
@rodarima Please consider closing this PR in favor of PR #1581.
1. We won't change the type of Fl::event_dx() and Fl::event_dy() for backward compatibility
2. PR #1581 adds float variables for the same reason as your PR.

Is there anything missing from #1581 that you would need?

PS: if you don't close it yourself it will be closed as soon as #1581 will be merged.

rodarima:
Closing in favor of https://github.com/fltk/fltk/pull/1581.

PR #1582, opened 2026-08-20
Fluid: add Preprocessor node type

MatthiasWM:
In Fluid, we had the Declaration Block node which was meant to allow conditional compiling depending on preprocessor defines. 

This node had a lot of issues. There was no way to add and #else branch or an #elif branch. The node could only be at the top level of the tree. It made parsing a lot harder because finding a node in the hierarchy had to skip this particular node. And last but not least, the generated code did not always work.

A new node type "Preprocessor Directive" solves all of the issues above. It can be inserted anywhere in the tree (except menu item arrays, but that is a whole other story). It can generate #if, #elif, #else, and #endif nodes, and the generated code has been tested and works well. 

Conditionally Preprocessor Directives must be place carefully, but that's no different to regular coding.

The documentation is updated, and a snapshot of the dialog box is generated.

All in all, I am pretty happy with this new Node. The old node type was rarely used, but it is still read and generates to original code.


MatthiasWM:
Compiles on all platforms

MatthiasWM:
I decided to merge this PR after using the new feature a little bit. There is still tons of room for improvement, but this is already such a nice feature, replacing the faulty old code. While of course staying back compatible. We wouldn't be FLTK if we weren't compatible all the way backt to 1997 ;-)

PR #1597, opened 2026-09-01
Add lambdas with captures as callback option.

MatthiasWM:
@Albrecht-S @ManoloFLTK @RokerHRO @erco77 

As discussed in #1442, this is another proposal for adding std::function as a callback option. It's type safe, convenient, and removes UB function pointer casting. Capturing lambdas are available since C++11.

This implementation uses a template class to create an interface that can be used with any class that needs a callback interface, and is not derived from Fl_Widget (Fl_File_Chooser, for example).

In phase 1, I implemented the basic idea without changing the API at all. Feedback is highly appreciated. Commit 1 is tested on macOS with all apps in test and Fluid and compiles without any new warnings.

In phase 2, I added the Fl_Callback_Interface to Fl_File_Chooser. This works well and is easy to implement, just as planned.

Phase 3 adds a user data destructor callback as discussed in #1442 for both Fl_Callback_User_Data and void* callbacks. Fl_Widget size increases from 2 words for old-style callback to 3 words by adding the second callback.

In phase 4, I added std::function as a new callback option without changing the logic of the old callback system. Fl_Widget size increase from 3 words to 4 words.This requires Fl_Widget to implement a simple on/off flag. If I don;t do that, alignment will require a 5th word in the worst case.

📍<- you are here

Phase 5 is optional because it may break the API, but fixes ambiguities and undefined behavior in our current code. For example, casting `void()(Fl_Widget*, long)` to `void()(Fl_Widget*, void*)` is UB

### DIFF
--- a/FL/Fl_Callback_Interface.H
+++ b/FL/Fl_Callback_Interface.H
@@ -21,6 +21,7 @@
 
 #include <functional>
 #include <new>
+#include <cstddef>
 
 
 #ifndef Fl_Callback_User_Data_H
--- a/FL/Fl_Callback_Interface.H
+++ b/FL/Fl_Callback_Interface.H
@@ -58,7 +58,28 @@ public:
 
 
 /**
- Add callback functionality to the class T.
+ An interface template class that adds callback functionality to the class T.
+
+ This interface adds modern C++ and classic FLTK callback functionality to any
+ class T. Use this interface by adding it as one base class in the declaration
+ of class T with T itself as the template parameter.
+
+ ```cpp
+ class MyClass : public Fl_Callback_Interface<MyClass> {
+    // ...
+ };
+ ```
+ The interface adds member functions set a type-safe std::function callback
+ or classic FLTK callback functions and user data that require type casting.
+
+ Fl_Callback_Interface needs to store a single bit of infomation in the derived
+ class. To do this, `MyClass` must implement the following functions:
+
+ ```cpp
+ bool callback_is_stdf() const;
+ void set_callback_stdf();
+ void clear_callback_stdf();
+ ```
 
  Classes that use this interface must implement the following functions that
  simply get, set, and clear a bit, must be false initially:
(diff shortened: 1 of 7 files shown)


MatthiasWM:
To sum my idea up, this implementation works very well. The code in Fl_Callback_Interface.H is quite packed, but this is a lot better than adding it to the already large Fl_Widget.H and Fl_Widget.cxx.

Using std::function is really neat and keeps the callback where the widget constructor is. My silly test is:
```
  Fl_Button *a = new Fl_Button(20, 40, 200, 30, "Click Me");
  Fl_Button *b = new Fl_Button(20, 80, 200, 30, "vanish");
  a->callback(
    [a]() {
      printf("%d\n", a->value());
      a->label("Clicked!");
      a->redraw();
    }
  );
  b->callback(
    [a]() {
      delete a;
    }
  );
```
There is not a single type cast in the code. Calling `a->value()` is no problem because a is a `Fl_Button*`, so this is all nice and type-safe.

There is never a win without negatives. So here we go:
* maintainability: the callback system has become complex. But I think the benefits and modernization outweighs that. Yes, I still need to add comments to the Fl_Callback_Interface class.
* base class vs. interface vs. multiple inheritance: Fl_Callback_Interface may look like a base class, but it's not. Its constructor and destructor are protected to make that clear. Multiple inheritance is possible if needed because Fl_Callback_Interface has a very small name pollution (`callback_interface_` besides the public methods) and no virtual destructor.
* size and speed: no worries, the compiler will optimize almost everything away.
* error messages: this is the only bummer. If there is an error in the user's callback code, the error messages become more cryptic, but far from the std library gibberish.

Anyway, if you made it all the way down here, any feedback is appreciated.

ManoloFLTK:
> Using std::function is really neat and keeps the callback where the widget constructor is.

I agree it can be nice to have a widget callback coded just below where its widget is constructed. My impression, however, is that this is practical only for very short (3  to 4 lines) callback functions. Making more complex ones into a separate function is generally necessary for code readability. A clear advantage of making a callback an `std::function` is when more than one data arguments are necessary; this approach seems a clear improvement over macros `FL_FUNCTION_CALLBACK_#` and friends.

I wonder whether class `Fl_Callback_Interface` could be used with `Fl::add_timeout()` to create a timeout?

MatthiasWM:
> A clear advantage of making a callback an `std::function` is when more than one data arguments are necessary; this approach seems a clear improvement over macros `FL_FUNCTION_CALLBACK_#` and friends.

Yes. `FL_FUNCTION_CALLBACK_#` was my reaction one FLTK being stuck with C++98. In retrospect, my macros are implemented very similar to the implementation of capturing lambdas. If we decide to use this PR, I will mark all `FL_FUNCTION_CALLBACK_#` macros as deprecated.

> I wonder whether class `Fl_Callback_Interface` could be used with `Fl::add_timeout()` to create a timeout?

Yes, but it needs a class. We would have to add a class like `Fl_App_Timer` with the callback interface. So the user could do stuff like:
```
main() {
  static Fl_App_Timer t;
  t->callback([]{ fl_beep(); }
  t->add_timeout(5.0);
}
```
I will check if it is possible to build a class that works as a function argument. It should be.

MatthiasWM:
Just looked the implementation of Fl_Timeout. Yes, that should work well.

There seems to be a pretty complete list of handler (or callbacks) in `Fl/core/function_types.H`:
```
typedef void (*Fl_Timeout_Handler)(void *data);
typedef void (*Fl_Awake_Handler)(void *data);
typedef void (*Fl_Idle_Handler)(void *data);
typedef void (*Fl_Old_Idle_Handler)();
typedef void (*Fl_FD_Handler)(FL_SOCKET fd, void *data);
typedef int (*Fl_Event_Handler)(int event);
typedef int (*Fl_System_Handler)(void *event, void *data);
typedef int (*Fl_Event_Dispatch)(int event, Fl_Window *w);
typedef void (*Fl_Abort_Handler)(const char *format,...);
typedef void (*Fl_Atclose_Handler)(Fl_Window *window, void *data);
typedef int (*Fl_Args_Handler)(int argc, char **argv, int &i);
typedef void (*Fl_Clipboard_Notify_Handler)(int source, void *data);
```
I'll have to check if there are any other handlers that could benefit.

MatthiasWM:
I added support for callbacks with the footprint `void ()(void*)` and `void ()(long)` to support `Fl_Timeout`. By inheritance, those become available for all widgets now as well. Nice!

ManoloFLTK:
@MatthiasWM Please, explain how to use the new code to create a timeout callback with several arguments transmitted as captures of a lambda, if that is indeed possible.

MatthiasWM:
> @MatthiasWM Please, explain how to use the new code to create a timeout callback with several arguments transmitted as captures of a lambda, if that is indeed possible.

I created PR #1600 to test this out. It's not fully implemented, but good enough for testing. So here is a silly code change for `test/button.cxx` that shows a timeout lambda inside a callback lambda with captures:

```
  Fl_Button *b2 = new Fl_Button(120,20, 80, 25, "&no op");
  b2->callback(
    [b1, b2]() {
      Fl::add_timeout(1.0, [b1, b2]() {
        fl_message("The label of b1 is currently '%s'", b1->label());
        b2->deactivate();
      });
    }
  );
```

I think the code is very readable. Better than spreading it across multiple callback functions and global variables.

An alternative syntax that can capture as well and avoids creating a lambda is `std::bind()`, (here we capture b1) for example:
```
      Fl::add_timeout(
        1.0,
        std::bind(
          fl_message,
          "The label of b1 is currently '%s'",
          b1->label())
      );
```
This can also bind C++ member functions if needed.

ManoloFLTK:
@MatthiasWM I think there's an error in function `Fl_Widget::do_callback()`. Its first statement
```
  if (!callback()) return;
```
should become
```
  if (!callback() && !callback_is_stdf()) return;
```
That's because the 2 member variables `callback_` and `stdf_callback_` are in an union but have distinct sizes, so that the second can be non-null while the first is null. I ommit all details here, but I had an error because of that when trying to replace the instances of `FL_INLINE_CALLBACK_#` in `src/drivers/PostScript/Fl_PostScript.cxx` by lambdas. The error occurred under macOS built with `FLTK_BACKEND_X11=1` and not under Linux, but it occurred.

MatthiasWM:
The method callback() is there for back compatibility. To make sure that it works with lambdas that do not have an address, callback() returns a pointer to a harmless fake function. Historic callback() should only be used to check against nullptr.

ManoloFLTK:
OK. The method `callback()` was fixed in commit [Phase 4: add std::function callbacks](https://github.com/fltk/fltk/pull/1597/changes/40f640c0b2c6a4936ae1557754ffba3ab81de956) which added the check to `callback_is_stdf()`. I was using an earlier commit where the error occurred. After pulling to the last commits, no error under macOS+X11.

RokerHRO:
Why using an unsafe `union` and not `std::variant<…>`, we have C++ here!

ManoloFLTK (commented):
I generally support this PR which adds the possibility to use lambdas as widget callbacks because it allows to easily capture the variables necessary to a specific callback.

I have used it in a simple test case and found it to work very well.

However, I would also believe there's no reason to do that for widget callbacks and not for other callback functions used by FLTK, for example timeout callbacks, file descriptor callbacks, etc. This would mean to also commit PR #1600 which is not yet complete and which is reported to require an undesired "templatification". I'm not familiar enough with advanced C++-11 features such as `std::forward, std::move, std::function` to really evaluate this question.

ManoloFLTK:
> Why using an unsafe union and not std::variant<…>, we have C++ here!

FLTK 1.5 limits itself to C++11 and  `cppreference.com` indicates that `std::variant` requires C++17.

MatthiasWM:
And variants are two words, increasing size of the callback class by yet another word to five CPU words.

MatthiasWM:
> This would mean to also commit PR #1600 which is not yet complete and which is reported to require an undesired "templatification". I'm not familiar enough with advanced C++-11 features such as `std::forward, std::move, std::function` to really evaluate this question.

Thanks, Manolo. #1600 was a test with all what C++11 has to offer. Instead, I'd like to rewrite this with a little less templates and cleaner interface.

std::foward lat's you map one call onto another by forwarding all or part of the function parameters. We have callback((*)(Fl_WIdget*,void*)), callback((*)(Fl_WIdget*,long)), callback(std::function<void()>), etc. . By writing Fl::add_timeout(double, std::forward<...>...), C++ automatically writes l::add_timeout(double, (*)(Fl_WIdget*,void*)) etc. for us.

I made the callback interface not movable and not copiable on purpose to avoid all the issues that arise from managing void*. So no std::move needed. Fl_Widget is not movable or copiable anyway.

std::function is very similar to what the callback macros do: it creates a struct that stores the captured values (to ones between `[` and `]`), and create the same static callback function that FLTK generally uses. It just does it in the background, keeping code readable. Lambdas are meant to be short (Python lambdas are limited to one function) and just call something more complex in a flexible and readable way.

